PR #30, opened 2022-01-18
Switch to downstream fork of cloudkitty

jovial:
This brings in a few additional features not in the regular wallaby release.

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -104,6 +104,17 @@ kolla_sources:
     type: git
     location: "{{ kolla_bifrost_source_url }}"
     reference: "{{ kolla_bifrost_source_version }}"
+  cloudkitty-base:
+    # Backport of NOTNUMBOOL: https://review.opendev.org/c/openstack/cloudkitty/+/774084
+    # Support customising Prometheus queries: https://review.opendev.org/c/openstack/cloudkitty/+/817275
+    type: git
+    location: https://github.com/stackhpc/cloudkitty.git
+    reference: stackhpc/wallaby
+  horizon-plugin-cloudkitty-dashboard:
+    # Fixes CloudKitty pricing panel: https://github.com/stackhpc/cloudkitty-dashboard/pull/1
+    type: git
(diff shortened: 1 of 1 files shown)


etc/kayobe/kolla.yml
```diff
@@ -181,6 +181,13 @@ kolla_build_blocks:
                -e '/#baseurl.*/a baseurl={{ repo.url }}' /etc/yum.repos.d/{{ repo.file }}{% if not loop.last %} &&{% endif %} \
     {% endfor %}
 
+  cloudkitty_base_footer: |
+    # Backport of NOTNUMBOOL: https://review.opendev.org/c/openstack/cloudkitty/+/774084
+    # Support customising Prometheus queries: https://review.opendev.org/c/openstack/cloudkitty/+/817275
+    RUN pip install git+https://github.com/stackhpc/cloudkitty.git@stackhpc/wallaby
+  horizon_footer: |
+    # Fixes CloudKitty pricing panel: https://github.com/stackhpc/cloudkitty-dashboard/pull/1
+    RUN pip install git+https://github.com/stackhpc/cloudkitty-dashboard.git@stackhpc/wallaby
```
markgoddard: Could you use the kolla_sources approach rather than footers?

jovial: Oh yeah, missed that thanks.

etc/kayobe/kolla.yml
```diff
@@ -104,6 +104,17 @@ kolla_sources:
     type: git
     location: "{{ kolla_bifrost_source_url }}"
     reference: "{{ kolla_bifrost_source_version }}"
+  cloudkitty-base:
+    # Backport of NOTNUMBOOL: https://review.opendev.org/c/openstack/cloudkitty/+/774084
```
jovial: Not sure if I should bother maintaining these comments here as they may get out of sync with the branch. Thoughts?

markgoddard: They could go stale, maybe leave them out.

markgoddard (approved):

PR #32, opened 2022-01-19
Update image tags

jovial:
This brings in:

- https://github.com/stackhpc/kolla/pull/90
- https://github.com/stackhpc/stackhpc-kayobe-config/pull/30

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -2,6 +2,11 @@
 docker_yum_baseurl: "{{ stackhpc_repo_docker_url }}"
 docker_yum_gpgkey: "https://download.docker.com/linux/{% raw %}{{ ansible_distribution | lower }}{% endraw %}/gpg"
 
+cloudkitty_tag: wallaby-20220119T122428
+horizon_tag: wallaby-20220119T122428
 manila_tag: wallaby-20211210T140839
 neutron_tag: wallaby-20220104T102739
 neutron_sriov_agent_tag: wallaby-20220110T113538
+prometheus_jiralert_tag: wallaby-20220119T122428
+prometheus_msteams_tag: wallaby-20220119T122428
+prometheus_openstack_exporter_tag: wallaby-20220119T122428


etc/kayobe/kolla/globals.yml
```diff
@@ -2,6 +2,11 @@
...
 docker_yum_gpgkey: "https://download.docker.com/linux/{% raw %}{{ ansible_distribution | lower }}{% endraw %}/gpg"
 
+cloudkitty_tag: wallaby-20220119T122428
+horizon_tag: wallaby-20220119T122428
 manila_tag: wallaby-20211210T140839
 neutron_tag: wallaby-20220104T102739
 neutron_sriov_agent_tag: wallaby-20220110T113538
+prometheus_jiralert_tag: wallaby-20220119T122428
+prometheus_msteams_tag: wallaby-20220119T122428
+prometheus_openstack_exporter_tag: wallaby-20220119T122428
```
jovial, line 12: Note to self: Check this was built. I gave Mark an incomplete list.

markgoddard: twas not

markgoddard: now built and pushed

jovial:
I did hit the following issue on one of my controllers:

In horizon.log:

```
[IP:39466] compressor.exceptions.OfflineGenerationError: You have offline compression enabled but key "dad03ce06cb01a989fde7e905c986a37847a72819bd3221fbb378833292ab128" is missing from offline manifest. You may need to run "python manage.py compress". Here is the original content:
```

My workaround was to run that command inside the docker container:

```
(horizon)[root@controller /]# /var/lib/kolla/venv/bin/manage.py compress
/var/lib/kolla/venv/lib64/python3.6/site-packages/scss/namespace.py:172: DeprecationWarning: inspect.getargspec() is deprecated since Python 3.0, use inspect.signature() or inspect.getfullargspec()
  argspec = inspect.getargspec(function)
Compressing... done
Compressed 7 block(s) from 12 template(s) for 0 context(s).
```
and then restart it:
```
[root@controller horizon]# sudo docker restart horizon
horizon
```
I did a service deploy and not an upgrade or reconfigure; could that explain it?

markgoddard (approved):

PR #36, opened 2022-02-04
Adds pulp_proxy_url

jovial:
Pulp requires you to set the proxy for each remote.


jovial:
This is untested when the pulp_proxy_url is unmodified, but I think you can use omit like this:
```
---

- hosts: localhost
  vars:
    test: "{{ omit }}"
  tasks:
    - debug:
        msg: "{{ test }}"
```

```
[will@juno ~]$ ansible-playbook /tmp/test.yml 
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the
implicit localhost does not match 'all'

PLAY [localhost] *********************************************************************

TASK [Gathering Facts] ***************************************************************
ok: [localhost]

TASK [debug] *************************************************************************
ok: [localhost] => {
    "msg": "Hello world!"
}

PLAY RECAP ***************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0 
```

etc/kayobe/pulp.yml
```diff
@@ -10,6 +10,9 @@ pulp_url: "http://{{ admin_oc_net_name | net_ip(groups['seed'][0]) }}:80"
 pulp_username: admin
 pulp_password: "{{ secrets_pulp_password }}"
 
+# Proxy to use when adding remotes
+pulp_proxy_url: "{{ omit }}"
```
markgoddard, line 14: In Yoga we get proper support for proxies in kayobe: https://review.opendev.org/c/openstack/kayobe/+/814046

Should this default to `https_proxy` when available? Either we could backport the proxy patch downstream, or add a comment here to update the default.

jovial:
I've cherry-picked the proxy support to wallaby here: https://github.com/stackhpc/kayobe/pull/18

markgoddard (dismissed):

priteau (dismissed):

etc/kayobe/pulp.yml
```diff
@@ -10,6 +10,9 @@ pulp_url: "http://{{ admin_oc_net_name | net_ip(groups['seed'][0]) }}:80"
 pulp_username: admin
 pulp_password: "{{ secrets_pulp_password }}"
 
+# Proxy to use when adding remotes.
+pulp_proxy_url: "{{ https_proxy | default(omit) }}"
```
markgoddard: In the case where we have a squid proxy on the seed, this would probably not make sense. Which case is more common? Corporate or squid?

jovial: Should we start with `pulp_proxy_url: "{{ omit }}"` so you have to explicitly define it? We can refine the default at a later time?

markgoddard: :+1:

markgoddard (commented):
Could you add the same for the Apt repos?

etc/kayobe/pulp.yml
jovial: Just to clarify, I set this in https://github.com/stackhpc/stackhpc-kayobe-config/pull/36/commits/5b24e5b4f27fd4991c1704c2aa0aff73b2965184 and then reverted it as we realised it was only a sensible default when using an external proxy. If the proxy is deployed on the seed, we probably don't want to use it.

jovial:
> Could you add the same for the Apt repos?

Done

markgoddard (approved):

PR #60, opened 2022-04-29
CI: seed container image builds, trigger image sync to test pulp

markgoddard:


.github/workflows/container-image-build.yml
```diff
@@ -77,12 +88,71 @@ jobs:
           kayobe overcloud container image build ${{ github.events.input.regexes }} --push
         env:
           KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
+        if: github.events.input.overcloud
+
+      - name: Display the overcloud container image tag
+        run: |
+          echo "$(cat ~/kolla_tag)"
+        if: github.events.input.overcloud
```
priteau: I think all occurrences of `github.events` should be `github.event`.

However I haven't yet verified that the change for `input.regexes` above fixed it.

priteau (changes requested):

.github/workflows/container-image-build.yml
markgoddard: Good catch

priteau: And it should actually be `github.event.inputs` rather than `github.event.input` 😓

priteau (approved):

priteau:
@markgoddard Please check that you are happy with the additional commit I pushed.

.github/workflows/container-image-build.yml
markgoddard: > @markgoddard Please check that you are happy with the additional commit I pushed.

Thank you for fixing up. I spotted this in another workflow :( Would be better if github failed on these invalid references.

PR #69, opened 2022-05-25
Bump up Prometheus stack version

cityofships:
Related: https://github.com/stackhpc/stackhpc-kayobe-config/pull/67

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -15,6 +15,7 @@ nova_tag: wallaby-20220328T103755
 octavia_tag: wallaby-20220524T085900
 ovn_tag: wallaby-20220223T143249
 openvswitch_tag: wallaby-20220223T143249
+prometheus_tag: wallaby-20220525T174700
 prometheus_jiralert_tag: wallaby-20220119T122428
 prometheus_libvirt_exporter_tag: wallaby-20220325T122042
 prometheus_msteams_tag: wallaby-20220119T122428


markgoddard (approved):

etc/kayobe/kolla/globals.yml
```diff
@@ -15,6 +15,7 @@ nova_tag: wallaby-20220328T103755
 octavia_tag: wallaby-20220524T085900
 ovn_tag: wallaby-20220223T143249
 openvswitch_tag: wallaby-20220223T143249
+prometheus_tag: wallaby-20220525T174700
```
markgoddard, line 18: Did you rebuild all prometheus images? Should we remove the ones below?

markgoddard (changes requested):
Waiting for answer

etc/kayobe/kolla/globals.yml
cityofships: The regexp feature is broken, thus all get rebuilt always. Fix in https://github.com/stackhpc/stackhpc-kayobe-config/pull/68
Does ~30 minutes build time of all images match your experience? I saw other jobs that had run for ~4h 30m

markgoddard: 30 minutes would be pretty quick, even if using the cache.

markgoddard: Presumably you at least want to upgrade alertmanager, given you changed the version

cityofships: By default: ```prometheus_alertmanager_tag: "{{ prometheus_tag }}"```
The ones below are pretty custom - I'm not confident they'll work correctly in the newer build, thus leaving them.

markgoddard (approved):

PR #77, opened 2022-06-02
Draft: Wallaby package refresh

cityofships:


etc/kayobe/dnf.yml
```diff
@@ -8,6 +8,8 @@
 # dnf_config:
 #   proxy: http://proxy.example.com
 #dnf_config:
+dnf_config:
+  install_weak_deps: False
```
markgoddard, line 12: Why is this required?

cityofships: It's to keep hosts free of crap like PackageKit

markgoddard:
Stale

PR #78, opened 2022-06-03
Fix evaluation of step conditions

priteau:
Despite specifying the boolean type for inputs, GitHub workflow
variables are actually strings so 'false' evaluates to true. This was
causing the seed container image build to always be executed.

See https://github.com/actions/runner/issues/1483 for more details.


cityofships (approved):
Let's consider splitting into two separate workflows

markgoddard:
:( thanks for fixing

markgoddard:
> Let's consider splitting into two separate workflows

What are the pros & cons?

PR #80, opened 2022-06-07
Generate release notes

markgoddard:


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,28 @@
...
+'on':
+  pull_request:
+jobs:
+  tox:
+      runs-on: ubuntu-20.04
+      strategy:
+        matrix:
+          include:
+            - environment: pep8
+              python-version: 3.6
```
mnasiadka: why not 22.04 and python3.10? It doesn't make sense to use EOL python in a CI job that only generates renos, does it?

mnasiadka (approved):

.github/workflows/pull-request.yml
markgoddard: I was just going with distros & versions that we're actually using in wallaby.

markgoddard:
reno didn't like 3.10

PR #83, opened 2022-06-29
nfv: bump repo version to 20220622T042524

mnasiadka:
Current one doesn't contain OVN 21.12 packages


cityofships:
LGTM, @markgoddard ?

markgoddard:
Should this just be part of the wallaby minor update at this point?

mnasiadka:
Not really, Wallaby minor update touches OS packages IMHO - this is only openvswitch/ovn - and we should be bumping this every now and then to get newer packages (with newer back ports).

markgoddard (approved):

PR #89, opened 2022-07-18
Add support for Ubuntu repositories

markgoddard:
Adds support for using Ubuntu as a host and container OS.

* Sync new Ubuntu Focal repositories in Ark to the local Pulp server
* Configure hosts to use the APT repositories on the local Pulp server
* Build Kolla container images, using Ubuntu repositories in the local Pulp server


etc/kayobe/apt.yml
```diff
@@ -11,6 +11,58 @@
...
+# * filename: name of a file in /etc/apt/apt.conf.d/ in which to write the
+#   configuration
+# Default is an empty list.
+apt_config:
+  # NOTE: Currently the Pulp verbatim publisher does not sync translation
+  # files, which results in apt update failing. Disable translations until this
+  # is resolved.
+  - content: |
+      Acquire::Languages "none";
+    filename: 99no-languages
```
priteau, line 26: Should the second line be aligned with the first?

etc/kayobe/kolla.yml
```diff
@@ -220,6 +220,10 @@ kolla_build_blocks:
     {% endfor %}
   grafana_plugins_install: |
     RUN grafana-cli plugins install vonage-status-panel
+  ironic_conductor_header: |
+    # Fix dependency resolution taking forever on Ubuntu.
+    # Remove when https://review.opendev.org/c/openstack/ironic/+/849491 lands.
```
priteau: This is being merged.

etc/kayobe/apt.yml
markgoddard: No - this is a dict with keys content and filename

etc/kayobe/kolla.yml
markgoddard: Removed from the commit history.

etc/kayobe/apt.yml
priteau: Makes sense!

priteau (approved):

PR #92, opened 2022-07-21
Adds support for cold migration to drain script

jovial:


jovial:
This introduces a change in behaviour. You need to add `openstack_instances` to the limit for the dynamic inventory. Should I have just gone from include tasks instead? Or made it into a role?

jovial:
Closing in favour of simpler approach in https://github.com/stackhpc/stackhpc-kayobe-config/pull/93.

PR #95, opened 2022-07-22
Remove unsupported variables from overcloud-dib.yml

markgoddard:
These came in with the Rocky Linux backport, but without the accompanying support in Kayobe.

### DIFF
--- a/etc/kayobe/overcloud-dib.yml
+++ b/etc/kayobe/overcloud-dib.yml
@@ -10,14 +10,6 @@
 # is {{ os_distribution == 'rocky' }}. This will change in a future release.
 #overcloud_dib_build_host_images:
 
-# List of overcloud host disk images to build. Each element is a dict defining
-# an image in a format accepted by the stackhpc.os-images role. Default is to
-# build an image named "deployment_image" configured with the overcloud_dib_*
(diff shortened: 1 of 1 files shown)


cityofships (approved):

priteau:
Wouldn't it be better to backport the associated commit?

markgoddard:
> Wouldn't it be better to backport the associated commit?

It's not necessary for rocky linux afaik, the config came in by mistake via a merge conflict.

priteau:
I didn't mean for Rocky specifically, but rather if we want to build multiple images or if we want to use custom elements stored in Git.

markgoddard:
> I didn't mean for Rocky specifically, but rather if we want to build multiple images or if we want to use custom elements stored in Git.

I don't mind either way. If someone wants to backport, I'll approve.

markgoddard:
> > I didn't mean for Rocky specifically, but rather if we want to build multiple images or if we want to use custom elements stored in Git.
> 
> I don't mind either way. If someone wants to backport, I'll approve.

Seems I might need the git support today...

priteau:
https://github.com/stackhpc/kayobe/pull/23

PR #99, opened 2022-07-28
Xena: overcloud DIB host image, LVM, swap & growroot

markgoddard:


etc/kayobe/inventory/group_vars/all/stackhpc-lvm
```diff
@@ -0,0 +1,30 @@
...
+###############################################################################
+# StackHPC LVM Logical Volume (LV) configuration.
+
+# StackHPC LVM lv_swap LV size.
+stackhpc_lvm_lv_swap_size: 16g
+
+# StackHPC LVM lv_root LV size.
+stackhpc_lvm_lv_root_size: 50g
+
+# StackHPC LVM lv_tmp LV size.
```
oneswig, line 46: Was there an issue with glance image conversions using /tmp?  We have some users who make very large images, potentially bigger than this.

markgoddard: Wouldn't that be using /tmp inside the glance_api container? Which I think would use the container overlay filesystem by default.

etc/kayobe/stackhpc-lvm.yml
```diff
@@ -0,0 +1,117 @@
+---
```
markgoddard: Might need to move all of this to group_vars, since it's not possible to override these variables on a per-group basis. e.g. `stackhpc_lvm_group_rootvg_disks`, and `stackhpc_lvm_group_rootvg_lvs_extra`.

etc/kayobe/stackhpc-overcloud-dib.yml
```diff
@@ -0,0 +1,212 @@
...
+  elements: "{{ stackhpc_overcloud_dib_elements }}"
+  env: "{{ stackhpc_overcloud_dib_env_vars }}"
+  packages: "{{ stackhpc_overcloud_dib_packages }}"
+
+# StackHPC overcloud DIB image name.
+stackhpc_overcloud_dib_name: "deployment_image"
+
+# StackHPC overcloud DIB image elements.
+stackhpc_overcloud_dib_elements:
+  - "{{ os_distribution }}-minimal"
```
priteau: This won't work for rocky since there isn't a rocky-minimal element.

priteau (changes requested):

etc/kayobe/ansible/growroot.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Ensure growpart is installed
+      package:
+        name: "{% if os_distribution in ['centos', 'rocky'] %}cloud-utils-growpart{% else %}cloud-guest-utils{% endif %}"
+        state: present
+        cache_valid_time: "{{ apt_cache_valid_time if ansible_facts.os_family == 'Debian' else omit }}"
+        update_cache: "{{ True if ansible_facts.os_family == 'Debian' else omit }}"
+      become: True
+
+    - name: Get root PV device
+      command: "pvs --select vg_name=rootvg --reportformat json"
```
markgoddard: @priteau says please make it a variable

etc/kayobe/ansible/swap.yml
```diff
@@ -0,0 +1,26 @@
+---
+- name: Configure swap
+  hosts: overcloud
```
markgoddard: Make optional

jovial: Not all hosts may want swap

etc/kayobe/overcloud-dib.yml
```diff
@@ -10,6 +10,10 @@
 # is False. This will change in a future release.
 #overcloud_dib_build_host_images:
 
+# List of additional host packages to install.
+overcloud_dib_host_packages_extra:
+  - "{% if 'ubuntu-minimal' in overcloud_dib_host_images | map(attribute='elements') | flatten | list %}debootstrap{% endif %}"
```
markgoddard, line 15: This requires a backport of https://review.opendev.org/c/openstack/kayobe/+/850903 in Kayobe.

markgoddard:
Addressed comments, fixed up commits, and rebased onto the `xena-wallaby-merge` branch.

If you would like to compare against the original branch, it is `xena-overcloud-dib-orig`.

All 3 distros build, and deploy in VMs on SMS lab with the expected partition layout. I didn't test bare metal because of the NIC drivers.

etc/kayobe/stackhpc-overcloud-dib.yml
```diff
@@ -0,0 +1,212 @@
...
+###############################################################################
+# Diskimage-builder configuration for overcloud host disk images.
+
+# StackHPC overcloud host disk image Diskimage Builder recipe. This may be used
+# as an item in the overcloud_dib_host_images list when
+# overcloud_dib_build_host_images is true.
+# Example:
+# overcloud_dib_host_images:
+#   - "{{ stackhpc_overcloud_dib_host_image }}"
+stackhpc_overcloud_dib_host_image:
```
markgoddard, line 13: Wondering if this name should be any more specific than this? Will this be the only image we provide?

mnasiadka: I guess let's start with that and see if we need more (hope not).
Do we plan to build those images in some CI and upload to Ark? Would help environments without ,,normal'' internet access.

mnasiadka (dismissed):

etc/kayobe/stackhpc-overcloud-dib.yml
markgoddard: Yes. I have a workflow to build the images in progress, but not uploading to Ark yet. Probably @jackhodgkiss will look at that.

etc/kayobe/overcloud-dib.yml
priteau: This hasn't merged yet although the depending change has and Workflow+1 has been added. I am asking on IRC.

priteau: This should be fixed by https://review.opendev.org/c/openstack/project-config/+/853003.

etc/kayobe/ansible/swap.yml
```diff
@@ -0,0 +1,35 @@
...
+        dev: "{{ swap_device }}"
+
+    - name: Ensure swap device present in fstab
+      mount:
+        name: "none"
+        src: "{{ swap_device }}"
+        fstype: "swap"
+        state: "present"
+
+    # It does no harm to run this when swap is already active
```
jovial, line 31: Why use the:
```
      when:
        - ansible_facts.swaptotal_mb == 0
```
condition then?

markgoddard: Because that was already in the playbook when I picked it up :)

I guess it also helps with not showing things as changed. There may be cases where we avoid activating swap though, e.g. if facts are stale, or there is another swap device/file.

Should we remove it?

jovial: Ahh, yeah, of course... For idempotency since that task will always report a changed status. Makes sense, thanks. I did find that in verbose mode swapon produces some output you could parse:
```
[will@juno ~]$ sudo swapon -av
swapon: /swapfile: found signature [pagesize=4096, signature=swap]
swapon: /swapfile: pagesize=4096, swapsize=2147483648, devsize=2147483648
swapon /swapfile
[will@juno ~]$ sudo swapon -av
swapon: /swapfile: already active -- ignored
```
to set changed_when, but feels like bike shedding.

jovial (dismissed):

etc/kayobe/ansible/growroot.yml
```diff
@@ -0,0 +1,62 @@
...
+    ansible_python_interpreter: /usr/bin/python3
+    # Work around no known_hosts entry on first boot.
+    ansible_ssh_common_args: "-o StrictHostKeyChecking=no"
+    # Name of the LVM VG containing the root PV.
+    growroot_vg: "rootvg"
+
+  tasks:
+    - name: Ensure growpart is installed
+      package:
+        name: "{% if os_distribution in ['centos', 'rocky'] %}cloud-utils-growpart{% else %}cloud-guest-utils{% endif %}"
```
jovial: This uses facts but the playbook doesn't collect them:

```
- name: Grow root PV
  hosts: "{{ growroot_group | default('overcloud') }}"
  # Avoid using facts because this may be used as a pre overcloud host
  # configure hook, and we don't want to populate the fact cache (if one is in
  # use) with the bootstrap user's context.
  gather_facts: false
```
Noticed this when trying to run it. Can we invalidate or refresh the cache for the kayobe playbooks?

jovial: I seems you can use `meta: clear_facts` to clear the fact cache, but I guess clearing the cache is a nasty side effect!

jovial: Possibly we can use `setup` with `gather_subset` to only collect the ones we need?

markgoddard: Good catch. I'll use facts, with a fallback to `os_distribution`.

etc/kayobe/overcloud-dib.yml
markgoddard: https://github.com/stackhpc/kayobe/pull/39

jovial (dismissed):

PR #105, opened 2022-08-08
CI: Allow building Ubuntu Kolla container images (Wallaby)

markgoddard:


markgoddard:
Requires some release train side config for image promotion & other things, but this allows you to build Ubuntu containers.

In future I think we should build both sets at once with a single tag, but that requires a bit more effort.

priteau (approved):

markgoddard:
Resolved merge conflicts.

priteau (approved):

PR #135, opened 2022-08-17
README: fix ci-aio and ci-builder setup instructions

markgoddard:


README.rst
```diff
@@ -238,6 +238,13 @@ Setup
 
 Access the host via SSH.
 
+Install package dependencies. On Ubuntu:
+
+.. code-block:: console
+
+   sudo apt update
+   sudo apt install python3-virtualenv
```
jackhodgkiss, line 246: Would it not also be beneficial to add [acl](https://pkgs.org/download/acl) as a required package?

markgoddard: Good question. I'm not sure where is the right place to install that package. Ansible docs suggest it happens when using `become` and `become_user` with an unprivileged user, which is not that common. Still, it might be useful to install the package in Kayobe.

In fact, we already do, just in a specific code path that you didn't hit:

```
# NOTE(mgoddard): The acl package is required for the setfacl command, used by
# become_user.
- name: Ensure acl package is installed
  package:
    name: acl
    cache_valid_time: "{{ apt_cache_valid_time if ansible_facts.os_family == 'Debian' else omit }}"
    update_cache: "{{ True if ansible_facts.os_family == 'Debian' else omit }}"
  become: true
```

So we could move that code somewhere more general.

markgoddard: https://review.opendev.org/c/openstack/kayobe/+/853424

jackhodgkiss (approved):

PR #137, opened 2022-08-18
fix setting admin password for pulp

bbezak:
Add name to pulp container to fix setting admin password

this part is checking for name:
https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/wallaby/etc/kayobe/containers/pulp/post.yml#L13

error log:

`TASK [deploy-containers : Set the Pulp admin password] *********************************************************************************************************************************************************
fatal: [intern-os-seed-vm]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'name'\n\nThe error appears to be in '/home/stackhpc/wallaby-upgrade/src/kayobe-config/etc/kayobe/containers/pulp/post.yml': line 10, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set the Pulp admin password\n  ^ here\n"}`

### DIFF
--- a/etc/kayobe/containers/pulp/post.yml
+++ b/etc/kayobe/containers/pulp/post.yml
@@ -10,7 +10,7 @@
 - name: Set the Pulp admin password
   become: true
   command: >-
-    docker exec -u root {{ seed_containers.pulp.name }}
+    docker exec -u root pulp
     bash -c
       'pulpcore-manager reset-admin-password -p {{ pulp_password }}'
   no_log: true
--- a/etc/kayobe/seed.yml
+++ b/etc/kayobe/seed.yml
@@ -93,7 +93,6 @@ seed_pulp_container_enabled: true
 
 seed_pulp_container:
   pulp:
-    name: pulp
     image: pulp/pulp
     pre: "{{ kayobe_config_path }}/containers/pulp/pre.yml"
     post: "{{ kayobe_config_path }}/containers/pulp/post.yml"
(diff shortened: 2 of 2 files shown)


markgoddard (changes requested):
Could you instead fix the post task? The dict key defines the container's name.

bbezak:
> Could you instead fix the post task? The dict key defines the container's name.

I didn't find a elegant way to do it programatically, except that: `{{ seed_pulp_container.keys()[0] }}` 
However this looks like bike shedding way as this post task is only for that container which has hardcoded name to pulp anyway. Moreover this enumerating method it seems is unstable across python versions - https://stackoverflow.com/questions/5629023/order-of-keys-in-dictionaries-in-old-versions-of-python#:~:text=As%20of%20Python%203.6%2C%20for,across%20other%20implementations%20of%20Python

markgoddard:
> > Could you instead fix the post task? The dict key defines the container's name.
> 
> I didn't find a elegant way to do it programatically, except that: `{{ seed_pulp_container.keys()[0] }}` However this looks like bike shedding way as this post task is only for that container which has hardcoded name to pulp anyway. Moreover this enumerating method it seems is unstable across python versions - https://stackoverflow.com/questions/5629023/order-of-keys-in-dictionaries-in-old-versions-of-python#:~:text=As%20of%20Python%203.6%2C%20for,across%20other%20implementations%20of%20Python

Yes, it doesn't really work like that. There has to be some immutable way to reference the item in the dict/list, and in this case we are reusing that key as the container name. So hard coding is fine.

markgoddard (approved):

PR #138, opened 2022-08-19
Ubuntu CI

markgoddard:


markgoddard:
Got further - various tempest tests failed due to missing external network?

.github/workflows/stackhpc-pull-request.yml
```diff
@@ -45,10 +45,10 @@ jobs:
       neutron_plugin: ovs
       OS_CLOUD: sms-lab-release
     secrets: inherit
-    if: github.repository == 'stackhpc/stackhpc-kayobe-config'
+    if: false && github.repository == 'stackhpc/stackhpc-kayobe-config'
```
markgoddard: Need to test with the original condition

.github/workflows/stackhpc-pull-request.yml
```diff
@@ -57,4 +57,32 @@ jobs:
...
+      - build-kayobe-image
+    uses: ./.github/workflows/stackhpc-all-in-one.yml
+    with:
+      kayobe_image: ${{ needs.build-kayobe-image.outputs.kayobe_image }}
+      os_distribution: ubuntu
+      neutron_plugin: ovn
+      vm_image: Ubuntu-20.04-new
+      OS_CLOUD: sms-lab-release
+    secrets: inherit
+    if: false && github.repository == 'stackhpc/stackhpc-kayobe-config'
```
markgoddard: Need to remove `false &&`

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -158,10 +158,15 @@ jobs:
...
+          filter='${{ inputs.regexes }}'
+          if [[ ${{ inputs.seed }} == 'true' ]]; then
+            filter="$filter bifrost"
+          fi
           gh workflow run \
           container-sync.yml \
           --repo stackhpc/stackhpc-release-train \
-          --ref main
+          --ref main \
+          -f filter="$filter"
```
markgoddard: This is kind of unrelated, perhaps it should be pulled out

.github/workflows/stackhpc-pull-request.yml
```diff
@@ -57,4 +57,32 @@ jobs:
       neutron_plugin: ovn
       OS_CLOUD: sms-lab-release
     secrets: inherit
+    if: false && github.repository == 'stackhpc/stackhpc-kayobe-config'
```
markgoddard: ditto

terraform/aio/scripts/configure-local-networking.sh
```diff
@@ -9,50 +9,50 @@ cat << EOF | sudo tee -a /etc/hosts
 EOF
```
markgoddard: Apart from this part, this script is a noop. Will be interesting to see how CentOS copes with that.

terraform/aio/scripts/configure-local-networking.sh
```diff
@@ -9,50 +9,50 @@ cat << EOF | sudo tee -a /etc/hosts
...
 # Configure an IP on the 'public' network to allow access to/from the cloud.
-if ! sudo ip a show dev breth1 | grep $public_ip/24 >/dev/null 2>&1; then
-  sudo ip a add $public_ip/24 dev breth1
-fi
+#if ! sudo ip a show dev breth1 | grep $public_ip/24 >/dev/null 2>&1; then
+#  sudo ip a add $public_ip/24 dev breth1
+#fi
 
 # This prevents network.service from restarting correctly.
 sudo killall dhclient || true
```
markgoddard: and this

terraform/aio/scripts/configure-local-networking.sh
markgoddard: Actually we can drop this due to the userdata.cfg.tpl change

markgoddard: Oh, actually we've' dropped this file from the terraform config entirely.

markgoddard (commented):
This patch adds Rocky Linux aio jobs: https://github.com/stackhpc/stackhpc-kayobe-config/pull/194/files. It was based on these changes, but evolved a bit. Let's merge that one first, then rework anything here to use a similar approach.

.github/workflows/stackhpc-pull-request.yml
MoteHue: Tests pass on all four aios

markgoddard:
Rocky aio now merged

.github/workflows/stackhpc-container-image-build.yml
markgoddard: Pulled out into https://github.com/stackhpc/stackhpc-kayobe-config/pull/239

markgoddard:
I squashed the commits and pulled out the unrelated change. Should be good to go.

markgoddard (commented):
LGTM

PR #139, opened 2022-08-22
add new skydive tags

GregWhiteyBialas:
Skydive has a bug which prevents app from starting (https://review.opendev.org/c/openstack/kolla/+/853245). New build contains downgraded version.


markgoddard (approved):

mnasiadka:
> downgraded

downgraded? v0.28.0 is the latest version (previous versions had incompatibility with newer libc)

PR #140, opened 2022-08-23
Ubuntu: Stop using StackHPC host package repos, start using StackHPC container images (Wallaby)

markgoddard:
* The on_demand mirror in Ark does not work if the upstream mirror pulls packages (which it does sometimes).
* A set of Ubuntu images has been built with tag `wallaby-20220819T112725`.

### DIFF
--- a/etc/kayobe/apt.yml
+++ b/etc/kayobe/apt.yml
@@ -17,21 +17,13 @@
 # * filename: name of a file in /etc/apt/apt.conf.d/ in which to write the
 #   configuration
 # Default is an empty list.
-apt_config:
-  # NOTE: Currently the Pulp verbatim publisher does not sync translation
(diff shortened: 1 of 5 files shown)


mnasiadka (dismissed):
I don't think there's much value in snapshotting OS repo that does proper security updates.

mnasiadka:
> I don't think there's much value in snapshotting OS repo that does proper security updates.

Unless we're thinking about customer deployments that only have ark.stackhpc.com whitelisted.

markgoddard:
> > I don't think there's much value in snapshotting OS repo that does proper security updates.
> 
> Unless we're thinking about customer deployments that only have ark.stackhpc.com whitelisted.

The value is the stability & reproducibility.

PR #141, opened 2022-08-24
Support filtering Kolla container images to sync/publish

markgoddard:
It can take a long time to sync and publish all Kolla container images
to the local Pulp service. Sometimes we know we just need to update a
few images. This change adds support for filtering the image list, using
regexes in a similar way to kolla-build.

Example:

  kayobe playbook run ... -e stackhpc_pulp_images_kolla_filter='"^glance nova-compute$"'

### DIFF
--- a/etc/kayobe/pulp.yml
+++ b/etc/kayobe/pulp.yml
@@ -428,6 +428,15 @@ stackhpc_pulp_images_kolla:
   - redis-base
   - redis-sentinel
 
+# Whitespace-separated list of regular expressions matching Kolla image names.
+# Usage is similar to kolla-build CLI arguments.
+# Example:
+# kayobe playbook run ... -e stackhpc_pulp_images_kolla_filter='"^glance nova-compute$"'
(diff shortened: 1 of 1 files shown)


etc/kayobe/pulp.yml
```diff
@@ -428,6 +428,15 @@ stackhpc_pulp_images_kolla:
   - redis-base
   - redis-sentinel
 
+# Whitespace-separated list of regular expressions matching Kolla image names.
```
jovial, line 431: White space separated matches what we have in https://github.com/stackhpc/stackhpc-release-train/pull/121/files. Not related to this PR, as none of the image names contain anyway whitespace, but I had the sudden realisation that this means that the short names for the repositories can't have any whitespace in.

jovial (approved):

etc/kayobe/pulp.yml
markgoddard: That's correct.

PR #143, opened 2022-08-26
pulp: use pulp_url as advertised content origin

markgoddard:
This fixes an issue where a container image pull would fail if the
seed's hostname is not resolvable from the host pulling the image.
Typically it will not be.

This hasn't always been an issue, and it's unclear what has changed.
Perhaps something in the Pulp container registry code.


bbezak (approved):
Lgtm. Thx

PR #151, opened 2022-08-31
Bump nfv openvswitch repo to 20220824T113700

mnasiadka:


markgoddard (commented):
Is this for the OVN testing? Are we going to combine that with the general wallaby update testing?

mnasiadka:
> 

Right, SMS Lab is now running a stale-but-updated Wallaby snapshot, probably best would be to refresh that - but then we might not make it in those 10 days mentioned today.

markgoddard:
> > 
> 
> Right, SMS Lab is now running a stale-but-updated Wallaby snapshot, probably best would be to refresh that - but then we might not make it in those 10 days mentioned today.

Would it be better to build an image on the existing snapshot update PR? (#77)

markgoddard:
Either way, I don't think we should merge this yet - you can specify a branch when building an image.

mnasiadka:
> > Right, SMS Lab is now running a stale-but-updated Wallaby snapshot, probably best would be to refresh that - but then we might not make it in those 10 days mentioned today.
> 
> Would it be better to build an image on the existing snapshot update PR? (#77)

Yes, but needs to be rebased - it's two months old - I'll pursue it there.

PR #152, opened 2022-08-31
Adds Rocky 8 pulp repositories to dnf.yml

jovial:
This configures the hosts to use the mirrors in the local pulp. Prior to this change kayobe would attempt to configure
CentOS stream mirrors on a Rocky host.


etc/kayobe/dnf.yml
```diff
@@ -41,7 +41,27 @@
 #     file: myrepo
 #     gpgkey: http://gpgkey
 #     gpgcheck: yes
-dnf_custom_repos:
+dnf_custom_repos: "{{ dnf_custom_repos_el8 | combine(lookup('vars', 'dnf_custom_repos_' ~ os_distribution )) }}"
```
markgoddard: Might make sense to use `ansible_facts.distribution` here, if that works?

jovial: Tested this and we do have access to facts during the plays that configure the dnf repos.

markgoddard (approved):

markgoddard (approved):

markgoddard (approved):

PR #156, opened 2022-09-05
Create multinode env

jackhodgkiss:
Create `kayobe-config` environment called `ci-multinode`. These three commits represent the initial configuration of the environment. Majority of the content is copied from `ci-aio`. You can find the actual changes within;

- `etc/kayobe/environments/ci-multinode/inventory/hosts`
- `etc/kayobe/environments/ci-multinode/network-allocation.yml`
- `etc/kayobe/environments/ci-multinode/networks.yml`


etc/kayobe/environments/ci-multinode/inventory/hosts
```diff
@@ -0,0 +1,14 @@
+# A single all-in-one controller/compute host.
```
markgoddard: This file should be generated by terraform

etc/kayobe/environments/ci-multinode/network-allocation.yml
```diff
@@ -0,0 +1,7 @@
+admin_ips:
```
markgoddard: This file should be generated by terraform. Although I thought we also needed some other config for the admin network - CIDR, gateway, etc.

markgoddard (changes requested):

etc/kayobe/environments/ci-multinode/network-allocation.yml
jackhodgkiss: The admin network config is included in `tf-networks.yml` which will be included in the next PR.

markgoddard: Shouldn't that be generated by TF?

markgoddard (approved):

PR #160, opened 2022-09-15
Adds a templated dynamic overview dashboard

k-s-dean:


etc/kayobe/ansible/gendash.yml
```diff
@@ -0,0 +1,68 @@
+- hosts: localhost
+  vars: 
+    ansible_python_interpreter: /usr/bin/python3
+  tasks:
+    - name: Include the generated kolla.yml
+      include_vars:
+        file: ../../kolla/globals.yml
```
dougszumski, line 7: When you run this as a Kayobe custom playbook, you should have access to service flags without having to include this. Eg. `kolla_enable_foo` etc.

etc/kayobe/ansible/dashboard.overview.j2
```diff
@@ -0,0 +1,314 @@
...
+                     ],
+                     targets=[
+                     Target(
+                            expr='sum(haproxy_backend_up{backend="{{ service.backend }}"})',
+                            datasource="$datasource",
+                        ),
+                    ],
+                    gridPos=GridPos(h=3, w=8, x={{ counter.counter1}}, y={{ counter.counter2 }}),
+                    {{ increment1(counter, 'counter1') }}
+                    {% if (counter.counter1 == 24) %}
```
dougszumski, line 62: Does this depend on screen resolution? It would be nice if we could somehow rely on auto-layout, I guess that's not possible?

etc/kayobe/ansible/gendash.yml
```diff
@@ -0,0 +1,68 @@
...
+        name: services 
+
+    - name: 
+      vars:
+        openstack_services: 
+          - { name: "aodh", backend: "aodh_api_back",  enabled: "{{ services.enable_aodh | default('false') | bool }}" }
```
dougszumski, line 13: longer term, we could add support for templating an overview dashboard in KA, and we could pull this info from there so we don't have to main this list. Fine for now.

dougszumski (commented):
Please could you propose this to smslab kayobe config and apply it there?

It looks good overall - nice work. I'd like to see it live before adding here.

etc/kayobe/ansible/dashboard.overview.j2
k-s-dean: It does not depend on screen resolution. This sets the position of the panel within the row. if X is a smaller value than the previous panel it will place it underneath. The counter resets at 24, but depends on the size of the H and W of the panel. a Smaller panel will require a different value.

etc/kayobe/ansible/gendash.yml
k-s-dean: I'll double check this but im pretty sure that they weren't included.

etc/kayobe/ansible/dashboard.overview.j2
```diff
@@ -0,0 +1,314 @@
+{% set counter = { 'counter1': 0, 'counter2': 1 } %}
```
markgoddard, line 1: Maybe put this in `etc/kayobe/ansible/templates/dashboard.overview.j2`? It's already a bit cluttered in here

etc/kayobe/ansible/gendash.yml
```diff
@@ -0,0 +1,68 @@
...
+      set_fact: 
+        openstack_enabled_services: >-
+          {{ openstack_services | 
+             selectattr('enabled')
+          }}
+
+    - name: "Template out dashboard config" 
+      template:
+        src: ./dashboard.overview.j2
+        dest: ./dashboard.py
```
markgoddard, line 63: Could use ansible's tempfile module to get a path to write to. (Then remove it later)

etc/kayobe/ansible/gendash.yml
```diff
@@ -0,0 +1,68 @@
...
+      delegate_to: localhost
```
markgoddard, line 64: Already on localhost

etc/kayobe/ansible/gendash.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    - name:
```
markgoddard, line 66: Name me please

etc/kayobe/ansible/gendash.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    - name: "Template out dashboard config" 
+      template:
+        src: ./dashboard.overview.j2
+        dest: ./dashboard.py
+      delegate_to: localhost
+
+    - name: 
+      shell: >- 
+        generate-dashboard -o frontend.json dashboard.py
```
markgoddard, line 68: Is there a standard path in which to put the dashboard in kayobe-config?

requirements.txt
```diff
@@ -1 +1,2 @@
 git+https://github.com/stackhpc/kayobe@stackhpc/wallaby
+grafanalib # Dashboard templating library
```
markgoddard, line 2: Maybe clamp to a major/minor version?

etc/kayobe/ansible/gendash.yml
markgoddard: `s/services./kolla_/g`

markgoddard:
What's the status of this?

markgoddard:
I'll keep the branch around in case any one wants it.

PR #174, opened 2022-09-29
Xena: Ansible dependencies

cityofships:


mnasiadka (dismissed):

etc/kayobe/ansible/requirements.yml
```diff
@@ -1,8 +1,4 @@
 ---
 collections:
```
markgoddard, line 2: Why drop these versions? Using specific versions provides reproducibility.

markgoddard (changes requested):

markgoddard (approved):

etc/kayobe/ansible/requirements.yml
```diff
@@ -1,8 +1,6 @@
...
   - name: stackhpc.cephadm
-    version: 1.8.0
+    version: 1.10.0
   - name: stackhpc.pulp
     version: 0.3.0
-  - name: pulp.squeezer
-    version: 0.0.11
```
markgoddard, line 8: When I first added this it made sense - we need functionality from 0.0.11. Probably by now everyone has it already though. Even so, it sometimes makes sense to pin your transitive dependencies for the same reasons as your top-level ones. Let's see how we go without this one.

cityofships: > it sometimes makes sense to pin your transitive dependencies

Definitely - but shouldn't it rather happen in Pulp's collection dependencies? S-k-c doesn't make use of the squeezer module directly.

markgoddard: I can't remember for sure, but I think it's more about the parameters we pass in. stackhpc.pulp can just pass `omit` when they're not defined, so it's more version agnostic.

markgoddard (approved):

PR #175, opened 2022-09-29
Xena: Add system logging custom playbook

cityofships:
Works on CentOS and Ubuntu.

### DIFF
--- a/etc/kayobe/ansible/rsyslog.yml
+++ b/etc/kayobe/ansible/rsyslog.yml
@@ -1,0 +1,36 @@
+---
+# Custom playbook to configure rsyslog to forward operating system logs to
+# fluentd, eventually reaching Kolla's Elasticsearch cluster.
+#
+# Variables:
+# * rsyslog_group: Host pattern against which to target the playbook. Default
+#   is 'overcloud'.
+
+- name: Configure rsyslog to forward messages
+  hosts: "{{ rsyslog_group | default('overcloud') }}"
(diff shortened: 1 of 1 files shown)


etc/kayobe/hooks/overcloud-service-deploy/post.d/10-rsyslog.yml
```diff
@@ -0,0 +1 @@
+../../../ansible/rsyslog.yml
```
markgoddard: With other playbooks like this I've not added the hook symlink, thinking that not all deployments may want this, and they can opt in easily by adding the symlink.

etc/kayobe/ansible/rsyslog.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: Configure rsyslog to forward messages
+  hosts: overcloud
```
markgoddard: For the swap and growpart playbooks, we used this pattern:

```
hosts: "{{ swap_group | default('overcloud') }}"
```

etc/kayobe/ansible/rsyslog.yml
```diff
@@ -0,0 +1,29 @@
+---
```
markgoddard, line 1: Could you add a comment to explain what this playbook does?

markgoddard (changes requested):

markgoddard (approved):

PR #176, opened 2022-09-29
Xena: Stop dealing with 'base' type images

cityofships:
We really really never need these.


markgoddard:
> We really really never need these.

markgoddard (approved):

PR #179, opened 2022-10-05
Xena: Add basic monitoring & alerting stack

cityofships:
It's meant to be _minimal_ working configuration for all clouds. Draws some inspiration from https://github.com/stackhpc/stackhpc-kayobe-config/pull/18 but it's stripped down version.

### DIFF
--- a/etc/kayobe/environments/ci-aio/kolla/globals.yml
+++ b/etc/kayobe/environments/ci-aio/kolla/globals.yml
@@ -11,3 +11,6 @@ openstack_service_rpc_workers: "1"
 
 docker_yum_baseurl: "{{ stackhpc_repo_docker_url }}"
 docker_yum_gpgkey: "https://download.docker.com/linux/{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}/gpg"
+
+# Elasticsearch memory tuning
+es_heap_size: 1g
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -524,7 +524,7 @@ kolla_enable_grafana: true
 #kolla_enable_keystone:
 #kolla_enable_keystone_federation:
 #kolla_enable_keystone_horizon_policy_file:
-kolla_enable_kibana: true
+#kolla_enable_kibana:
 #kolla_enable_kuryr:
 #kolla_enable_loadbalancer:
 #kolla_enable_magnum:
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -456,7 +456,7 @@ kolla_build_args:
 #kolla_enable_ceilometer_horizon_policy_file:
 #kolla_enable_ceilometer_ipmi:
 #kolla_enable_cells:
-#kolla_enable_central_logging:
+kolla_enable_central_logging: true
 #kolla_enable_ceph_rgw:
 #kolla_enable_ceph_rgw_loadbalancer:
 #kolla_enable_cinder:
@@ -474,7 +474,7 @@ kolla_build_args:
 #kolla_enable_designate:
 #kolla_enable_destroy_images:
 #kolla_enable_elasticsearch:
-#kolla_enable_elasticsearch_curator:
+kolla_enable_elasticsearch_curator: true
 #kolla_enable_etcd:
 #kolla_enable_external_mariadb_load_balancer:
 #kolla_enable_fluentd:
@@ -484,7 +484,7 @@ kolla_build_args:
 #kolla_enable_glance_image_cache:
 #kolla_enable_gnocchi:
 #kolla_enable_gnocchi_statsd:
-#kolla_enable_grafana:
+kolla_enable_grafana: true
 #kolla_enable_hacluster:
 #kolla_enable_haproxy:
 #kolla_enable_haproxy_memcached:
@@ -524,7 +524,7 @@ kolla_build_args:
 #kolla_enable_keystone:
 #kolla_enable_keystone_federation:
 #kolla_enable_keystone_horizon_policy_file:
-#kolla_enable_kibana:
+kolla_enable_kibana: true
 #kolla_enable_kuryr:
 #kolla_enable_loadbalancer:
 #kolla_enable_magnum:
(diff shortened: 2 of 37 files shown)


cityofships:
CI will probably fail due to `prometheus-msteams` inventory group missing.

dougszumski:
> CI will probably fail due to `prometheus-msteams` inventory group missing.

@k-s-dean is working on a patch to add it

etc/kayobe/environments/ci-aio/kolla/globals.yml
```diff
@@ -11,3 +11,8 @@ openstack_service_rpc_workers: "1"
 
 docker_yum_baseurl: "{{ stackhpc_repo_docker_url }}"
 docker_yum_gpgkey: "https://download.docker.com/linux/{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}/gpg"
+
+# Elasticsearch memory tuning
+es_heap_size: 1g
+
+rabbitmq_monitoring_user: monitoring
```
dougszumski: Will this be inherited from the base setting? I.e we could remove it?

Also, I think it is a relic left over from the RabbitMQ exporter container. Now we use the RabbitMQ plugin, I think we can remove it entirely?

cityofships:
https://github.com/stackhpc/kayobe/pull/59

dougszumski (commented):
Thanks @cityofships, looks good, these are the most recent dashboards from sms-lab? A couple of minor  questions.

dougszumski:
> [stackhpc/kayobe#59](https://github.com/stackhpc/kayobe/pull/59)

Thanks, Kyle has made https://review.opendev.org/c/openstack/kayobe/+/860670

etc/kayobe/kolla.yml
```diff
@@ -524,7 +524,7 @@ kolla_build_args:
 #kolla_enable_keystone:
 #kolla_enable_keystone_federation:
 #kolla_enable_keystone_horizon_policy_file:
-#kolla_enable_kibana:
+kolla_enable_kibana: true
```
dougszumski: nit: You can leave this as the default since we enable central logging:
```
group_vars/all/kolla:kolla_enable_kibana: "{{ 'yes' if kolla_enable_central_logging | bool or kolla_enable_monasca | bool else 'no' }}"

```

etc/kayobe/environments/ci-aio/kolla/globals.yml
cityofships: Correct and correct; I copied this from original Will's patch

cityofships:
> Thanks @cityofships, looks good, these are the most recent dashboards from sms-lab? A couple of minor questions.

Hmm, let's review https://github.com/stackhpc/smslab-kayobe-config/pull/159 and update here if merged - the rest is up to date

dougszumski (dismissed):
Excellent, thanks @cityofships

cityofships:
Rebased

markgoddard:
Will try to review tomorrow, but immediate concern is that when using multiple environments the config in etc/kayobe/kolla/config/ won't get picked up.

etc/kayobe/kolla.yml
```diff
@@ -456,7 +456,7 @@ kolla_build_args:
 #kolla_enable_ceilometer_horizon_policy_file:
 #kolla_enable_ceilometer_ipmi:
 #kolla_enable_cells:
-#kolla_enable_central_logging:
+kolla_enable_central_logging: true
```
markgoddard, line 460: Undecided on whether we should set these flags by default, or just provide config & allow users to enable when they want to. I suppose we will want them enabled in more cases than not.

etc/kayobe/kolla/config/grafana/dashboards/ceph/ceph_mds.json
```diff
@@ -0,0 +1,316 @@
+{% raw %}
```
markgoddard, line 1: As mentioned in the previous comment, config files in `etc/kayobe/kolla/config` won't get pulled into kayobe environments. The solution/workaround for this has been to symlink them, but there is a [bug](https://review.opendev.org/c/openstack/kayobe/+/803882) associated with that that may need fixing it we are going to rely on that method. We'd also need clear docs on how to do it.

Alternatively we can merge and backport Will's patches to merge these configs. Unclear if we have time to do that though.

etc/kayobe/kolla/config/prometheus/blackbox.rules
```diff
@@ -0,0 +1,77 @@
+{% raw %}
```
markgoddard, line 3: We should really provide some attribution for these rules.

etc/kayobe/kolla/config/prometheus/cadvisor.rules
```diff
@@ -0,0 +1,54 @@
+# Taken from https://awesome-prometheus-alerts.grep.to/rules
```
markgoddard, line 1: Like this

etc/kayobe/kolla.yml
markgoddard: As discussed in infra R&D meeting, this will be correct in more cases than not, so let's keep it.

etc/kayobe/kolla/config/grafana/dashboards/ceph/ceph_mds.json
markgoddard: We can't do anything about this issue here, and it's good to have this config in place even if we need to symlink to it from an environment directory.

markgoddard:
@cityofships could we move this forward so that it's ready for the SMS lab upgrade?

cityofships:
@markgoddard Please merge if needed - the inline comments and/or README don't have an influence on the upgrade.

markgoddard (commented):
Nice to see the CI passing. One thing that could be improved is if the aio environment used the custom configuration that's being added here. Currently that requires symlinking it into the ci-aio environment.

markgoddard:
> Nice to see the CI passing. One thing that could be improved is if the aio environment used the custom configuration that's being added here. Currently that requires symlinking it into the ci-aio environment.

It would also be a good opportunity to test the documentation (which I assume still needs writing) on how to use the config.

markgoddard (approved):

PR #186, opened 2022-10-11
Configure IPA with useful inspection settings

priteau:
Add configuration to include extra-hardware and mellanox elements, with other useful collectors and kernel settings.

### DIFF
--- a/etc/kayobe/ipa.yml
+++ b/etc/kayobe/ipa.yml
@@ -29,7 +29,9 @@
 
 # List of additional Diskimage Builder (DIB) elements to use when building IPA
 # images. Default is none.
-#ipa_build_dib_elements_extra:
+ipa_build_dib_elements_extra:
+  - extra-hardware
+  - mellanox
 
 # List of Diskimage Builder (DIB) elements to use when building IPA images.
 # Default is combination of ipa_build_dib_elements_default and
@@ -115,7 +117,10 @@
 #ipa_collectors_default:
 
 # List of additional inspection collectors to run.
-#ipa_collectors_extra:
+ipa_collectors_extra:
+  - "dmi-decode"
+  - "extra-hardware"
+  - "numa-topology"
 
 # List of inspection collectors to run.
 #ipa_collectors:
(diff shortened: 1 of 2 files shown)


etc/kayobe/ipa.yml
```diff
@@ -5,7 +5,7 @@
 # Ironic Python Agent (IPA) image build configuration.
 
 # Whether to build IPA images from source.
-#ipa_build_images:
+ipa_build_images: True
```
markgoddard: This does change the workflow a bit - now you need to build an image, both for bifrost & overcloud ironic. I'd at least avoid putting this in wallaby.

My medium term goal is to build IPA images and put them on Ark, then just include config to pull them down in here. The build flag would go in the `ci-builder` environment.

markgoddard:
What should we do here?

priteau:
> What should we do here?

Are we still planning to build IPA images and host them on Ark?

markgoddard:
> > What should we do here?
> 
> Are we still planning to build IPA images and host them on Ark?

That is in the plan, although not scheduled in the near future.

priteau:
> > > What should we do here?
> > 
> > 
> > Are we still planning to build IPA images and host them on Ark?
> 
> That is in the plan, although not scheduled in the near future.

Thanks. I need to revisit this because IPA images on recent releases will use CentOS Stream 9.

etc/kayobe/ipa.yml
priteau: I removed `ipa_build_images` for now, it is up to the operator to enable it.

MoteHue (dismissed):
LGTM

etc/kayobe/ipa.yml
```diff
@@ -134,6 +139,10 @@
 
 # List of additional kernel parameters for Ironic python agent.
 #ipa_kernel_options_extra:
```
markgoddard: Need to uncomment?

markgoddard (changes requested):

etc/kayobe/ipa.yml
priteau: D'oh!

markgoddard (approved):

MoteHue (approved):

MoteHue (approved):

Alex-Welsh:
I've resolved the final conversation so this is now mergable, but it was last approved over a month ago. @priteau or @MoteHue could you quickly check it's not stale since then, and click the button?

MoteHue:
> I've resolved the final conversation so this is now mergable, but it was last approved over a month ago. @priteau or @MoteHue could you quickly check it's not stale since then, and click the button?

Still LGTM, let's merge :)

PR #194, opened 2022-10-14
Adds a Rocky 8 AIO test

jovial:


.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -151,10 +171,14 @@ jobs:
         run: |
           cat << EOF > etc/kayobe/environments/$KAYOBE_ENVIRONMENT/zz-aio-scenario.yml
           ---
+          os_distribution: ${{ env.OS_DISTRIBUTION }}
           kolla_enable_ovn: ${{ env.ENABLE_OVN }}
+          bootstrap_user: ${{ env.BOOTSTRAP_USER }}
```
jovial: This bit is unnecessary now that rocky and centos both use cloud-user. Will need to pull in: https://github.com/stackhpc/stackhpc-kayobe-config/pull/227

terraform/aio/templates/userdata.cfg.tpl
```diff
@@ -2,3 +2,20 @@
...
+runcmd:
+  - 'echo "10.0.0.34 pelican pelican.service.compute.sms-lab.cloud" >> /etc/hosts'
+  - 'echo "10.205.3.187 pulp-server pulp-server.internal.sms-cloud" >> /etc/hosts'
+# Configure SSH keys here, to avoid creating an ephemeral keypair.
+# This means only the instance needs to be cleaned up if the destroy fails.
+ssh_authorized_keys:
+ - ${ssh_public_key}
+
+write_files:
+  # WORKAROUND: https://bugs.launchpad.net/kolla-ansible/+bug/1995409
```
markgoddard, line 15: Not too ugly

terraform/aio/vm.tf
```diff
@@ -41,53 +46,49 @@ data "openstack_networking_subnet_v2" "network" {
...
+  user_data    = templatefile("templates/userdata.cfg.tpl", {ssh_public_key = file(var.ssh_public_key)})
   network {
     name = var.aio_vm_network
   }
 
   block_device {
     uuid                  = data.openstack_images_image_v2.image.id
     source_type           = "image"
-    volume_size           = 100
+    volume_size           = 200
```
markgoddard: wow

terraform/aio/vm.tf
```diff
@@ -41,53 +46,49 @@ data "openstack_networking_subnet_v2" "network" {
...
     boot_index            = 0
     destination_type      = "volume"
     delete_on_termination = true
   }
 
   provisioner "file" {
```
markgoddard: I ended up dropping this script in the Ubuntu PR, but haven't confirmed if it can also be dropped for CentOS/Rocky.

terraform/aio/vm.tf
jovial: Shall I drop this back down? I only needed this when I was testing the stackhpc lvm layout with default sizing.

markgoddard: Yeah, let's drop it back down

terraform/aio/vm.tf
```diff
@@ -16,10 +16,15 @@ variable "aio_vm_image" {
   default = "CentOS-stream8"
 }
 
-variable "aio_vm_keypair" {
+variable "aio_vm_user" {
```
markgoddard: Unused

terraform/aio/vm.tf
```diff
@@ -1,8 +1,8 @@
-variable "ssh_private_key" {
+variable "ssh_public_key" {
   type = string
 }
 
-variable "ssh_public_key" {
+variable "ssh_private_key" {
```
markgoddard: Can be dropped now I think

jovial: Cheers, done.

terraform/aio/vm.tf
jovial: Good spot, thanks. Just to mention I'm not 100% sure that removing configure-local-networking will work, but thought I'd try it again now that the CI was passing. Will revert if it fails.

.github/workflows/stackhpc-all-in-one.yml
jovial: Removed in https://github.com/stackhpc/stackhpc-kayobe-config/pull/194/commits/72e96ff7cb4ab560ffbda2e786488117a06926b8 and https://github.com/stackhpc/stackhpc-kayobe-config/pull/194/commits/8b906b3a0c0201396adf9bfce245ec20da20be3e

markgoddard (approved):

PR #195, opened 2022-10-14
Rocky 8: Generate a more complete overcloud image

jovial:
The standard rocky container is very bare-bones and is missing many useful utilities. The container-generic image is supposed to be more like the generic cloud image.


jovial:
Hmm, only in xena

PR #196, opened 2022-10-14
Rocky 8: Generate a more complete overcloud image

jovial:
The standard rocky container is very bare-bones and is missing many useful utilities. The container-generic image is supposed to be more like the generic cloud image.

### DIFF
--- a/etc/kayobe/overcloud-dib.yml
+++ b/etc/kayobe/overcloud-dib.yml
@@ -71,7 +71,8 @@ overcloud_dib_host_packages_extra:
 overcloud_dib_git_elements_extra:
   - repo: "https://github.com/stackhpc/stackhpc-image-elements"
     local: "{{ source_checkout_path }}/stackhpc-image-elememts"
-    version: "v1.4.0"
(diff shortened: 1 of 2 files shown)


etc/kayobe/overcloud-dib.yml
```diff
@@ -71,7 +71,8 @@ overcloud_dib_host_packages_extra:
 overcloud_dib_git_elements_extra:
   - repo: "https://github.com/stackhpc/stackhpc-image-elements"
     local: "{{ source_checkout_path }}/stackhpc-image-elememts"
-    version: "v1.4.0"
+    # FIXME: merge and tag a new release
+    version: "feature/rocky-container-generic"
```
jovial, line 75: FIX this before merging. Just proposing the change for consideration.

jovial: We are using cloud-user for the the centos images, so rocky 8 should match.

jovial:
Was one year too early with this one, see : https://github.com/stackhpc/stackhpc-kayobe-config/pull/721

PR #207, opened 2022-10-24
changing release to point to latest build

grzegorzkoper:


etc/kayobe/kolla.yml
```diff
@@ -96,7 +96,7 @@ kolla_docker_registry_password: "{{ stackhpc_docker_registry_password }}"
 
 # Kolla OpenStack release version. This should be a Docker image tag.
 # Default is {{ openstack_release }}.
-kolla_openstack_release: xena-20220728T100435
+kolla_openstack_release: xena-20221010T103511
```
markgoddard, line 99: Need to drop the overrides in etc/kayobe/kolla/globals.yml

markgoddard (approved):

PR #222, opened 2022-11-10
Rephrase the match logic for interfaces monitored for package drops

oneswig:
OVS bridge interfaces drop packets during normal operation, generating lots of spurious alerts.  Change the regex to filter out interfaces that don't matter for packet drops.

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/ceph.rules
+++ b/etc/kayobe/kolla/config/prometheus/ceph.rules
@@ -154,7 +154,7 @@ groups:
 
   # alert on nic packet errors and drops rates > 1 packet/s
   - alert: NetworkPacketsDropped
-    expr: irate(node_network_receive_drop_total{device=~"en.*|eth.*"}[5m]) + irate(node_network_transmit_drop_total{device=~"en.*|eth.*"}[5m]) > 1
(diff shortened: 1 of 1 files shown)


dougszumski (approved):
This should help reduce alert spam, thanks Stig.

etc/kayobe/kolla/config/prometheus/ceph.rules
```diff
@@ -154,7 +154,7 @@ groups:
 
   # alert on nic packet errors and drops rates > 1 packet/s
   - alert: NetworkPacketsDropped
-    expr: irate(node_network_receive_drop_total{device=~"en.*|eth.*"}[5m]) + irate(node_network_transmit_drop_total{device=~"en.*|eth.*"}[5m]) > 1
+    expr: irate(node_network_receive_drop_total{device!~"lo|br.*|.*-ovs|tap.*"}[5m]) + irate(node_network_transmit_drop_total{device!~"lo|br.*|.*-ovs|tap.*"}[5m]) > 1
```
cityofships, line 157: This will catch, among others:
```
bond.*
docker0
genev_sys_6081
ovs-system
p-brbond.*
qbr.*
qvb.*
vxlan_sys_4789
```
The previous regex has already been filtering out OVS bridges - I thought we were only expanding it with `|bond.*`.

Does by the way `|ib.*` make sense too?

markgoddard: Perhaps this needs to be addressed. Could be a follow up?

markgoddard:
Backported to Xena: https://github.com/stackhpc/stackhpc-kayobe-config/pull/233

PR #226, opened 2022-11-14
Make prometheus datasource configurable.

jovial:
The use case is when you have a single grafana showing metrics from multiple clouds. This allows you to toggle between the clouds by selecting the datasource.


markgoddard:
This will get synced into yoga in https://github.com/stackhpc/stackhpc-kayobe-config/pull/254

PR #227, opened 2022-11-15
CI: use cloud-user as bootstrap user on CentOS

markgoddard:
cloud-init was changed to make CentOS more similar to RHEL, using a
cloud-user account rather than centos.

https://github.com/canonical/cloud-init/pull/1639

### DIFF
--- a/etc/kayobe/environments/ci-aio/controllers.yml
+++ b/etc/kayobe/environments/ci-aio/controllers.yml
@@ -1,0 +1,7 @@
+---
+###############################################################################
+# Controller node configuration.
+
+# User with which to access the controllers via SSH during bootstrap, in order
+# to setup the Kayobe user account. Default is {{ os_distribution }}.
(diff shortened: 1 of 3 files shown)


jovial:
Terraform also has references to the centos user: https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/yoga/terraform/aio/vm.tf#L88 :see_no_evil:

markgoddard:
> Terraform also has references to the centos user: https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/yoga/terraform/aio/vm.tf#L88 see_no_evil

Right you are

jovial (commented):
Seems to be working. Did we want to set the target branch to wallaby and then sync to later releases? Just notice the branch name is `wallaby-cloud-user`.

markgoddard:
> Seems to be working. Did we want to set the target branch to wallaby and then sync to later releases? Just notice the branch name is `wallaby-cloud-user`.

argh, forgot the default has changed

jovial (approved):

PR #233, opened 2022-11-16
Rephrase the match logic for interfaces monitored for package drops

markgoddard:
OVS bridge interfaces drop packets during normal operation.  Change
the regex to filter out interfaces that don't matter for packet
drops.

(cherry picked from commit 9c3f15a1f374d24d08a0bdac63886220247d2a4e)


markgoddard:
This made it in via a branch merge.

PR #239, opened 2022-11-22
CI: add bifrost to container sync filter

markgoddard:
Without this, if an image build regex is provided then bifrost images
would not also be synced.


jovial (approved):
LGTM

PR #240, opened 2022-11-22
feat: bump stackhp.pulp from `0.3.0` ~> `0.4.0`

jackhodgkiss:


markgoddard:
Can you propose it to stackhpc/wallaby first, then we can merge forward from there

PR #246, opened 2022-11-24
Multinode Wallaby

jackhodgkiss:
- Fix `_bootstrap_user` for all groups to use `cloud-user`
- Use pulp proxy to avoid slow downloads
- Include `seed` within `growroot_group` as `seed` uses LVM images
- Enable Octavia backed by OVN
- Bump cephadm
- Reduce `MTU` on `external_network` to enable communication with VMs on `admin-vxlan`


etc/kayobe/environments/ci-multinode/networks.yml
```diff
@@ -81,7 +81,7 @@ internal_vlan: 101
 
 # External network
 external_cidr: 192.168.38.0/24
-external_mtu: 1400
+external_mtu: 1350
```
markgoddard, line 84: I think @sd109 hit this with his aufn work, and possibly @darmach with his magnum work. It's because we're using VXLAN networks with OVN, but internally it sometimes uses GENEVE tunnels which have a larger header size.

One option is to switch to geneve tenant networks when using OVN, which generally work better.

etc/kayobe/environments/ci-multinode/kolla/globals.yml
```diff
@@ -14,3 +14,8 @@ nova_backend_ceph: "yes"
 # list, and add the config that would be generated from it here.
 neutron_bridge_name: "{{ vxlan_interfaces[0].device }}-ovs"
 neutron_external_interface: "{{ vxlan_interfaces[0].device }}"
+
+# Octavia load balancer configuration
+octavia_auto_configure: "no"
+octavia_provider_drivers: "ovn:OVN provider"
+octavia_provider_agents: "ovn"
```
markgoddard, line 21: We probably only want this when kolla_enable_ovn is true.

etc/kayobe/environments/ci-multinode/kolla.yml
```diff
@@ -2,3 +2,5 @@
 kolla_enable_cinder: true
 kolla_enable_cinder_backup: true
 kolla_enable_neutron_provider_networks: true
+kolla_enable_ovn: true
```
markgoddard, line 5: Unsure whether we should change this default. Ideally it should be possible & easy to use either OVS or OVN.

etc/kayobe/environments/ci-multinode/kolla.yml
```diff
@@ -2,3 +2,5 @@
...
+kolla_enable_octavia: true
```
markgoddard, line 6: Perhaps only enable octavia when OVN is enabled, since we don't have the config for the amphora provider in place currently.

markgoddard (changes requested):

markgoddard (commented):
Do you need to reduce TEMPEST_CONCURRENCY as you did in xena?

markgoddard (commented):
```
 [dashboard]
-dashboard_url = https://192.168.39.2
+dashboard_url = http://192.168.39.2
```

markgoddard:
You'll probably also want to merge in stackhpc/wallaby, maybe in a separate PR.

markgoddard (approved):

PR #247, opened 2022-11-24
Switch to using absolute limit for low memory alert

jovial:
Nodes with alot of ram can end up triggering this low memory alert even when they have a significant amount of free memory e.g for a node with 512GiB ram, when we hit the current alert (which is configured to be 80%) we still have 102.5GiB free.

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/system.rules
+++ b/etc/kayobe/kolla/config/prometheus/system.rules
@@ -16,13 +16,13 @@ groups:
       description: "{{ $labels.device }} is {{ $value }}% full."
 
   - alert: LowMemory
-    expr: ( ( node_memory_MemTotal_bytes - node_memory_MemAvailable_bytes - (node_memory_Hugepagesize_bytes * node_memory_HugePages_Total)) / (node_memory_MemTotal_bytes - (node_memory_Hugepagesize_bytes * node_memory_HugePages_Total))) * 100 >= 80
+    expr: (node_memory_MemAvailable_bytes / 1024^3) < {{ alertmanager_low_memory_threshold_gb }}
     for: 1m
     labels:
       severity: alert
     annotations:
       summary: "Prometheus exporter at {{ $labels.instance }} reports low memory"
-      description: "Memory is {{ $value }}% full."
+      description: "Available memory is {{ $value }}."
 
   - alert: HostOomKillDetected
     expr: increase(node_vmstat_oom_kill[5m]) > 0
--- a/etc/kayobe/kolla/config/prometheus/system.rules
+++ b/etc/kayobe/kolla/config/prometheus/system.rules
@@ -16,7 +16,7 @@ groups:
       description: "{{ $labels.device }} is {{ $value }}% full."
 
   - alert: LowMemory
-    expr: (node_memory_MemAvailable_bytes / 1024^3) < {{ alertmanager_low_memory_threshold_gb }}
+    expr: (node_memory_MemAvailable_bytes / 1024^3) < {{ alertmanager_low_memory_threshold_gib }}
     for: 1m
     labels:
       severity: alert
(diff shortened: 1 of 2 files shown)


priteau (changes requested):
In commit message: alot -> a lot

Add new line at end of globals.yml

etc/kayobe/kolla/globals.yml
```diff
@@ -7,6 +7,14 @@ docker_yum_gpgkey: "https://download.docker.com/linux/centos/gpg"
 bifrost_tag: xena-20221102T110017
 {% endif %}
 
+#############################################################################
 # Monitoring and alerting related settings
+
 es_heap_size: 8g
 prometheus_cmdline_extras: "--storage.tsdb.retention.time=30d"
+# Threshold to trigger a LowMemory alert in Gibibytes (GiB). When the
+# amount of free memory is lower that this value an alert will
```
priteau: that -> than

etc/kayobe/kolla/globals.yml
```diff
@@ -7,6 +7,14 @@ docker_yum_gpgkey: "https://download.docker.com/linux/centos/gpg"
...
+# be triggered.
+alertmanager_low_memory_threshold_gb: 5
```
priteau: Should we use gib instead of gb?

priteau (changes requested):

etc/kayobe/kolla/config/prometheus/system.rules
```diff
@@ -16,13 +16,13 @@ groups:
       description: "{{ $labels.device }} is {{ $value }}% full."
 
   - alert: LowMemory
-    expr: ( ( node_memory_MemTotal_bytes - node_memory_MemAvailable_bytes - (node_memory_Hugepagesize_bytes * node_memory_HugePages_Total)) / (node_memory_MemTotal_bytes - (node_memory_Hugepagesize_bytes * node_memory_HugePages_Total))) * 100 >= 80
+    expr: (node_memory_MemAvailable_bytes / 1024^3) < {{ alertmanager_low_memory_threshold_gb }}
```
jovial:
```suggestion
    expr: (node_memory_MemAvailable_bytes / 1024^3) < {{ alertmanager_low_memory_threshold_gib }}
```

etc/kayobe/kolla/globals.yml
jovial:
```suggestion
# amount of free memory is lower than this value an alert will
```

etc/kayobe/kolla/globals.yml
jovial:
```suggestion
alertmanager_low_memory_threshold_gib: 5
```

etc/kayobe/kolla/globals.yml
```diff
@@ -7,6 +7,14 @@ docker_yum_gpgkey: "https://download.docker.com/linux/centos/gpg"
...
+
+#############################################################################
```
jovial:
```suggestion
#############################################################################

```

etc/kayobe/kolla/globals.yml
jovial: Done, thanks.

etc/kayobe/kolla/globals.yml
jovial: Good idea, I've changed it.

dougszumski (approved):
Good change. Unrelated CI fail. Would be helpful to squash on merge.

markgoddard (approved):

bbezak:
we're using RAW statement in the beginning of the file, therefore this change is not enough

`ts=2022-11-29T13:20:50.654Z caller=manager.go:973 level=error component="rule manager" msg="loading groups failed" err="/etc/prometheus/system.rules: 19:11: group \"Node\", rule 2, \"LowMemory\": could not parse expression: 1:46: parse error: unexpected left brace '{'"`

bbezak:
we could probably add endraw/raw between this new variable @jovial

but then, it should probably be a kayobe var, not k-a

PR #252, opened 2022-11-28
Move README content into documentation

markgoddard:


jovial (approved):
Have built locally. Content looks good. This is how it looks:
image

@technowhizz is hoping this lands so he can base his docs change on top.

PR #260, opened 2022-11-29
Fix templating issue

jovial:
Kolla doesn't template these rules files:

https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/prometheus/tasks/config.yml#L42-L56


markgoddard (approved):

bbezak:
thx

PR #262, opened 2022-11-29
Switch from certificates to RBAC

Alex-Welsh:


etc/kayobe/ansible/requirements.yml
```diff
@@ -3,6 +3,8 @@ collections:
   - name: stackhpc.cephadm
     version: 1.10.0
   - name: stackhpc.pulp
-    version: 0.3.0
+    version: 0.4.0
```
markgoddard: Do we need a minimum version of squeezer here?

etc/kayobe/ansible/requirements.yml
```diff
@@ -3,6 +3,8 @@ collections:
...
+  - name: community.crypto
+    version: 2.0.2
```
markgoddard: Where does this requirement come from?

doc/source/configuration/release-train.rst
```diff
@@ -153,3 +148,30 @@ with the push repository using the pulp CLI:
...
+        name: centos-stream-8-baseos-development
+        policy: on_demand
+        proxy_url: __omit_place_holder__d35452c39719f081229941a64fd2cdce1188a287
+        remote_password: <password>
+        remote_username: <username>
+        required: true
+        state: present
+        sync_policy: mirror_complete
+        url: https://ark.stackhpc.com/pulp/content/centos/8-stream/BaseOS/x86_64/os/development
+      msg: Task failed to complete. (failed; 404, message='Not Found', url=URL('https://ark.stackhpc.com/pulp/content/centos/8-stream/BaseOS/x86_64/os/development')) '''
```
markgoddard: nit: development would normally be a datetime stamp

etc/kayobe/ansible/requirements.yml
markgoddard: (an existing installation would not otherwise get updated)

etc/kayobe/pulp.yml
```diff
@@ -23,11 +23,6 @@ stackhpc_release_pulp_url: "https://ark.stackhpc.com"
 stackhpc_release_pulp_username:
 stackhpc_release_pulp_password:
 
-# Client certificates used to access StackHPC Ark repositories.
-# They are trusted by the 'release' cert guard's CA.
-stackhpc_release_pulp_client_cert: "{{ lookup('file', kayobe_config_path ~ '/ansible/certs/ark.stackhpc.com/client-cert.pem') | trim }}"
-stackhpc_release_pulp_client_key: "{{ lookup('file', kayobe_config_path ~ '/ansible/certs/ark.stackhpc.com/client-key.pem') | trim }}"
```
markgoddard, line 29: Could you remove the placeholder files?

etc/kayobe/ansible/certs/ark.stackhpc.com/client-*.pem

markgoddard (changes requested):
We're going to need this change in each of our supported branches, victoria through to yoga. We could push to victoria first then merge up through the branches, although some things are different - README moved to docs in xena, repos change between releases etc.

etc/kayobe/ansible/requirements.yml
Alex-Welsh: stackhpc.pulp has the latest version of squeezer as a dependency in its galaxy.yml. Would that not force an update if an older version already existed?

etc/kayobe/ansible/requirements.yml
Alex-Welsh: I think I added this by mistake when pasting in the new stackhpc.pulp version

etc/kayobe/pulp.yml
Alex-Welsh: Do you want to keep the directory?

etc/kayobe/ansible/requirements.yml
Alex-Welsh: Added squeezer as a dependency

etc/kayobe/pulp.yml
markgoddard: No - git won't track it if the files are removed.

etc/kayobe/ansible/requirements.yml
```diff
@@ -3,6 +3,8 @@ collections:
   - name: stackhpc.cephadm
     version: 1.10.0
   - name: stackhpc.pulp
-    version: 0.3.0
+    version: 0.4.1
+  - name: pulp.squeezer
+    version: 0.0.13
```
markgoddard: Drop squeezer

markgoddard (approved):

PR #263, opened 2022-11-29
DNM: prometheus & grafana in ci-aio

markgoddard:
- CI: symlink to prometheus & grafana config in ci-aio
- DNM: disable most aio CI jobs


cityofships:
Looks good:

- Alert rules loaded

Screenshot 2022-11-30 at 13 16 33

- Targets added

Screenshot 2022-11-30 at 13 16 48

- Dashboards loaded
Screenshot 2022-11-30 at 13 18 16

PR #268, opened 2022-12-01
Fix oom-killer graph

technowhizz:
Changes the oom-killer graph from a smoothed irate to a discrete delta function.

This is because the Irate would smooth out the event and would make it difficult to notice


dougszumski (approved):
I've seen this working, LGTM.

PR #272, opened 2022-12-02
Use intermediate var for Rocky Powertools repo

m-bull:


jovial:
Think we need this in wallaby too

priteau (approved):

m-bull:
Should we abandon this in favour of https://github.com/stackhpc/stackhpc-kayobe-config/pull/273 @jovial ?

bbezak:
> Should we abandon this in favour of #273 @jovial ?
looks like a good idea. #273 is for wallaby though, so we would need to sync it to yoga as well

markgoddard:
> > Should we abandon this in favour of #273 @jovial ?
> > looks like a good idea. #273 is for wallaby though, so we would need to sync it to yoga as well

https://github.com/stackhpc/stackhpc-kayobe-config/pull/277
https://github.com/stackhpc/stackhpc-kayobe-config/pull/278

m-bull:
Thanks @markgoddard

PR #241, opened 2022-11-22
feat: bump stackhp.pulp from `0.3.0` ~> `0.4.0`

jackhodgkiss:

PR #277, opened 2022-12-05
xena: wallaby merge

markgoddard:
- Sync kayobe-config with kayobe changes
- playbook for installing and running cardiff
- Add regex pattern, add line to eof
- Review changes
- packages moved to requirements file
- Change name from cardiff to ADVise
- Separate processing from visualisation
- feat: bump stackhp.pulp from `0.3.0` ~> `0.4.0` (#241)
- Fixes for Rocky 8 pulp snapshots

### DIFF
--- a/etc/kayobe/ansible/requirements.yml
+++ b/etc/kayobe/ansible/requirements.yml
@@ -1,6 +1,4 @@
 ---
 collections:
   - name: stackhpc.pulp
-    version: 0.4.0
-  - name: pulp.squeezer
-    version: 0.0.11
+    version: 0.4.1
--- a/README.rst
+++ b/README.rst
@@ -13,7 +13,7 @@ StackHPC provides packages and container images for OpenStack via `Ark
 
 Deployments should use a local `Pulp <https://pulpproject.org/>`__ repository
 server to synchronise content from Ark and serve it locally. Access to the
-repositories on Ark is controlled via X.509 certificates issued by StackHPC.
+repositories on Ark is controlled via user accounts issued by StackHPC.
 
 This configuration is a base, and should be merged with any existing Kayobe
 configuration. It currently provides the following:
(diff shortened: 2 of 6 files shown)


etc/kayobe/stackhpc.yml
```diff
@@ -104,15 +104,15 @@ stackhpc_repo_treasuredata_4_url: "{{ stackhpc_repo_mirror_url }}/pulp/content/t
 stackhpc_repo_treasuredata_4_version: "{{ stackhpc_repo_distribution }}"
 
 # Rocky 8 BaseOS
-stackhpc_repo_rocky_baseos_url: "{{ stackhpc_repo_mirror_url }}/pulp/content/rocky/8.6/BaseOS/x86_64/os/{{ stackhpc_repo_rocky_baseos_version }}"
+stackhpc_repo_rocky_baseos_url: "{{ stackhpc_repo_mirror_url }}/pulp/content/rocky/8/BaseOS/x86_64/os/{{ stackhpc_repo_rocky_baseos_version }}/"
```
m-bull: Think this is where the CI trips up:

`s|rocky/8|rocky/8.{{ stackhpc_pulp_repo_rocky_8_minor_version }}`

markgoddard: Does that need to go back to wallaby?

m-bull: In CI terms yes, I think so - that repo has always been available at 8.x rather than 8. Other clients may have a different experience depending on their local configuration...

markgoddard: This should all be pretty standard shouldn't it? Or is the ci-aio environment doing something different than a client would?

m-bull: Yeah, I agree, this should all be standard - I was just trying to think of how it might've slipped in there. On reflection it probably came from a propagation of the `centos/8/` style URLs above.

etc/kayobe/ansible/requirements.yml
```diff
@@ -3,6 +3,6 @@ collections:
   - name: stackhpc.cephadm
     version: 1.10.0
   - name: stackhpc.pulp
-    version: 0.3.0
+    version: 0.4.0
```
Alex-Welsh: 0.4.1 is now the latest

jovial (approved):

PR #279, opened 2022-12-05
Replace hardcoded threshold by temp_max value

priteau:
The temp_max value is dynamically gathered from the device [1]. With Xeon CPUs (coretemp driver), it is often 90C, but sometimes lower.

This can help reduce the frequency of alerts with busy hypervisors.

[1] https://docs.kernel.org/hwmon/coretemp.html

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/system.rules
+++ b/etc/kayobe/kolla/config/prometheus/system.rules
@@ -34,7 +34,7 @@ groups:
       description: "OOM kill detected"
 
   - alert: Overheating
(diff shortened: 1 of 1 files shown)


markgoddard:
In general it's better to push config changes into the earliest branch first, then merge through to later branches.

markgoddard (dismissed):

priteau:
I rebased on `stackhpc/xena`, since the wallaby branch doesn't include these alerts.

markgoddard (approved):

PR #283, opened 2022-12-06
Include Rocky minor version in repo URL

m-bull:


m-bull:
Some discussion here: https://github.com/stackhpc/stackhpc-kayobe-config/pull/277#discussion_r1039538771

etc/kayobe/stackhpc.yml
```diff
@@ -96,15 +96,15 @@ stackhpc_repo_treasuredata_4_url: "{{ stackhpc_repo_mirror_url }}/pulp/content/t
 stackhpc_repo_treasuredata_4_version: "{{ stackhpc_repo_distribution }}"
 
 # Rocky 8 BaseOS
-stackhpc_repo_rocky_baseos_url: "{{ stackhpc_repo_mirror_url }}/pulp/content/rocky/8/BaseOS/x86_64/os/{{ stackhpc_repo_rocky_baseos_version }}/"
+stackhpc_repo_rocky_baseos_url: "{{ stackhpc_repo_mirror_url }}/pulp/content/rocky/8.{{ stackhpc_pulp_repo_rocky_8_minor_version }}/BaseOS/x86_64/os/{{ stackhpc_repo_rocky_baseos_version }}/"
```
markgoddard: Based on `base_path` in etc/kayobe/pulp.yml, I think just 8 is correct. This was done to avoid needing to update repo files for minor version bumps.

markgoddard (changes requested):

etc/kayobe/stackhpc.yml
markgoddard: I suppose the problem is that test pulp on SMS lab does not follow this pattern - it has a different base_path for each minor version.

We could override the rocky URL variables in etc/kayobe/environments/ci-aio/stackhpc-ci.yml

m-bull: Yeah - all good points, I got very confused!

Overrode Rocky URL vars in `stackhpc-ci.yml` as suggested.

markgoddard (approved):

PR #286, opened 2022-12-06
Hammer playbook for rabbitmq

MoteHue:


mnasiadka (commented):
Wouldn't it be better to push that upstream to Kolla-Ansible?

markgoddard (approved):

markgoddard:
> Wouldn't it be better to push that upstream to Kolla-Ansible?

Perhaps. It probably wouldn't get accepted in this state though. Let's have it in here for now.

PR #288, opened 2022-12-08
Adds CIS benchmark hardening

jovial:


etc/kayobe/ansible/requirements.yml
```diff
@@ -6,3 +6,6 @@ collections:
     version: 0.4.1
 roles:
   - src: stackhpc.vxlan
+  - name: ansible-lockdown.rhel8_cis
+    src: https://github.com/ansible-lockdown/RHEL8-CIS
+    version: 1.3.0
```
markgoddard, line 11: This is from July 2021, would be nice to update to 2.1.0 when there is time to validate it.

markgoddard (approved):

etc/kayobe/ansible/requirements.yml
mnasiadka: Newer versions do not support CentOS Stream anymore, but it would be good to have a newer version on Ubuntu/Rocky.

markgoddard: Doh. Probably something to update in zed/Antelope then.

priteau: Note that it doesn't support `inject_facts_as_vars = False`.

markgoddard: hmm. The playbook should at least have a comment to that effect. I suggested to @GregWhiteyBialas to use the [config lookup](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/config_lookup.html) plugin to check that the option is true for the Wazuh playbook.

PR #290, opened 2022-12-08
Enable running of additional ceph commands before cephadm deployment

sd109:


mnasiadka:
Why before? cephadm shell won't work before deploy...

markgoddard:
> Why before? cephadm shell won't work before deploy...

Yeah, should go just after cephadm-deploy.yml

sd109:
Well spotted, thanks guys. My fault for trying to cut corners...

markgoddard (approved):

PR #291, opened 2022-12-09
Bypass pulp on seed hypervisor and seed vm

sd109:
This avoids the problem of trying to install packages from the local pulp server on the seed VM before that server exists.

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -1,7 +1,11 @@
 # yamllint disable-file
 ---
-docker_yum_baseurl: "{{ stackhpc_repo_docker_url }}"
-docker_yum_gpgkey: "https://download.docker.com/linux/centos/gpg"
+
+# To work around issue of trying to install docker from 
+# empty pulp server, use upstream docker dnf repo on 
+# non-overcloud hosts
+enable_docker_repo: {% raw %}"{{ 'overcloud' not in group_names }}"{% endraw %}
+
 
 {% if kolla_base_distro == 'centos' %}
 bifrost_tag: xena-20221128T101757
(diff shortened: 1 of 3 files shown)


priteau (commented):
This skips using Pulp for infra VMs too. Just pointing it out in case this is not what we want to do.

markgoddard:
> This skips using Pulp for infra VMs too. Just pointing it out in case this is not what we want to do.

That's true. I don't know if we have any assumption of ordering between the seed VM and infra VMs, but a user could easily add the config if necessary. The release train docs page needs filling out a bit, I can add something about this.

priteau (approved):

markgoddard:
Verified it's doing the right thing from the logs.

etc/kayobe/kolla/globals.yml
```diff
@@ -1,7 +1,11 @@
 # yamllint disable-file
 ---
-docker_yum_baseurl: "{{ stackhpc_repo_docker_url }}"
-docker_yum_gpgkey: "https://download.docker.com/linux/centos/gpg"
+
+# To work around issue of trying to install docker from 
+# empty pulp server, use upstream docker dnf repo on 
+# non-overcloud hosts
+enable_docker_repo: {% raw %}"{{ 'overcloud' not in group_names }}"{% endraw %}
```
markgoddard, line 7: Oops, looks like this broke Ubuntu: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/3677794367/jobs/6220373647

Perhaps it also needs to be enabled when `ansible_facts.os_family == 'Debian'`

markgoddard: (This is because we don't use release train for host packages on Ubuntu)

markgoddard: https://github.com/stackhpc/stackhpc-kayobe-config/pull/295

markgoddard: Merged

PR #295, opened 2022-12-12
Ubuntu: always use kolla-ansible to install docker-ce repo

markgoddard:
Currently Release Train does not provide Ubuntu package repos.


markgoddard:
Passed in https://github.com/stackhpc/stackhpc-kayobe-config/pull/293

PR #297, opened 2022-12-13
Sets kolla_docker_registry_insecure to true

jovial:
We currently don't configure TLS for the the local pulp registry.  This adds the pulp server to the list of insecure-registries, so that we can pull images.

### DIFF
--- a/etc/kayobe/environments/ci-aio/stackhpc-ci.yml
+++ b/etc/kayobe/environments/ci-aio/stackhpc-ci.yml
@@ -5,11 +5,6 @@
 # Docker namespace to use for Kolla images. Default is 'kolla'.
 kolla_docker_namespace: stackhpc-dev
 
-# Whether docker should be configured to use an insecure registry for Kolla
-# images. Default is false, unless docker_registry_enabled is true and
-# docker_registry_enable_tls is false.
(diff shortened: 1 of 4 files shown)


markgoddard:
We've set this in the ci-* environments, so it could be dropped there now.

jovial:
> We've set this in the ci-* environments, so it could be dropped there now.

Done. It did make me think we could have a better default rather than hard-coding `true`, but kolla_docker_registry has the protocol stripped, so you'd have to look at one of the intermediate variables e.g pulp_url and that seemed a bit fragile. What do you reckon? Anyone running pulp with https already will need to be careful syncing in this change.

markgoddard:
> > We've set this in the ci-* environments, so it could be dropped there now.
> 
> Done. It did make me think we could have a better default rather than hard-coding `true`, but kolla_docker_registry has the protocol stripped, so you'd have to look at one of the intermediate variables e.g pulp_url and that seemed a bit fragile. What do you reckon? Anyone running pulp with https already will need to be careful syncing in this change.

Yes, it's not ideal. I suppose we could use the same value that was in the environment config, which is most likely to be correct.

markgoddard (approved):

bbezak (approved):

PR #322, opened 2023-01-06
Add docs for rolling out changes to RabbitMQ

MoteHue:


doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+
+.. code-block:: console
+
+  kayobe playbook run etc/kayobe/ansible/pulp-container-sync.yml pulp-container-publish.yml -e stackhpc_pulp_images_kolla_filter=rabbitmq
+
+Generate the new config files for the overcloud services.
+
+.. code-block:: console
+
+  kayobe overcloud servuce configuration generate
```
dougszumski: nit: s/servuce/service

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
+===============================
+Rolling out changes to RabbitMQ
+===============================
+
+Prior to upgrading, some changes need to be rolled out for RabbitMQ.
+This should make RabbitMQ more stable during the upgrade(s). These changes
+are:
+
+* Enable the high availability setting via Kolla-Ansible
+  ``om_enable_rabbitmq_high_availability``.
```
dougszumski, line 15: Should we mention the potential for a performance hit here? Perhaps only some sites with slow disks / controllers will be affected. I suppose we won't be doing them all at once, so we can gather data as we go.

You would imagine that there would be significant overhead replicating everything to disk at large deployments.

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+* Update RabbitMQ to version 3.9.22
+
+**NOTE:** There is guaranteed to be downtime during this procedure, as it
+requires restarting RabbitMQ and all the OpenStack services that use it. The
+state of RabbitMQ may also need to be reset.
```
dougszumski: Is it a necessary step to flatten Rabbit so that all of the long-lived queues are converted to durable? I'm wondering if we should always do it.

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+  kayobe overcloud host command run --command 'docker exec rabbitmq rabbitmqctl list_queues name durable'
+
+The cluster status should list all controllers. The queues listed should be
+durable if their names do not start with the following:
+
+* amq.
+* .\*\_fanout\_
+* reply\_
+
+At this stage, you may find that the above is not correct. If this is the case,
```
dougszumski: As above, I'm wondering if we should /always/ do this. IIRC you need to re-declare queues to convert them to durable?

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+
+If you do not need to use the playbook here, then you will need to bring the
+services back up yourself:
+
+.. code-block:: console
+
+  kayobe overcloud host command run --command 'docker ps -a | egrep '(barbican|blazar|ceilometer|cinder|cloudkitty|designate|heat|ironic|keystone|magnum|manila|masakari|neutron|nova|octavia)' | awk '{ print $NF }' | xargs docker start'
+
+If there are issues with the services after this, you may still need to run the
+hammer playbook.
```
dougszumski: /me shudders!

dougszumski (dismissed):
Looks good overall - a brave step!

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
+===============================
+Rolling out changes to RabbitMQ
+===============================
+
+Prior to upgrading, some changes need to be rolled out for RabbitMQ.
```
markgoddard: Could do with a bit more background here. e.g.

* what you get in kolla-ansible by default & potential problems with it
* what you get when the flag is set
* when we change the default (xena)

Then move onto how to roll it out, with conditions for different starting versions

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+**NOTE:** There is guaranteed to be downtime during this procedure, as it
+requires restarting RabbitMQ and all the OpenStack services that use it. The
+state of RabbitMQ may also need to be reset.
+
+Instructions
+------------
+
+If you are upgrading from Wallaby to Xena, you will need to enable the HA
+config option in
+``/home/ubuntu/kayobe/config/src/kolla-ansible/ansible/group_vars/all.yml``.
```
markgoddard: This is the Kolla Ansible source code which should not be modified (and is an absolute path which won't always be valid). Instead, add the variable in `etc/kayobe/kolla/globals.yml`.

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+* Update RabbitMQ to version 3.9.22
+
+**NOTE:** There is guaranteed to be downtime during this procedure, as it
+requires restarting RabbitMQ and all the OpenStack services that use it. The
+state of RabbitMQ may also need to be reset.
+
+Instructions
+------------
+
+If you are upgrading from Wallaby to Xena, you will need to enable the HA
```
markgoddard: Or even if you're on Wallaby and want a more stable Rabbit

doc/source/operations/rabbitmq.rst
markgoddard: RabbitMQ docs claim otherwise, but it's hard to know for sure without trying. It's worth calling out.

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+
+.. code-block:: console
+
+  kayobe overcloud container image pull -kt rabbitmq
+
+Stop all the OpenStack services which use RabbitMQ.
+
+.. code-block:: console
+
+  kayobe overcloud host command run --command 'docker ps -a | egrep '(barbican|blazar|ceilometer|cinder|cloudkitty|designate|heat|ironic|keystone|magnum|manila|masakari|neutron|nova|octavia)' | awk '{ print $NF }' | xargs docker stop'
```
markgoddard: We've got some nested quote issues here

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+
+  kayobe overcloud service upgrade -kt rabbitmq --skip-prechecks
+
+Check to see if RabbitMQ is functioning as expected.
+
+.. code-block:: console
+
+  kayobe overcloud service upgrade -kt rabbitmq --skip-prechecks
+  kayobe overcloud host command run --command 'docker exec rabbitmq rabbitmqctl cluster_status'
+  kayobe overcloud host command run --command 'docker exec rabbitmq rabbitmqctl list_queues name durable'
```
markgoddard: Better add `--show-output` to these

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+
+.. code-block:: console
+
+  kayobe overcloud service upgrade -kt rabbitmq --skip-prechecks
+
+Check to see if RabbitMQ is functioning as expected.
+
+.. code-block:: console
+
+  kayobe overcloud service upgrade -kt rabbitmq --skip-prechecks
```
markgoddard: Already did this :)

doc/source/operations/rabbitmq.rst
markgoddard: +1

doc/source/operations/rabbitmq.rst
markgoddard: :hammer: :rabbit:

doc/source/index.rst
```diff
@@ -22,3 +22,4 @@ Contents
    usage
    configuration/index
    contributor/index
+   operations/index
```
markgoddard: Could you rebase onto stackhpc/xena and propose there instead?

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
+===============================
+Rolling out changes to RabbitMQ
```
markgoddard: Could just be `RabbitMQ`, with a subsection for `High Availability`

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,90 @@
...
+===============================
+
+Prior to upgrading, some changes need to be rolled out for RabbitMQ.
+This should make RabbitMQ more stable during the upgrade(s). These changes
+are:
+
+* Enable the high availability setting via Kolla-Ansible
+  ``om_enable_rabbitmq_high_availability``.
+* Update RabbitMQ to version 3.9.22
```
markgoddard: Need a bit more info on when to do this (wallaby & xena if running old images), and how to do so.

markgoddard (changes requested):

doc/source/operations/rabbitmq.rst
MoteHue: I've just tested this, and indeed we do need to reset rabbit in order to convert queues to durable. Just restarting the services is not enough

doc/source/operations/rabbitmq.rst
MoteHue: I've changed the outer quotes to double quotes. This step needs to be looked into further, as the formatting of the output is not great when running the command through kayobe.

doc/source/operations/rabbitmq.rst
MoteHue: I've added some background here, would like to know if you think it should go more in-depth on the potential problems?

doc/source/operations/rabbitmq.rst
MoteHue: I've added a bit to mention the potential performance hit, it does seem like a good idea to come back and update this once we've tried it out on production systems.

doc/source/index.rst
```diff
@@ -32,3 +32,4 @@ Contents
    configuration/index
```
markgoddard, line 32: Could you move operations under configuration?

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,103 @@
...
+  or Xena, by synchronising and then pulling a new RabbitMQ container from
+  Pulp.
+* Enable the high availability setting via Kolla-Ansible
+  ``om_enable_rabbitmq_high_availability``.
+
+By default in Kolla-Ansible, two key options for the high availability of
+RabbitMQ are disabled. These are durable queues, where messages are persisted
+to disk; and classic queue mirroring, where messages are replicated across
+multiple exchanges. Without these, a deployment will be a poor state for
+managing any updates for RabbitMQ, or any outages that cause it to be stopped.
```
markgoddard: Might need to be a bit more diplomatic here.

```suggestion
multiple exchanges. Without these, a deployment has a higher risk of experiencing
issues when updating RabbitMQ, or recovering from network outages.
```

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,103 @@
...
+Messages held in RabbitMQ nodes that are stopped will be lost, which causes
+knock-on effects to the OpenStack services which either sent or were expecting
+to receive them. The Kolla-Ansible flag
+``om_enable_rabbitmq_high_availability`` can be used to enable both of these
+options. This will default to ``true`` from Xena onwards.
```
markgoddard:
```suggestion
options. The default will be overridden to ``true`` from Xena onwards in StackHPC Kayobe configuration.
```

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,103 @@
...
+large deployments. These changes would therefore be a tradeoff of performance
+for stability.
+
+**NOTE:** There is guaranteed to be downtime during this procedure, as it
+requires restarting RabbitMQ and all the OpenStack services that use it. The
+state of RabbitMQ will also be reset.
+
+Instructions
+------------
+
```
markgoddard:
```suggestion
If you are planning to perform an upgrade, it is recommended to first roll out these changes.
```

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,103 @@
...
+If you are upgrading from Wallaby to Xena, or if you're on Wallaby and want
+RabbitMQ to be more stable, you will need to enable the HA config option in
```
markgoddard:
```suggestion
If you are currently running Wallaby, you will need to enable the HA config option in
```

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,103 @@
...
+``etc/kayobe/kolla/globals.yml``.
+
+.. code-block:: console
+
+  om_enable_rabbitmq_high_availability: true
+
+Synchronise the Pulp containers.
```
markgoddard:
```suggestion
If you are running Wallaby or Xena, synchronise the Pulp containers.
```

doc/source/operations/rabbitmq.rst
```diff
@@ -0,0 +1,103 @@
...
+
+  kayobe overcloud service upgrade -kt rabbitmq --skip-prechecks
+
+In order to convert the queues to be durable, you will need to reset the state
+of RabbitMQ, and restart the services which use it. This can be done with the
+RabbitMQ hammer playbook:
+
+.. code-block:: console
+
+  kayobe playbook run stackhpc-kayobe-config/etc/kayobe/ansible/rabbitmq-reset.yml
```
markgoddard, line 85: Does this playbook restart all of the services we stopped?

markgoddard (changes requested):

doc/source/operations/rabbitmq.rst
MoteHue: Oh, no it just restarts the services that have been seen to break when RabbitMQ goes down. IMO it's best to keep the playbook only targeting this subset so it can be used separately if rabbit breaks. So I'll add the step to start all the services back into the instructions again.

doc/source/index.rst
MoteHue: Do you mean just changing the order of this index, or moving the entire section to be a part of configuration?

markgoddard: Just change the order

markgoddard (approved):

markgoddard:
It says there are unresolved comments but I can't see any.

markgoddard:
Oh, it's @dougszumski's review

PR #324, opened 2023-01-09
Add cephadm_commands placeholder variable

jovial:
Support for this feature was added in #290. This adds a variable and some documentation to raise awareness of its existence.

### DIFF
--- a/doc/source/configuration/cephadm.rst
+++ b/doc/source/configuration/cephadm.rst
@@ -243,7 +243,7 @@ Ceph Commands
 ~~~~~~~~~~~~~
 
 It is possible to run an arbitrary list of commands against the cluster after deployment
-by setting the ``ceph_commands`` variable. ``ceph_commands`` should be a list of commands
+by setting the ``cephadm_commands`` variable. ``cephadm_commands`` should be a list of commands
 to pass to ``cephadm shell -- ceph``. For example:
 
 .. code:: yaml
--- a/doc/source/configuration/cephadm.rst
+++ b/doc/source/configuration/cephadm.rst
@@ -239,6 +239,21 @@ for Cinder, Cinder backup, Glance, and Nova in Kolla Ansible.
          mgr: "profile rbd pool=images"
        state: present
 
+Ceph Commands
+~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


doc/source/configuration/cephadm.rst
```diff
@@ -239,6 +239,21 @@ for Cinder, Cinder backup, Glance, and Nova in Kolla Ansible.
          mgr: "profile rbd pool=images"
        state: present
 
+Ceph Commands
+~~~~~~~~~~~~~
+
+It is possible to run an arbitrary list of commands against the cluster after deployment
+by setting the ``ceph_commands`` variable. ``ceph_commands`` should be a list of commands
```
markgoddard:
```suggestion
by setting the ``cephadm_commands`` variable. ``cephadm_commands`` should be a list of commands
```

markgoddard (changes requested):

doc/source/configuration/cephadm.rst
```diff
@@ -239,6 +239,21 @@ for Cinder, Cinder backup, Glance, and Nova in Kolla Ansible.
...
+to pass to ``cephadm shell -- ceph``. For example:
+
+.. code:: yaml
+
+   # A list of commands to pass to cephadm shell -- ceph. See stackhpc.cephadm.commands
+   # for format.
+   cephadm_commands:
+     - "fs new cephfs cephfs_metadata cephfs_data"
```
mnasiadka: I would argue that's a good example, changing placement should rather be done using cephadm role, not the commands role.

jovial: That is the example from the docs in the collection. I was going to use this approach to do:

```
ceph config set mgr mgr/prometheus/server_addr 10.0.0.1
```

So it wasn't bound to all interfaces. Is that a better example?

EDIT: Assuming you meant **not** a good example

mnasiadka: Yes, an mgr var is a good example :-) I meant the old one is NOT a good example.

jovial: Thanks, makes sense to not encourage a bad pattern of usage.

markgoddard (approved):

PR #325, opened 2023-01-09
Improve description of Blackbox SSL alerts

priteau:


markgoddard (changes requested):
Could you propose to stackhpc/xena instead?

priteau:
> Could you propose to stackhpc/xena instead?

Sorry, I forgot about it. Done now.

markgoddard (approved):

PR #326, opened 2023-01-10
Enable RabbitMQ HA flag

MoteHue:


markgoddard (changes requested):
No need for a PR to yoga - we'll merge the xena changes into yoga.

markgoddard:
On second thoughts, it's useful to have this commit to cherry pick.

PR #327, opened 2023-01-10
Enable RabbitMQ HA flag and add new rabbitmq_tags

MoteHue:


markgoddard (changes requested):
Could you add a release note?

MoteHue:
> Could you add a release note?

Should I add one to the Wallaby branch too, or will this cover both?

markgoddard:
> > Could you add a release note?
> 
> Should I add one to the Wallaby branch too, or will this cover both?

Please add one for wallaby also

markgoddard (approved):

PR #337, opened 2023-01-16
Make pulp startup timeout variable

Alex-Welsh:


mnasiadka (commented):
Wouldn't it be better to increase delay?

Alex-Welsh:
> Wouldn't it be better to increase delay?

Why would that be preferable?

mnasiadka:
> > Wouldn't it be better to increase delay?
> 
> Why would that be preferable?

Less cycles to use ;-) I'm not a fan of hammering a service every 2 seconds for a long time. We could also make it configurable, because the commit message doesn't say why we need it.

Alex-Welsh:
Reworked to make the delay user configurable

etc/kayobe/containers/pulp/post.yml
```diff
@@ -4,8 +4,8 @@
     url: "{{ pulp_url }}/pulp/api/v3/status/"
   register: pulp_status
   until: pulp_status is success
-  retries: 30
-  delay: 2
+  retries: "{{ pulp_timeout_retries | default(30) }}"
+  delay: "{{ pulp_timeout_delay | default(2) }}"
```
markgoddard: Let's increase the default enough to fix your problem @Alex-Welsh

Alex-Welsh: Done

markgoddard (approved):

PR #338, opened 2023-01-17
Use Rocky Linux 8.7 repositories

m-bull:


markgoddard (changes requested):
Could you add a release note?

markgoddard (approved):

PR #340, opened 2023-01-19
CI: Workaround container image builder power cycle issue

markgoddard:
While not pretty, this workaround avoids manual intervention after the
runner is power cycled.

### DIFF
--- a/.github/workflows/stackhpc-container-image-build.yml
+++ b/.github/workflows/stackhpc-container-image-build.yml
@@ -66,6 +66,14 @@ jobs:
           sudo ip l set dummy1 up
           sudo ip l set dummy1 master breth1
 
+      # FIXME: Without this workaround we see the following issue after the runner is power cycled:
+      # TASK [MichaelRigart.interfaces : RedHat | ensure network service is started and enabled] ***
+      #   Unable to start service network: Job for network.service failed because the control process exited with error code.
+      #   See \"systemctl status network.service\" and \"journalctl -xe\" for details.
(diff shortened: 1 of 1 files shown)


priteau:
Could this ever cause the runner to lose its DHCP lease?

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -66,6 +66,14 @@ jobs:
           sudo ip l set dummy1 up
           sudo ip l set dummy1 master breth1
 
+      # FIXME: Without this workaround we see the following issue after the runner is power cycled:
+      # TASK [MichaelRigart.interfaces : RedHat | ensure network service is started and enabled] ***
+      #   Unable to start service network: Job for network.service failed because the control process exited with error code.
+      #   See \"systemctl status network.service\" and \"journalctl -xe\" for details.
+      - name: Kill dhclient (workaround)
+        run: |
+          sudo killall dhclient
```
jovial: Do we need to handle the case when this process does not exist?

```
[will@juno ~]$ killall doesnotexist
doesnotexist: no process found
[will@juno ~]$ echo $?
1
```
By either adding `|| true` or allowing the step to fail?

markgoddard: Good catch

markgoddard:
> Could this ever cause the runner to lose its DHCP lease?

Good question. Needs some testing.

priteau (approved):

PR #345, opened 2023-01-20
xena: wallaby merge

priteau:


markgoddard (dismissed):

priteau:
Is this the proper way of merging changes from one branch to another? I believe CI will automatically create a Git tag after the merge, and it will look like this tag has the UEFI fix (because there is the release note), while I still need to build Nova images for Xena and bump the container image tag.

markgoddard:
> Is this the proper way of merging changes from one branch to another? I believe CI will automatically create a Git tag after the merge, and it will look like this tag has the UEFI fix (because there is the release note), while I still need to build Nova images for Xena and bump the container image tag.

I usually just merge one branch into the next, then check the diff and make any changes necessary for the current branch. In this case that should probably include building a container and bumping the tag.

priteau:
I have triggered a nova container build using this branch: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/3967356367

markgoddard (approved):

PR #347, opened 2023-01-20
adding alert for nf_conntrack

Alex-Welsh:


etc/kayobe/kolla/config/prometheus/system.rules
```diff
@@ -78,4 +78,13 @@ groups:
...
       description: "{{ $labels.instance }} has had {{ printf \"%.0f\" $value }} uncorrectable memory errors reported by EDAC in the last 5 minutes."
 
+  - alert: HostConntrackLimit
+    expr: node_nf_conntrack_entries / node_nf_conntrack_entries_limit > 0.8
+    for: 5m
+    labels:
+      severity: warning
+    annotations:
+      summary: Host conntrack limit (instance {{ $labels.instance }})
+      description: "The number of conntrack is approaching limit\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"
```
markgoddard: I think we dropped the VALUE & LABELS bits
```suggestion
      description: "The number of conntrack is approaching limit"
```

markgoddard (changes requested):
Could you also rebase and target against stackhpc/xena?

PR #348, opened 2023-01-20
Remove trailing white space

priteau:


mnasiadka (approved):
Should we have some rst linter? ;-)

cityofships:
We could use pre-commit hooks for such stuff. Plus, CI checks that would prevent from re-occuring.

PR #354, opened 2023-01-25
Update the hardware overview dashboard

technowhizz:
Adds the temperature of drives as a column to the table.
Adds a timeseries to track the history of the temperature.
Introduces a variable dropdown picker for nodes so that only information about disks on a specific node is shown.


dougszumski (approved):
Good job, this is live on sms-lab if anyone wants to see it.

PR #359, opened 2023-01-25
Add timing information to tasks

m-bull:
Add ansible.posix.profile_tasks to callback_whitelist, which enables the output of task timing information when running `kayobe` commands.


etc/kayobe/ansible.cfg
```diff
@@ -6,6 +6,8 @@ stdout_callback = yaml
 bin_ansible_callbacks = True
 # Disable fact variable injection to improve performance.
 inject_facts_as_vars = False
+# Add timing information to output
+callback_whitelist = ansible.posix.profile_tasks
```
markgoddard, line 10: Downside: now there is written evidence of how slow it is :)

markgoddard (approved):

PR #366, opened 2023-01-30
added vim to overcloud dib image

g0rgamesh:


priteau (approved):
Of course.

bbezak:
thx!, `overcloud host configure` is installing vim, but it is useful to have it before that.

PR #376, opened 2023-02-06
Multinode env

jackhodgkiss:
Move wallaby multinode environment into `stackhpc/wallaby` in order to centralise efforts as opposed to it being its own branch.


markgoddard (changes requested):
We can't do this because it would bring features from xena+ that we don't want in wallaby. I was suggesting we merge the other way

jackhodgkiss:
> We can't do this because it would bring features from xena+ that we don't want in wallaby. I was suggesting we merge the other way

Sorry the misunderstanding. However this branch is based off of stackhpc/wallaby so where would the xena features originate from?

markgoddard:
> > We can't do this because it would bring features from xena+ that we don't want in wallaby. I was suggesting we merge the other way
> 
> Sorry the misunderstanding. However this branch is based off of stackhpc/wallaby so where would the xena features originate from?

We backported some things into multinode-env - LVM, cephadm, etc.

markgoddard:
can this be closed?

PR #386, opened 2023-02-10
Use stricter permissions for Pulp TLS cert and key

priteau:
Pulp runs as root inside its container. It has no problem reading files with 0600 permissions. There is no reason use 0644, especially for the key which is readable by any user on the seed.


m-bull (approved):
Good spot - thanks @priteau.

PR #388, opened 2023-02-13
Add 2m wait to some blackbox rules

technowhizz:
Adds a 2m wait to some blackbox rules to allow for temporary DNS failures to resolve. This also should stop false positives in the case of a probe timeout that gets resolved by the next probe. Also shows the status code on probe http failures.


cityofships (approved):
Thanks, looking good

PR #342, opened 2023-01-19
Local kolla container image builds

m-bull:
- Sync all repositories to local Pulp

PR #405, opened 2023-02-22
Bump the default worker count

jovial:
Hoping this will help with the slowness when pulling container images
across multiple hosts. Credit to @m-bull for noticing the option:

https://github.com/pulp/pulp-oci-images/commit/29c29df5bf218bfed80a52316356fc0dc89d7bf1

I've bumped pulp to 3.22 so that we don't have to force pull 3.21 to get
an image with this feature enabled.

### DIFF
--- a/etc/kayobe/seed.yml
+++ b/etc/kayobe/seed.yml
@@ -106,11 +106,14 @@ seed_pulp_container:
     image: pulp/pulp
     pre: "{{ kayobe_config_path }}/containers/pulp/pre.yml"
     post: "{{ kayobe_config_path }}/containers/pulp/post.yml"
-    tag: "{{ '3.21-https' if pulp_enable_tls | bool else '3.21' }}"
+    tag: "{{ '3.22-https' if pulp_enable_tls | bool else '3.22' }}"
     network_mode: host
     # Override deploy_containers_defaults.init == true to ensure
     # s6-overlay-suexec starts as pid 1
     init: false
+    env:
+      PULP_CONTENT_WORKERS: "{{ ansible_facts.processor_vcpus * 2 + 1 }}"
+      PULP_API_WORKERS: "{{ ansible_facts.processor_vcpus * 2 + 1 }}"
     volumes:
       - /opt/kayobe/containers/pulp:/etc/pulp
       - pulp_storage:/var/lib/pulp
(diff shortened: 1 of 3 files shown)


etc/kayobe/seed.yml
```diff
@@ -106,11 +106,14 @@ seed_pulp_container:
     image: pulp/pulp
     pre: "{{ kayobe_config_path }}/containers/pulp/pre.yml"
     post: "{{ kayobe_config_path }}/containers/pulp/post.yml"
-    tag: "{{ '3.21-https' if pulp_enable_tls | bool else '3.21' }}"
+    tag: "{{ '3.22-https' if pulp_enable_tls | bool else '3.22' }}"
```
markgoddard, line 109: Noting concerns about changes to distribution updates raised in slack https://stackhpc.slack.com/archives/C044YBYMJBZ/p1677077026176839

jovial: I hit that on 3.21 (not tried 3.22 yet as I didn't want to deviate from the upgrade plan). The issue was that the distributions I had created for the local image build feature deviated from the ones that actually landed in https://github.com/stackhpc/stackhpc-kayobe-config/pull/342. This caused:

```
failed: [localhost] (item={'name': 'treasuredata-4-development', 'base_path': 'treasuredata/4/redhat/8/x86_64/development', 'repository': 'TreasureData 4', 'state': 'present', 'required': True}) => changed=false                         
  ansible_loop_var: item                                                                                                                                                                                             
  item:                                                                                                                                                                                                                                     
    base_path: treasuredata/4/redhat/8/x86_64/development                                                                                                                                                                                   
    name: treasuredata-4-development                                                                                                                                                                                                        
    repository: TreasureData 4                                                                                                                                                                                                              
    required: true                                                                                                                                                                                                                          
    state: present                                                                                                                                                                                                   
  msg: 'HTTP Error 400: Bad Request b''{"base_path":["This field must be unique."]}''' 
```
I guess the other option is that we use the `pull` option on the docker_container module, but that can be non-desirable. We just need a newer build of the container to make sue we have the `PULP_*_WORKERS` variables available (You might have an old version of 3.21 that is missing this feature).

markgoddard: Ack. Has 3.22 been tested?

jovial: Not yet. Is that a requirement for merging?

markgoddard: Yes :) ideally on a system with some real data

priteau: @MoteHue and I are testing it now.

MoteHue: We've tested 3.22 now and it seems to be working fine :+1:

priteau (approved):

etc/kayobe/seed.yml
jovial: Thanks guys - marking as resolved :)

PR #406, opened 2023-02-22
Update Openstack Dashboard

technowhizz:
Updates the Openstack dashboard to show http 300 status codes as green as this is the expected return code from some services.

### DIFF
--- a/releasenotes/notes/openstack-dash-ff7bf179e37c8e4b.yaml
+++ b/releasenotes/notes/openstack-dash-ff7bf179e37c8e4b.yaml
@@ -1,0 +1,5 @@
+---
+other:
+  - |
+    Changes the openstack dashboard to show http status 300 as green instead of
+    orange.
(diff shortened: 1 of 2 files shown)


markgoddard (changes requested):
Please propose to stackhpc/xena instead

technowhizz:
@markgoddard proposed against xena

releasenotes/notes/openstack-dash-ff7bf179e37c8e4b.yaml
```diff
@@ -0,0 +1,5 @@
+---
+other:
+  - |
+    Changes the openstack dashboard to show http status 300 as green instead of
```
markgoddard:
```suggestion
    Changes the Grafana OpenStack dashboard to show HTTP status 300 as green instead of
```

technowhizz: Applied

markgoddard (approved):

PR #407, opened 2023-02-23
Bump neutron containers to support DellOS10/trunking

jovial:
Brings in the following networking-generic-switch-changes:

- Adds support for DellOS10
- Adds support for trunk ports


etc/kayobe/kolla/globals.yml
```diff
@@ -1,16 +1,18 @@
 # yamllint disable-file
 ---
 
-# To work around issue of trying to install docker from 
-# empty pulp server, use upstream docker dnf repo on 
+# To work around issue of trying to install docker from
```
jovial: Remove this whitespace change

markgoddard (approved):

PR #410, opened 2023-02-28
Fix build of release notes

priteau:


priteau:
Release note build failed in CI, but I think it may work once merged. Reno looks at the actual branch, e.g. `stackhpc/xena`, instead of the topic branch. It works locally when I commit to `stackhpc/xena`.

jackhodgkiss (approved):
LGTM.

bbezak (approved):

markgoddard:
> Release note build failed in CI, but I think it may work once merged. Reno looks at the actual branch, e.g. `stackhpc/xena`, instead of the topic branch. It works locally when I commit to `stackhpc/xena`.

Seems like a bug - shouldn't it use the PR branch if we want to test added renos?

priteau:
> > Release note build failed in CI, but I think it may work once merged. Reno looks at the actual branch, e.g. `stackhpc/xena`, instead of the topic branch. It works locally when I commit to `stackhpc/xena`.
> 
> Seems like a bug - shouldn't it use the PR branch if we want to test added renos?

Possibly Zuul does something different to allow generating release notes during CI jobs?

Also we may be hitting limitations of Reno's design, since we don't seem to follow its expected workflow: https://docs.openstack.org/reno/latest/user/design.html#assumptions

PR #411, opened 2023-02-28
Various multinode fixes and features

Alex-Welsh:


Alex-Welsh:
This PR is twinned with [this](https://github.com/stackhpc/openstack-config-multinode/pull/4) change for the multinode openstack configuration which uses the seed as the gateway on the external subnet.

etc/kayobe/ansible/configure-vxlan.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Configure VXLAN
   hosts: controllers,compute,seed,storage
-  gather_facts: false
+  gather_facts: true
```
markgoddard, line 4: There is potential for some incorrect behaviour here, since facts are gathered as the bootstrap user using the system python interpreter, some facts may be incorrect. In general we disable fact gathering during the user & venv bootstrap phase of host configure to avoid this. I think this is mostly an issue when using fact caching, but I can't remember all the details.

Let's try it like this, but keep an eye out for odd owner/permissions issues.

etc/kayobe/environments/ci-multinode/network-allocation.yml
```diff
@@ -0,0 +1,2 @@
+external_ips:
+  alex-mn-seed: 192.168.38.3
```
markgoddard: This isn't going to work if you change the cluster name...

etc/kayobe/environments/ci-multinode/networks.yml
```diff
@@ -73,7 +73,7 @@ storage_mgmt_net_name: storage_mgmt
 
 # Internal network
 internal_cidr: 192.168.37.0/24
-internal_mtu: 1450
+internal_mtu: 1392
```
markgoddard: As discussed, this probably needs to be the MTU of admin_oc_interface minus 50 bytes.

etc/kayobe/environments/ci-multinode/seed.yml
```diff
@@ -8,3 +8,5 @@ seed_lvm_groups:
 seed_extra_network_interfaces:
   - "external"
   - "public"
+
+seed_enable_snat: true
```
markgoddard, line 14: Please add a comment

releasenotes/notes/add-monitoring-to-multinode-958603f026edbc7c.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Set monitoring services be enabled by default in the multinode environment.
```
markgoddard:
```suggestion
    Set monitoring services be enabled by default in the ``ci-multinode`` environment.
```

releasenotes/notes/add-vm-overcloud-support-92baa7c2ce87ac1d.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Adds support for using a VMs as compute and controller nodes in the
+    multinode-ci environment by decreasing the MTU of the networks in
```
markgoddard:
```suggestion
    ``ci-multinode`` environment by decreasing the MTU of the networks in
```

markgoddard (changes requested):

etc/kayobe/environments/ci-multinode/network-allocation.yml
Alex-Welsh: yeah, painfully obvious once I actually thought about it. Is there some magic way of getting the name of the seed in Kayobe or will it need to be terraform templated?

markgoddard: Possibly you could use `{{ groups['seed'][0] }}` as the key, but I don't think the ip-allocation playbook would handle this.

I think currently terraform does not know about the IP subnets in this config, and it should probably stay that way.

Alternatively we could rely on the fact that `kayobe seed host configure` runs first and will grab this IP.

Alex-Welsh: > Possibly you could use `{{ groups['seed'][0] }}` as the key, but I don't think the ip-allocation playbook would handle this.

I gave that a go but it doesn't seem to template properly. Surprisingly it didn't error out but it also didn't apply the change. 

I think we may just have to assume it's on `192.168.38.3` and change either the kayobe or openstack configuration if the environment is different

releasenotes/notes/add-vm-overcloud-support-92baa7c2ce87ac1d.yaml
Alex-Welsh: Done

releasenotes/notes/add-monitoring-to-multinode-958603f026edbc7c.yaml
Alex-Welsh: Done

etc/kayobe/environments/ci-multinode/seed.yml
Alex-Welsh: Done

etc/kayobe/environments/ci-multinode/networks.yml
Alex-Welsh: Done

Alex-Welsh:
Closing this PR and swtiching target to stackhpc/xena

PR #414, opened 2023-03-01
Xena: generate all release notes

markgoddard:
- Fix release train 02-2023 renos
- release notes: Remove default_branch config option
- Use yoga upper constraints for releasenotes tox env

### DIFF
--- a/releasenotes/config.yaml
+++ b/releasenotes/config.yaml
@@ -1,4 +1,3 @@
 ---
 # This needs to be updated to the latest release.
-default_branch: stackhpc/xena
 release_tag_re: stackhpc/11\.\d+\.\d+\.\d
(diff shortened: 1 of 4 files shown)


priteau (commented):
How would you feel about using zed upper constraints? This would fix an issue issue with building lxml on macOS, at least with Apple Silicon. On the other hand we would lose support for Python 3.6…

markgoddard:
> How would you feel about using zed upper constraints? This would fix an issue issue with building lxml on macOS, at least with Apple Silicon. On the other hand we would lose support for Python 3.6…

Works for me, but can it be a separate change?

priteau (approved):

priteau:
> > How would you feel about using zed upper constraints? This would fix an issue issue with building lxml on macOS, at least with Apple Silicon. On the other hand we would lose support for Python 3.6…
> 
> Works for me, but can it be a separate change?

Sure, let's merge!

PR #419, opened 2023-03-01
Fix InstanceDown alert

technowhizz:
Fixes the InstanceDown alerting rule wait time to be consistent with the alert message. The alert message says "for 5 minutes" but the rule was set to wait for 1 minute.

Setting the wait to 5m also matches the example alert provided by prometheus


jovial (approved):
LGTM

PR #421, opened 2023-03-01
Change fix-networking.yml so it doesn't overwrite all of /etc/hosts

MoteHue:


jovial (dismissed):

etc/kayobe/ansible/fix-networking.yml
```diff
@@ -11,11 +11,10 @@
     ansible_ssh_common_args: "-o StrictHostKeyChecking=no"
   tasks:
     - name: Ensure `hosts` file contains pulp entries
-      ansible.builtin.copy:
-        content: |
-          127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
-          ::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
+      blockinfile:
+        path: /etc/hosts
+        marker_begin: BEGIN Kayobe Pulp entries
```
markgoddard: Should it use `marker` instead? Otherwise the end marker will be unchanged

MoteHue: Yes, unless we still want it to specify BEGIN and END in the markers. In that case, we should add `marker_end`.

markgoddard (approved):

PR #422, opened 2023-03-01
Add ceph_mgr_exporter endpoints

technowhizz:
Adds code to the globals.yml file to add endpoints for the ceph_mgr_exporter. If ceph is configured correctly, managers will be under the mgrs inventory group. If this group is empty, then the variable will just be empty.


etc/kayobe/kolla/globals.yml
```diff
@@ -31,6 +31,11 @@ prometheus_node_exporter_extra_volumes:
   - "textfile:/var/lib/node_exporter/textfile_collector"
 prometheus_node_exporter_cmdline_extras: "--collector.textfile.directory=/var/lib/node_exporter/textfile_collector"
 
+# Set Ceph manager exporter endpoints
+prometheus_ceph_mgr_exporter_endpoints:
```
markgoddard, line 35: If the mgrs group is empty this will end up as:

```
prometheus_ceph_mgr_exporter_endpoints:
```

Which is equivalent to setting it to a Python `None`. Kolla Ansible will then fail to template prometheus.yml:

```
{% for exporter in prometheus_ceph_mgr_exporter_endpoints %}
```

etc/kayobe/kolla/globals.yml
```diff
@@ -31,6 +31,11 @@ prometheus_node_exporter_extra_volumes:
...
+{% for host in groups['mgrs'] %}
+  - "{{ host }}:9283"
```
markgoddard, line 37: This uses the Ansible `inventory_hostname`, which is not guaranteed to be resolvable. You could do this:

```
 - {{ admin_oc_net_name | net_ip(host) }}:9283
```

markgoddard (changes requested):

jovial:
Could you add something to the docs to mention that you need to enable the scraping with:
```
kolla_enable_prometheus_ceph_mgr_exporter: true
```
in etc/kayobe/kolla.yml

PR #431, opened 2023-03-07
Bump neutron containers to support NGS batching

jovial:
This is used by scientific-openstack and is opt in via a config option.

See: https://github.com/stackhpc/networking-generic-switch/pull/54


markgoddard (approved):

jovial:
Promotion jobs:
- ubuntu: https://github.com/stackhpc/stackhpc-release-train/actions/runs/4362953617
- centos: https://github.com/stackhpc/stackhpc-release-train/actions/runs/4362950644

PR #471, opened 2023-04-05
Re-enable Ubuntu repos in pulp

Alex-Welsh:


Alex-Welsh:
Waiting on CI before I mark as ready for review

etc/kayobe/apt.yml
```diff
@@ -39,12 +47,25 @@
 # * architecture: whitespace-separated list of architectures that will be used
 #   (optional, default is unset)
 # Default is an empty list.
-#apt_repositories:
+apt_repositories:
+  - url: "{{ stackhpc_repo_ubuntu_focal_url }}"
+    suites: "{{ ansible_facts.distribution_release }} {{ ansible_facts.distribution_release }}-updates {{ ansible_facts.distribution_release }}-backports {{ ansible_facts.distribution_release }}-security"
```
markgoddard: Wondering why we have security here as well as the security below.

markgoddard: In commit 31bc7455823f788f1becc5e063914802d68ebbb5 we did not have security here.

markgoddard (changes requested):

etc/kayobe/apt.yml
Alex-Welsh: My mistake, just a transcription error I think

markgoddard (approved):

PR #480, opened 2023-04-13
Change default Ceph release to Quincy

priteau:


markgoddard (commented):
My main concern here is the potential for surprise upgrades.

priteau:
> My main concern here is the potential for surprise upgrades.

Do you prefer to wait until we have a zed branch?

markgoddard:
> > My main concern here is the potential for surprise upgrades.
> 
> Do you prefer to wait until we have a zed branch?

Let's discuss in today's infra meeting, but that might be safer. Deployments can always make these changes in their branches.

markgoddard:
I think we agreed to stay on pacific for now

mnasiadka:
Pacific is EOL-ish - https://github.com/ceph/ceph/blob/main/doc/releases/releases.yml

markgoddard:
> Pacific is EOL-ish - https://github.com/ceph/ceph/blob/main/doc/releases/releases.yml

Yes. I'm not saying we shouldn't update clients to Quincy, but should we make such changes on a stable branch?

mnasiadka:
In this case it probably is not problematic (you would need to run bootstrap against a running cluster to make some serious problems), but I understand where you're going - I don't know even if we should have a default in s-k-c.

markgoddard:
> In this case it probably is not problematic (you would need to run bootstrap against a running cluster to make some serious problems), but I understand where you're going - I don't know even if we should have a default in s-k-c.

If we don't have a default in s-k-c then we depend on the default in the collection, which is less visible. We do specify the container image tag in s-k-c too, which is tied to the release

mnasiadka:
So let's just assume we just update the default once, in the early stages of a branch - and that's it.

markgoddard:
Perhaps we should better document how to change the release in SKC docs?

Alex-Welsh:
Is it time to close this PR @priteau @markgoddard?

priteau:
> Is it time to close this PR @priteau @markgoddard?

Should I target it to zed instead?

Alex-Welsh:
> > Is it time to close this PR @priteau @markgoddard?
> 
> Should I target it to zed instead?

It's already done: https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/zed/etc/kayobe/cephadm.yml

priteau:
Not required anymore.

PR #483, opened 2023-04-13
Restore upstream mirrors in container image package repos

markgoddard:
Currently the Kolla container images configure Yum repositories to use
pulp-server on SMS Cloud:

(nova-compute) $ grep baseurl /etc/yum.repos.d/CentOS-Stream-BaseOS.repo
baseurl=http://pulp-server.internal.sms-cloud:8080/pulp/content/centos/8-stream/BaseOS/x86_64/os/20211122T102435

This makes it difficult to install/update packages for debugging issues
in customer environments.

This change restores Yum repository files at the end of the container image
build to point to distribution repositories. It also makes an attempt to
do the same for Apt repos, although these are not enabled in Wallaby
currently.

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -216,11 +216,18 @@ stackhpc_ubuntu_focal_repos:
   - "deb {{ stackhpc_repo_ubuntu_focal_security_url }} focal-security main universe"
   - "deb {{ stackhpc_repo_ubuntu_cloud_archive_url }} focal-updates/wallaby main"
 
+# Whether to revert to the upstream mirrors in built Kolla container images.
+stackhpc_kolla_clean_up_repo_mirrors: true
+
 # Dict mapping Jinja2 block names in kolla's Docker images to their contents.
 kolla_build_blocks:
   base_header: |
     {% if kolla_base_distro == 'centos' %}
     RUN \
+    {% if stackhpc_kolla_clean_up_repo_mirrors | bool %}
+        mkdir -p /etc/yum.repos.d.backup && \
+        tar -czf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d . && \
+    {% endif %}
     {% for repo in stackhpc_centos_stream_repos %}
         sed -i -e 's/^\(mirrorlist *=.*\)/#\1/g' \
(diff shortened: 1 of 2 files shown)


markgoddard:
Tested here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/4689189856

markgoddard:
> Tested here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/4689189856

Tested out the image:
```
[centos@skc-builder-runner ~]$ docker run -u0 --rm -it --net=host 94df4f5bb497 bash
()[glance@skc-builder-runner /]$ grep -r baseurl /etc/yum.repos.d/ | grep pulp
()[glance@skc-builder-runner /]$ dnf upgrade
```

etc/kayobe/kolla.yml
```diff
@@ -216,11 +216,18 @@ stackhpc_ubuntu_focal_repos:
...
+stackhpc_kolla_clean_up_repo_mirrors: true
+
 # Dict mapping Jinja2 block names in kolla's Docker images to their contents.
 kolla_build_blocks:
   base_header: |
     {% if kolla_base_distro == 'centos' %}
     RUN \
+    {% if stackhpc_kolla_clean_up_repo_mirrors | bool %}
+        mkdir -p /etc/yum.repos.d.backup && \
+        tar -czf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d . && \
```
Alex-Welsh, line 229: Can we be sure tar exists at this point?

etc/kayobe/kolla.yml
```diff
@@ -231,11 +238,34 @@ kolla_build_blocks:
   base_centos_repo_overrides_post_yum: |
     {# fixme #}
         && \
+    {% if stackhpc_kolla_clean_up_repo_mirrors | bool %}
+        tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d && \
```
Alex-Welsh, line 242: Ditto

etc/kayobe/kolla.yml
markgoddard: At this point it's just the base CentOS container image, so we haven't installed tar, but it was present when I tested this.

mnasiadka (approved):

PR #486, opened 2023-04-17
rocky: disable elasticsearch-curator by default

mnasiadka:


priteau (approved):

etc/kayobe/kolla.yml
```diff
@@ -577,7 +577,7 @@ kolla_enable_central_logging: true
 #kolla_enable_designate:
 #kolla_enable_destroy_images:
 #kolla_enable_elasticsearch:
-kolla_enable_elasticsearch_curator: true
+kolla_enable_elasticsearch_curator: "{{ not os_distribution == 'rocky' }}"
```
markgoddard, line 580: Surely this should be handled by kayobe/kolla-ansible?

markgoddard (changes requested):

etc/kayobe/kolla.yml
mnasiadka: More like Kayobe, but we're explicitly enabling that - so what Kayobe/Kolla-Ansible should do - fail on prechecks?

mnasiadka: What about kolla_enable_elasticsearch_curator: "{{ kolla_enable_elasticsearch }}"? Is this acceptable?

markgoddard: Actually, if enable_elasticsearch is false on rocky, then does it matter whether curator is enabled?

mnasiadka: There is no container image for elasticsearch-curator - so it won't pull. But let me check.
Actually - @grzegorzkoper reported that, but now I think it's not needed - because he had kayobe code without the opensearch patch. Shouldn't be needed anymore.

PR #488, opened 2023-04-17
Fix multinode failures from seed misconfiguration

Alex-Welsh:
This patch contains two fixes.

The first changes the structure of `seed_extra_network_interfaces` as it was previously invalid, resulting in errors about concatenating a string with a dict.

The second catches a failure that occurs when manila is not enabled. It was assumed that the seed had a storage interface even though it not configured by default.

### DIFF
--- a/etc/kayobe/environments/ci-multinode/seed.yml
+++ b/etc/kayobe/environments/ci-multinode/seed.yml
@@ -3,9 +3,7 @@ seed_bootstrap_user: "{{ os_distribution if os_distribution == 'ubuntu' else 'cl
 seed_lvm_groups:
   - "{{ stackhpc_lvm_group_rootvg }}"
 
-seed_extra_network_interfaces: >
-  "{{ seed_extra_network_interfaces_external +
-  (seed_extra_network_interfaces_manila if (kolla_enable_manila | bool and kolla_enable_manila_backend_cephfs_native | bool) else []) }}"
+seed_extra_network_interfaces: "{{ seed_extra_network_interfaces_external + seed_extra_network_interfaces_manila if (kolla_enable_manila | bool and kolla_enable_manila_backend_cephfs_native | bool) else [] }}"
 
 # Seed has been provided an external interface
 # for tempest tests and SSH access to machines.
@@ -26,6 +24,6 @@ snat_rules_default:
     source_ip: "{{ ansible_facts.default_ipv4.address }}"
 snat_rules_manila:
   - interface: "{{ storage_interface }}"
-    source_ip: "{{ ansible_facts[storage_interface].ipv4.address }}"
+    source_ip: "{{ ansible_facts[storage_interface].ipv4.address | default }}"
 # Only add the storage snat rule if we are using manila-cephfs.
 snat_rules: "{{ snat_rules_default + snat_rules_manila if (kolla_enable_manila | bool and kolla_enable_manila_backend_cephfs_native | bool) else snat_rules_default }}"


etc/kayobe/environments/ci-multinode/seed.yml
```diff
@@ -26,6 +24,6 @@ snat_rules_default:
     source_ip: "{{ ansible_facts.default_ipv4.address }}"
 snat_rules_manila:
   - interface: "{{ storage_interface }}"
-    source_ip: "{{ ansible_facts[storage_interface].ipv4.address }}"
+    source_ip: "{{ ansible_facts[storage_interface].ipv4.address if ansible_facts[storage_interface].ipv4.address is defined else none }}"
```
markgoddard: Can you just do this?
```suggestion
    source_ip: "{{ ansible_facts[storage_interface].ipv4.address | default }}"
```

etc/kayobe/environments/ci-multinode/seed.yml
```diff
@@ -3,9 +3,7 @@ seed_bootstrap_user: "{{ os_distribution if os_distribution == 'ubuntu' else 'cl
 seed_lvm_groups:
   - "{{ stackhpc_lvm_group_rootvg }}"
 
-seed_extra_network_interfaces: >
-  "{{ seed_extra_network_interfaces_external +
-  (seed_extra_network_interfaces_manila if (kolla_enable_manila | bool and kolla_enable_manila_backend_cephfs_native | bool) else []) }}"
+seed_extra_network_interfaces: "{{ seed_extra_network_interfaces_external }} + {{ seed_extra_network_interfaces_manila if (kolla_enable_manila | bool and kolla_enable_manila_backend_cephfs_native | bool) else [] }}"
```
markgoddard: hmm, does this result in a proper list, or a string? I would have done it like this:
```suggestion
seed_extra_network_interfaces: "{{ seed_extra_network_interfaces_external + seed_extra_network_interfaces_manila if (kolla_enable_manila | bool and kolla_enable_manila_backend_cephfs_native | bool) else [] }}"
```

markgoddard: Although that looks like the original code. Perhaps you need to remove the trailing newline to make it parse as a list properly?
```
seed_extra_network_interfaces: >-
  "{{ seed_extra_network_interfaces_external +
  (seed_extra_network_interfaces_manila if (kolla_enable_manila | bool and kolla_enable_manila_backend_cephfs_native | bool) else []) }}"
```

Alex-Welsh: Your first suggestion works but the second has the same error as before. I was slightly wrong in the merge comment:
```

An unhandled exception occurred while templating '{{ network_interfaces | net_select_ethers | list }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while templating '{{ seed_network_interfaces | unique | list }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while templating '{{ (seed_default_network_interfaces +
        seed_extra_network_interfaces) | select | unique | list }}
    '. Error was a <class 'ansible.errors.AnsibleError'>, original message: Unexpected templating type error occurred on ({{ (seed_default_network_interfaces +
        seed_extra_network_interfaces) | select | unique | list }}
    ): can only concatenate list (not "str") to list
```

Alex-Welsh: > hmm, does this result in a proper list, or a string?

Results in a list (or at least it creates a string that can be parsed later as a valid list), though I do now see that it's a strange way of doing it

etc/kayobe/environments/ci-multinode/seed.yml
Alex-Welsh: I gave it a go and it looks good

etc/kayobe/environments/ci-multinode/seed.yml
markgoddard: Ah, the problem is the quotes, which result in a string with quotes in it because it's using a literal block scalar:

```
seed_extra_network_interfaces: >-
  {{ seed_extra_network_interfaces_external +
  (seed_extra_network_interfaces_manila if (kolla_enable_manila | bool and kolla_enable_manila_backend_cephfs_native | bool) else []) }}
```

It's fine as it is now though.

markgoddard (approved):

PR #492, opened 2023-04-19
Yoga: Grafana upgrade

cityofships:


markgoddard:
Does it not need to be bumped for Ubuntu?

cityofships:
> Does it not need to be bumped for Ubuntu?

Currently I don't have means to test on Ubuntu - but we can build and update it anyway.

cityofships:
Ubuntu builds are broken atm. https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/4751325578/jobs/8440355022

markgoddard (approved):

PR #501, opened 2023-04-27
Adds seed_redfish_exporter_container

jovial:


markgoddard (commented):
Still draft?

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -0,0 +1,25 @@
+{% if seed_redfish_exporter_container_enabled | bool %}
+scrape_configs:
+  - job_name: redfish-exporter-seed
```
bbezak, line 5: this name should be the same as in grafana dashboard (redfish-exporter)

jovial: Good call. I've updated the dashboard which was a little broken to use a regex:

`count(up{job=~"redfish-exporter.*"} == 1)`

jovial (commented):
My redfish exporter deployment was a bit broken at one customer site, so I updated this to fix my issues. Worked for me (TM)

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -0,0 +1,25 @@
...
+    metrics_path: /redfish
+    scrape_timeout: 120s
+    scrape_interval: 5m
```
markgoddard: These two need to scale with the size of the cloud. For RUG we ended up setting the scrape interval like this:

```
scrape_timeout: {{ 8 * groups['redfish_exporter_targets'] | length }}s
```

In a sufficiently large cloud the interval may need to increase too.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -0,0 +1,25 @@
...
+    relabel_configs:
+     - source_labels: [__address__]
+       target_label: __param_target
+     - source_labels: [__param_target]
+       target_label: instance
+     - target_label: __address__
+       replacement: "{{ lookup('vars', admin_oc_net_name ~ '_ips')[groups.seed.0] }}:9610"
+    static_configs:
+{% for host in groups.get('redfish_exporter_targets', []) %}
+{% if hostvars[host]["redfish_exporter_scrape_group"] | default('overcloud') == 'overcloud' %}
```
markgoddard, line 18: What's the idea of this?

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -0,0 +1,25 @@
...
+      - targets:
+          - '{{ hostvars[host]["redfish_exporter_target_address"] }}'
+        labels:
```
markgoddard, line 21: Should we override `instance` in the same way we do for other targets?

markgoddard (changes requested):

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
jovial: Possibly, but the dashboard is set up to use server

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
jovial: I was just thinking of when we want to monitor ironic nodes. We could put those in a separate group, but perhaps I should remove this stuff in this iteration (to keep it simple)?

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
jovial: Thanks, I'll use that

jovial:
```suggestion
    scrape_interval: {{ 8 * groups['redfish_exporter_targets'] | length }}s
```

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
jovial: Essentially we can use this label in the dashboard to prune down the list of servers. So you first select the group and then servers is dynamically updated to only show servers from that group.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
jovial: One further advantage is that we link to bmc via the instance label. If we overwrote that then the inventory hostname might not resolve. You could change the target to use a domain name if you have internal resolution and the link would still work.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
jovial: To illustrate my point:
image

I'm thinking there would be a dashboard variable to select: 'overcloud' or 'ironic'

markgoddard: :+1:

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
markgoddard: Is the scrape timeout scoped to each target or the whole scrape config? If it's for the whole scrape config then this would also need to change.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -0,0 +1,25 @@
+{% if seed_redfish_exporter_container_enabled | bool %}
+scrape_configs:
+  - job_name: redfish-exporter-seed
+    metrics_path: /redfish
+    scrape_timeout: 120s
+    scrape_interval: {{ 8 * groups['redfish_exporter_targets'] | length }}s
```
markgoddard: Perhaps we'd want a minimum interval here too to avoid excessive scraping in small clouds?

Alex-Welsh:
Merged Antelope into Yoga?

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
markgoddard: What would a sensible minimum be? 60s?

GregWhiteyBialas: I think everyone expects to have data of, at least, this granularity.

etc/kayobe/stackhpc-monitoring.yml
```diff
@@ -19,3 +19,16 @@ stackhpc_enable_os_capacity: true
...
 # exporter during Keystone authentication.
 stackhpc_os_capacity_openstack_verify: true
+# Redfish exporter
+
+# Whether the redfish exporter is enabled.
+stackhpc_enable_redfish_exporter: false
+
+# Credentials
+redfish_exporter_default_username: "{{ ipmi_admin_user }}"
+redfish_exporter_default_password: "{{ ipmi_admin_password }}"
```
markgoddard: Why are we using ipmi_admin_user / ipmi_admin_password rather than ipmi_username / ipmi_password? The latter are the standard variables in kayobe.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
markgoddard: This one needs an answer.

markgoddard (changes requested):
Needs a release note

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
GregWhiteyBialas: From docs https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config:

```
# How frequently to scrape targets from this job.
[ scrape_interval: [<duration>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration) | default = <global_config.scrape_interval> ]

# Per-scrape timeout when scraping this job.
[ scrape_timeout: [<duration>](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#duration) | default = <global_config.scrape_timeout> ]
```

So it looks like `scrape_timeout` is per target in job.

etc/kayobe/stackhpc-monitoring.yml
GregWhiteyBialas: Changed.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
markgoddard: Ok, let's apply a minimum of 60s then.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
markgoddard: Ack

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -0,0 +1,27 @@
...
+    relabel_configs:
+     - source_labels: [__address__]
+       target_label: __param_target
+     - source_labels: [__param_target]
+       target_label: instance
+     - target_label: __address__
+       replacement: "{{ lookup('vars', admin_oc_net_name ~ '_ips')[groups.seed.0] }}:9610"
+    static_configs:
+{% for host in groups.get('redfish_exporter_targets', []) %}
+{% if hostvars[host]["redfish_exporter_scrape_group"] | default('overcloud') == 'overcloud' %}
```
jovial, line 18: One issue I've noticed is that sometimes I get scrapes failing because the idrac is overloaded when multiple prometheus instances scrape the data at the same time. I had a few ideas:

- Shuffle the targets so the order was randomised on each host (easiest)
  - Use a constant seed for each host so that order is maintained i.e I want to stop config diff from showing a change each run.
  - See: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/shuffle_filter.html#examples
- Add caching layer (possibly with haproxy) that would serve stale data whilst a scrape was in progress i.e only allow one scrape at a time. No idea if haproxy can do this, but other caching proxies can.

jovial: See something like: https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_lock

markgoddard: Would a simple mutex/lock based on the BMC address in the exporter work?

jovial: > Would a simple mutex/lock based on the BMC address in the exporter work?

That sounds simpler than the cache. I must admit I was assuming that we didn't change the exporter. I guess we may have to bump the scrape timeout to account for the queued scrapes.  Though I was possibly slightly off in my earlier analysis of simultaneous scrapes... I found out it wasn't reusing sessions, so we were basically exhausting all the available sessions. On the idracs I am using there is a maximum session count of 8 with a timeout of 1800s. If scraping at 600s you basically use three sessions per prometheus. This is what you see in the logs:

```
t: 503: {\"error\":{\"@Message.ExtendedInfo\":[{\"Message\":\"The maximum number of user sessions is reached.\",\"MessageArgs\":[],\"MessageArgs@odata.count\":0,\"MessageId\":\"IDRAC.2.8.RAC0218\",\"RelatedProperties\":[],\"RelatedProperties@odata.count\":0,\"Resolution\":\"Make sure that the number of active sessions is less than or equal to the threshold limit. To end a current session and start a new session, do one of the following: 1) Log out of iDRAC by using a Graphical User Interface (GUI). 2) End any unused session by running the following RACADM command at the Command Line Interface (CLI): racadm closessn -i [session id].Warning: Ending a session may abruptly stop any running operation on that specific iDRAC session.\",\"Severity\":\"Informational\"}],\"code\":\"Base.1.12.GeneralError\",\"message\":\"A general error has occurred. See ExtendedInfo for more information\"}}" source="redfish_collector.go:48"
```
Adjusting the session timeout also affects the webui so it is not desirable to tweak that setting. The number of sessions is non-configurable.

jovial: So I need to raise an upstream bug as https://github.com/jenningsloy318/redfish_exporter/blob/d388474c4024d09df0f3ab0290068e9072a7cf21/collector/redfish_collector.go#L84 seems to be failing to end the session

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -0,0 +1,27 @@
+# yamllint disable-file
+---
+{% if seed_redfish_exporter_container_enabled | bool %}
+scrape_configs:
+  - job_name: redfish-exporter-seed
+    metrics_path: /redfish
+    scrape_timeout: 120s
+    scrape_interval: {{ 8 * groups['redfish_exporter_targets'] | length }}s
```
jackhodgkiss: If the size of the group `redfish_exporter_targets` is too small then the interval will be smaller than the timeout. 

```suggestion
    scrape_timeout: 120s
    scrape_interval: {{ [8 * groups['redfish_exporter_targets'] | length, 120] | max }}s
```

markgoddard: Might be better to keep the timeout of 120s and apply a minimum of 120s?

jackhodgkiss: Okay. Should it be capped at 120s or should it increase as the group grows larger. So replace `min` filter with `max`

markgoddard: and need to use the max filter

jackhodgkiss: Done.

markgoddard (approved):
Let's get this thing in.

markgoddard:
15 months later...

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -0,0 +1,27 @@
...
+       target_label: instance
+     - target_label: __address__
+       replacement: "{{ lookup('vars', admin_oc_net_name ~ '_ips')[groups.seed.0] }}:9610"
+    static_configs:
+{% for host in groups.get('redfish_exporter_targets', []) %}
+{% if hostvars[host]["redfish_exporter_scrape_group"] | default('overcloud') == 'overcloud' %}
+      - targets:
+          - '{{ hostvars[host]["redfish_exporter_target_address"] }}'
+        labels:
+          server: '{{ host }}'
```
cityofships, line 22: What's the motivation behind these additional labels?

markgoddard: Gives you a hostname rather than IP:port?

PR #502, opened 2023-04-28
AUFN-Ceph fixes and Rocky9 support

MoteHue:


sd109 (dismissed):

doc/source/contributor/environments/aufn-ceph.rst
```diff
@@ -75,7 +81,8 @@ Clone the Kayobe, Kayobe configuration (this one) and Tenks repositories:
    popd
    popd
 
-Create a virtual environment and install Kayobe:
+Create a virtual environment and install Kayobe. Note, if you are using Rocky
+you will need to use ``python3 -m venv`` as ``virtualenv`` is not available.
```
markgoddard: Let's keep it simple and switch to venv

etc/kayobe/environments/aufn-ceph/controllers.yml
```diff
@@ -0,0 +1,2 @@
+---
+controllers_bootstrap_user: "{{ lookup('env', 'USER') }}"
```
markgoddard: Is the same not required for storage and seed hosts?

etc/kayobe/environments/aufn-ceph/inventory/hosts
```diff
@@ -25,3 +25,6 @@ seed
 
 [hs-switches]
 # Add high speed switches here if required.
+
+[storage:children]
+storage-ceph
```
markgoddard: Group mappings should go in `groups`

markgoddard (changes requested):

etc/kayobe/environments/aufn-ceph/controllers.yml
MoteHue: I suppose it should be, although I didn't hit any issues without them defined...

doc/source/contributor/environments/aufn-ceph.rst
```diff
@@ -28,14 +28,20 @@ CentOS:
 
 .. parsed-literal::
 
-   sudo dnf install -y gcc python3-dev python3-virtualenv
+   sudo dnf install -y gcc
+
+Rocky:
```
markgoddard: CentOS & Rocky are now the same. Do we also need `python3-dev` though (on both), to build the kayobe package?

doc/source/contributor/environments/aufn-ceph.rst
```diff
@@ -181,9 +188,8 @@ We can deploy a test VM to ensure that our 'universe' is up and running by first
 
 .. parsed-literal::
 
-    VENV_DIR=~/openstack-env
-    python -m venv $VENV_DIR
-    source $VENV_DIR/bin/activate
+    python -m venv ~/openstack-env
```
markgoddard: Does this need to be `python3`? CentOS/Rocky no longer have an unversioned system `python`

etc/kayobe/environments/aufn-ceph/controllers.yml
markgoddard: It does make me wonder, why would control/compute be different than storage?

doc/source/contributor/environments/aufn-ceph.rst
MoteHue: Yes that does :+1:

doc/source/contributor/environments/aufn-ceph.rst
MoteHue: Oh yes I forgot to change that, the package is called `python3-devel` on CentOS and Rocky instead.

etc/kayobe/environments/aufn-ceph/controllers.yml
MoteHue: Just running through this again now, and it looks like we don't want to change the user for the seed. Fails in `seed host configure`:

MoteHue:
```TASK [Check if python is installed] *************************************************************************************************************************************************************************************
Wednesday 03 May 2023  12:57:44 +0000 (0:00:00.115)       0:00:02.268 *********
fatal: [seed]: UNREACHABLE! => changed=false 
  msg: 'Failed to connect to the host via ssh: cloud-user@192.168.33.5: Permission denied (publickey,gssapi-keyex,gssapi-with-mic).'                                                                                    
  unreachable: true```

Comment (MoteHue) [reply etc/kayobe/environments/aufn-ceph/controllers.yml]: I've ran through an AUFN again to nail down what's up with the different bootstrap users. Turns out I'd managed to misspell `controllers_bootstrap_user` (should be `controller` singular) so it was using `rocky` rather than `cloud-user`. I suspect I'd made a similar mistake with the computes the first time around...

The correct user for the latest Rocky9 image is `rocky`, so I've updated accordingly.

File: etc/kayobe/environments/aufn-ceph/seed-hypervisor.yml [hunk 6]
```diff
@@ -11,4 +11,4 @@ seed_hypervisor_extra_network_interfaces:
   - "{{ external_net_names[0] }}"
 
 # Workaround change to cloud-user default login name on CentOS-Stream8
-seed_hypervisor_bootstrap_user: "{{ lookup('env', 'USER') }}"
+seed_hypervisor_bootstrap_user: "{{ 'cloud-user' if os_distribution == 'centos' else os_distribution }}"
```
markgoddard: This one we might want to keep as USER? Does your seed hypervisor image have a rocky user?

etc/kayobe/environments/aufn-ceph/controllers.yml
```diff
@@ -0,0 +1,2 @@
+---
+controllers_bootstrap_user: "{{ 'cloud-user' if os_distribution == 'centos' else os_distribution }}"
```
markgoddard:
```suggestion
controller_bootstrap_user: "{{ 'cloud-user' if os_distribution == 'centos' else os_distribution }}"
```

etc/kayobe/environments/aufn-ceph/seed-hypervisor.yml
MoteHue: Oh yes, good catch. The Rocky image we use on SMS has cloud-user as the user

markgoddard (approved):

PR #507, opened 2023-05-05
Revert changes to use cloud-user

Alex-Welsh:
The upstream CentOS image uses centos as the user, not `cloud-user`. Reverting recent changes


MoteHue:
Ah I guess I should've tested this on CentOS too...
This change looks good to me :+1:

markgoddard (approved):

PR #529, opened 2023-06-07
net_ip templating fix

g0rgamesh:
changed net_ip filters which broke infra vm provision


etc/kayobe/pulp.yml
```diff
@@ -4,7 +4,7 @@
 
 # Base URL of the local Pulp service.
 # Default uses the seed node's IP on the admin network.
-pulp_url: "{{ 'https' if pulp_enable_tls | bool else 'http' }}://{{ admin_oc_net_name | net_ip(groups['seed'][0]) }}:{{ pulp_port }}"
+pulp_url: "{{ 'https' if pulp_enable_tls | bool else 'http' }}://{{ hostvars[groups['seed'][0]].admin_oc_net_name }}:{{ pulp_port }}"
```
markgoddard: This isn't the same thing - `admin_oc_net_name` is the string name of the admin overcloud network, not an IP address.

markgoddard (changes requested):

etc/kayobe/pulp.yml
g0rgamesh: Corrected

etc/kayobe/pulp.yml
```diff
@@ -4,7 +4,8 @@
 
 # Base URL of the local Pulp service.
 # Default uses the seed node's IP on the admin network.
-pulp_url: "{{ 'https' if pulp_enable_tls | bool else 'http' }}://{{ admin_oc_net_name | net_ip(groups['seed'][0]) }}:{{ pulp_port }}"
+seed_ip: "{{ provision_oc_net_name | net_ip }}"
```
markgoddard: Have you tested this? I expect it doesn't work because seed_ip will be evaluated for the host running the task (this could be localhost or a controller etc.)

markgoddard (changes requested):

etc/kayobe/pulp.yml
g0rgamesh: It didn't work, corrected to method which works.

etc/kayobe/pulp.yml
```diff
@@ -4,7 +4,7 @@
 
 # Base URL of the local Pulp service.
 # Default uses the seed node's IP on the admin network.
-pulp_url: "{{ 'https' if pulp_enable_tls | bool else 'http' }}://{{ admin_oc_net_name | net_ip(groups['seed'][0]) }}:{{ pulp_port }}"
+pulp_url: "{{ 'https' if pulp_enable_tls | bool else 'http' }}://{{ provision_oc_net_ips['seed0'] }}:{{ pulp_port }}"
```
markgoddard: In general the seed is not called `seed0`.

markgoddard (changes requested):

etc/kayobe/pulp.yml
g0rgamesh: I'm open to suggestions on how to maneuver around it.

markgoddard: Could try adding a `| default` filter to the end of the original expression?

g0rgamesh: Thanks Mark, added

markgoddard: Tested and working?

g0rgamesh: `kayobe seed host configure --tags dnf,apt` passed with success
seed ip templated correctly in /etc/docker/daemon.json on seed node

markgoddard:
It should work for the apt/dnf/docker case because it's basically unchanged
from before. Does it fix your original infra VM issue?

g0rgamesh:
> It should work for the apt/dnf/docker case because it's basically unchanged from before. Does it fix your original infra VM issue?

This latest change is tested against:
- `kayobe seed host configure --tags dnf,apt`
- `kayobe infra vm provision --limit wazuh-server`
Both tests are successful.

markgoddard (approved):
Looks like a good solution, nice work!

PR #533, opened 2023-06-12
Cap number of Pulp workers

priteau:
Running Pulp with a number of workers based on CPUs can cause errors such as PostgreSQL connection limits on hosts with many cores:

    psycopg.OperationalError: connection failed: FATAL:  remaining connection slots are reserved for non-replication superuser connections

Cap workers to 32 until large scale testing can identify a more suitable limit.


Alex-Welsh (approved):
LGTM

PR #534, opened 2023-06-15
Fix HAProxy dashboard

darmach:
Grafana Haproxy dashboard targets have been aligned with the format kolla-ansible templates them in prometheus - address followed by port number.


bbezak (approved):
tested in UZH

PR #536, opened 2023-06-19
Feature/ceph mgr exporter

technowhizz:
Add endpoints for ceph into the s-k-c defaults

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -77,6 +77,13 @@ prometheus_node_exporter_extra_volumes:
   - "textfile:/var/lib/node_exporter/textfile_collector"
 prometheus_node_exporter_cmdline_extras: "--collector.textfile.directory=/var/lib/node_exporter/textfile_collector"
 
+# Set Ceph manager exporter endpoints
+{% if groups['mgrs'] | length > 0 %}
+prometheus_ceph_mgr_exporter_endpoints:
+{% for host in groups['mgrs'] %}
+  - "{{ admin_oc_net_name | net_ip(host) }}:9283"
+{% endfor %}
+{% endif %}
 
 #############################################################################
 
--- a/releasenotes/notes/add-ceph-mgr-exporter-endpoints-f7d3c8b1131f6982.yaml
+++ b/releasenotes/notes/add-ceph-mgr-exporter-endpoints-f7d3c8b1131f6982.yaml
@@ -1,0 +1,8 @@
+---
+features:
+  - |
+    Adds code to the globals.yml file to add endpoints for the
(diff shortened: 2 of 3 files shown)


releasenotes/notes/add-ceph-mgr-exporter-endpoints-f7d3c8b1131f6982.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
```
markgoddard, line 3: It still requires you to set the flag to enable ceph monitoring in prometheus.

markgoddard: Which is correct, but the release note should mention it.

markgoddard (changes requested):
Could you update the monitoring docs too?

releasenotes/notes/add-ceph-mgr-exporter-endpoints-f7d3c8b1131f6982.yaml
technowhizz: Sorted :)

technowhizz:
> Could you update the monitoring docs too?

Done!

markgoddard (approved):

etc/kayobe/kolla/globals.yml
```diff
@@ -77,6 +77,13 @@ prometheus_node_exporter_extra_volumes:
   - "textfile:/var/lib/node_exporter/textfile_collector"
 prometheus_node_exporter_cmdline_extras: "--collector.textfile.directory=/var/lib/node_exporter/textfile_collector"
 
+# Set Ceph manager exporter endpoints
+{% if groups['mgrs'] | length > 0 %}
+prometheus_ceph_mgr_exporter_endpoints:
+{% for host in groups['mgrs'] %}
+  - "{{ admin_oc_net_name | net_ip(host) }}:9283"
```
Alex-Welsh, line 84: Have you considered making the port number a variable? I don't know enough about it to actually know if it's a sensible idea.

technowhizz: Its the standard port for the ceph dashboard. Also I dont think its exists as a variable within KA (unlike something like `prometheus_node_exporter_port` for example).

MoteHue: This port is the default, and I don't think it is ever changed anywhere. The variable to change this port in KA for example was removed a while ago: https://github.com/openstack/kolla-ansible/commit/4e6fe7a6da4d5471cd0425c781f25f3e9567d3a0

markgoddard (approved):

PR #537, opened 2023-06-19
Fix loop label syntax in cephadm-gather-keys.yml

flowerysong:
Loop labels are required to be strings. This was not properly enforced in previous versions of Ansible, but starting in ansible-core 2.15 this will cause a parse-time error.

```
ERROR! The field 'label' is supposed to be a string type, however the incoming data structure is a <class 'ansible.parsing.yaml.objects.AnsibleMapping'>

The error appears to be in 'cephadm-gather-keys.yml': line 46, column 9, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

      loop_control:
        label:
        ^ here
```


markgoddard (approved):
Thanks!

PR #538, opened 2023-06-20
Add playbooks to allow hotfixing containers

MoteHue:


Alex-Welsh:
Added docs

Alex-Welsh:
For the record, I'm not happy with how regex is handled, and the way it needs to match the entire name not just a substring. I don't know of a better way to do it though so any suggestions are more than welcome

JohnGarbutt:
So it's a bit late for this comment, but we could generate an image with a thin delta that is just the proposed hot fix, then push that image everywhere? We get to more easily track where it is applied then

Alex-Welsh:
> So it's a bit late for this comment, but we could generate an image with a thin delta that is just the proposed hot fix, then push that image everywhere? We get to more easily track where it is applied then

I'd be interested to hear other opinions but in my mind it might be an "as well as" rather than an "instead of". This playbook should really only be a last resort if we urgently need to get something out.

I can see the two different workflows being: 
Major issue discovered -> hotfix with this playbook -> get longer term fix into release train.
Minor issue discovered -> generate new image & diff -> roll out to deployments

MoteHue:
> in my mind it might be an "as well as" rather than an "instead of"

I'd agree with this. This hotfix playbook isn't meant to be the one true solution to urgently resolving a critical issue/vulnerability.

doc/source/operations/hotfix-playbook.rst
```diff
@@ -0,0 +1,69 @@
...
+Example Variables file
+----------------------
+
+.. code-block:: yaml
+
+    ---
+    container_hotfix_command: "/tmp/quick-fix.sh"
+    container_hotfix_files:
+      - src: "~/quick-fix.sh"
+        dest: "/tmp/quick-fix.sh"
```
jovial, line 64: Possibly you also need `mode` to optionally make these executable?

Alex-Welsh: Good point. Mode option added :+1:

markgoddard:
> So it's a bit late for this comment, but we could generate an image with a thin delta that is just the proposed hot fix, then push that image everywhere? We get to more easily track where it is applied then

It's an interesting alternative, since it reduces the risk of losing the temporary changes in the hotfix. I think the (your) intention here was for this to allow getting fixes out for zero day issues as fast as possible. If we need to build and distribute an image then it's going to take longer, and at some (unknown) point we might as well build images from scratch and do it "properly".

etc/kayobe/ansible/hotfix-containers.yml
```diff
@@ -0,0 +1,66 @@
...
+        - container_hotfix_container_regex == "" or
+          container_hotfix_command == "" and container_hotfix_files == []
+
+    - name: Get list of containers to hotfix
+      command: '{{ kolla_container_engine | default("docker")}} ps --format {% raw %}"{{.Names}}"{% endraw %}'
+      register: host_containers
+
+    - name: Fail if no containers match given regex
+      vars:
+        hotfix_containers: host_containers | split('\n') | regex_search(container_hotfix_container_regex)
```
markgoddard:
```suggestion
        hotfix_containers: "{{ host_containers | split('\n') | regex_search(container_hotfix_container_regex) }}"
```

markgoddard: Probably this needs to reference `.stdout`, same as L54.

markgoddard: Might be easier to set_fact once for the container list.

markgoddard (changes requested):
Like it! Just one issue with the fail task.

etc/kayobe/ansible/hotfix-containers.yml
MoteHue: Sounds like a good idea, done :+1:

markgoddard (approved):

PR #540, opened 2023-06-22
Add LVM config for AIOs

Alex-Welsh:
Adds an lvm configuration for the ci-aio environment and updates the automated setup script with an --lvm flag for additional configuration.

A minimum 30gb root disk is recommended

### DIFF
--- a/etc/kayobe/environments/ci-aio/controllers.yml
+++ b/etc/kayobe/environments/ci-aio/controllers.yml
@@ -5,3 +5,8 @@
 # User with which to access the controllers via SSH during bootstrap, in order
 # to setup the Kayobe user account. Default is {{ os_distribution }}.
 controller_bootstrap_user: "{{ os_distribution if os_distribution == 'ubuntu' else 'cloud-user' }}"
+
+# Controller lvm configuration. See intentory/group_vars/controllers/lvm.yml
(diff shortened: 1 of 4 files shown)


etc/kayobe/environments/ci-aio/controllers.yml
```diff
@@ -5,3 +5,8 @@
 # User with which to access the controllers via SSH during bootstrap, in order
 # to setup the Kayobe user account. Default is {{ os_distribution }}.
 controller_bootstrap_user: "{{ os_distribution if os_distribution == 'ubuntu' else 'cloud-user' }}"
+
+# Controller lvm configuration. See intentory/group_vars/controllers/lvm.yml
+# for the exact configuration.
+controller_lvm_groups:
+  - "{{ stackhpc_lvm_group_rootvg }}"
```
markgoddard, line 12: This will make CI jobs fail if the image is not using LVM

Alex-Welsh: I'll hold off until we're using LVM images in CI then

Alex-Welsh:
This will make CI jobs fail if the image is not using LVM. The config will be included in the same PR that switches to using LVM images

PR #541, opened 2023-06-23
Add opensearch to Rocky AIOs/Multinodes

Alex-Welsh:
Updating the Rocky opensearch tag and enabling centralised logging with opensearch in Rocky AIOs.
I haven't included a release note but can do if it's deemed necessary

### DIFF
--- a/etc/kayobe/environments/ci-aio/stackhpc-ci.yml
+++ b/etc/kayobe/environments/ci-aio/stackhpc-ci.yml
@@ -5,9 +5,6 @@
 # Docker namespace to use for Kolla images. Default is 'kolla'.
 kolla_docker_namespace: stackhpc-dev
 
-# Do not enable central logging if using Rocky Linux 9. It does not have elasticsearch
-kolla_enable_central_logging: "{{ os_release != '9' }}"
-
 ###############################################################################
 # Network configuration.
 
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -45,7 +45,7 @@ kayobe_image_tags:
     ubuntu: yoga-20230220T181235
   opensearch:
     centos: yoga-20230324T084510
-    rocky: yoga-20230324T090413
+    rocky: yoga-20230622T154317
     ubuntu: yoga-20230324T090345
   prometheus_node_exporter:
     centos: yoga-20230310T173747


Alex-Welsh:
Added the same for the Multinode environment

etc/kayobe/environments/ci-aio/stackhpc-ci.yml
```diff
@@ -5,8 +5,9 @@
 # Docker namespace to use for Kolla images. Default is 'kolla'.
 kolla_docker_namespace: stackhpc-dev
 
-# Do not enable central logging if using Rocky Linux 9. It does not have elasticsearch
-kolla_enable_central_logging: "{{ os_release != '9' }}"
+# Rocky Linux 9 does not have elasticsearch, use opensearch instead.
+kolla_enable_elasticsearch: "{{ os_release != '9' }}"
```
markgoddard: Let's go back to central logging for !RL9, since it enables kibana too.

mnasiadka: s/kibana/opensearch-dashboards/

mnasiadka: basically the default for Rocky9 on Yoga (upstream) is to use OpenSearch stack (and use Elasticsearch everywhere else) - so kolla_enable_central_logging: true should be enough

markgoddard: Actually we have kolla_enable_central_logging=true in kolla.yml. Let's make these changes there, so that they apply everywhere.

We also need to make sure that there's a sensible way for users to switch to opensearch on CS8/RL8.

markgoddard: > basically the default for Rocky9 on Yoga (upstream) is to use OpenSearch stack (and use Elasticsearch everywhere else) - so kolla_enable_central_logging: true should be enough

I'm not sure that's quite true:

```
enable_central_logging: "no"
enable_elasticsearch: "{{ 'yes' if not enable_opensearch | bool and (enable_central_logging | bool or enable_osprofiler | bool or enable_skydive | bool or enable_monasca | bool or (enable_cloudkitty | bool and cloudkitty_storage_backend == 'elasticsearch')) else 'no' }}"
enable_kibana: "{{ 'yes' if (enable_central_logging | bool and not enable_opensearch | bool) or enable_monasca | bool else 'no' }}"
enable_opensearch: "{{ kolla_base_distro == 'rocky' and enable_central_logging | bool }}"
```

So if you enable central logging on RL9 you get elastic and opensearch. Maybe this needs to be fixed upstream though?

markgoddard: Hold on, I missed the `not enable_opensearch | bool` bit. I think @mnasiadka is correct then.

Alex-Welsh: I've just removed the statement disabling it for rocky, which if I understand this correctly should now end up with central logging being enabled for all AIOs & Multinodes, with Rocky 9s using opensearch/opensearch dashboards rather than elasticsearch/kibana

markgoddard (approved):

etc/kayobe/kolla/globals.yml
```diff
@@ -45,7 +45,7 @@ kayobe_image_tags:
     ubuntu: yoga-20230220T181235
   opensearch:
     centos: yoga-20230324T084510
-    rocky: yoga-20230324T090413
+    rocky: yoga-20230622T154317
```
markgoddard, line 48: Remember to promote.

Alex-Welsh: Done: https://github.com/stackhpc/stackhpc-release-train/actions/runs/5376938075/jobs/9754763318

PR #542, opened 2023-06-23
Update/smartmon playbook

technowhizz:
Ensure that cron is installed before attempting to configure crontab

### DIFF
--- a/etc/kayobe/ansible/smartmon-tools.yml
+++ b/etc/kayobe/ansible/smartmon-tools.yml
@@ -2,12 +2,13 @@
 - hosts: overcloud
 
   tasks:
-    - name: Ensure smartmon-tools and nvme-cli is installed
+    - name: Ensure smartmon-tools, jq, nvme-cli and cron/cronie is installed
       package:
         name:
           - smartmontools
           - nvme-cli
           - jq
+          - "{{ 'cron' if ansible_facts['distribution'] == 'Ubuntu' else 'cronie' }}"
         state: present
       become: true
 
--- a/releasenotes/notes/update-smartmon-playbook-6bdf9624c579d2e2.yaml
+++ b/releasenotes/notes/update-smartmon-playbook-6bdf9624c579d2e2.yaml
@@ -1,0 +1,5 @@
+---
+upgrade:
+  - |
+    Updates the smartmon-tools.yml playbook to ensure that cron is installed
+    before attempting to configure crontab.


etc/kayobe/ansible/smartmon-tools.yml
```diff
@@ -23,6 +23,12 @@
         - nvmemon.sh
       become: yes
 
+    - name: Ensure cron is installed
+      ansible.builtin.package:
+        name: cron
```
jovial: Is the package `cronie` on RedHat?

technowhizz: Yeah it is. Thank you!

etc/kayobe/ansible/smartmon-tools.yml
```diff
@@ -2,12 +2,13 @@
...
 
   tasks:
-    - name: Ensure smartmon-tools and nvme-cli is installed
+    - name: Ensure smartmon-tools, jq, nvme-cli and cron/cronie is installed
       package:
         name:
           - smartmontools
           - nvme-cli
           - jq
+          - "{{ 'cron' if ansible_facts['distribution'] == 'Ubuntu' else 'cronie' }}"
```
Alex-Welsh, line 11: At what point will this be run? will you always have access to ansible_facts? could you use os_distribution?

jovial, line 11: nit: using ansible_facts['os_family'] == 'Debian' might be more generic, but given that that we only support Ubuntu/Redhat this looks OK.

jovial (approved):

etc/kayobe/ansible/smartmon-tools.yml
technowhizz: we're not setting gather_facts: false so facts should be available

etc/kayobe/ansible/smartmon-tools.yml
```diff
@@ -2,12 +2,13 @@
 - hosts: overcloud
 
   tasks:
-    - name: Ensure smartmon-tools and nvme-cli is installed
+    - name: Ensure smartmon-tools, jq, nvme-cli and cron/cronie is installed
```
markgoddard, line 5: This is why I usually just put "packages" or "package dependencies" :)

markgoddard (approved):

etc/kayobe/ansible/smartmon-tools.yml
technowhizz: Ah. Will bear in mind for next time

PR #544, opened 2023-06-26
Fix Rocky CI

Alex-Welsh:


Alex-Welsh:
CI should now all be green. I have seen some terraform failures but I think that's an SMS issue. Rerunning is usually successful

bbezak (approved):

etc/kayobe/kolla/globals.yml
```diff
@@ -6,8 +6,6 @@
 # non-overcloud hosts
 enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names or ansible_facts.os_family == 'Debian' }}{% endraw %}"
 
-kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"
```
markgoddard, line 9: Wasn't this added to support the Rocky 9 migration, deploying rocky images on rocky9 hosts? Let's remove it for now but we will need something similar.

markgoddard (approved):

etc/kayobe/kolla/globals.yml
Alex-Welsh: Ah, of course. I couldn't remember why I had added that. I'll put in a better solution once this passes

PR #546, opened 2023-06-27
Enable Designate and configure FQDNs for APIs

jackhodgkiss:
Enable Designate for (DNSaaS) with a default backend of Bind9.

Also take the opportunity to configure FQDNs for the internal/external APIs. These must be resolvable prior to deployment either through use of existing DNS provider or more likely to rely upon entries within `/etc/hosts`.


etc/kayobe/environments/ci-multinode/kolla.yml
```diff
@@ -5,7 +5,18 @@ kolla_enable_neutron_provider_networks: true
 kolla_enable_ovn: true
 kolla_enable_octavia: true
 kolla_enable_magnum: true
+kolla_enable_designate: true
+kolla_enable_horizon_designate: true
```
mnasiadka: that's the default, isn't it?

jackhodgkiss: I just checked, it will be enabled if designate is enabled.

etc/kayobe/environments/ci-multinode/networks.yml
```diff
@@ -78,13 +78,15 @@ internal_allocation_pool_start: 192.168.37.3
...
 internal_vlan: 101
+internal_fqdn: "internal.infra.mos.{{ root_domain | default(example.com) }}"
 
 # External network
 external_cidr: 192.168.38.0/24
 external_mtu: "{{ ansible_facts.default_ipv4.mtu - 50 }}"
 external_allocation_pool_start: 192.168.38.3
 external_allocation_pool_end: 192.168.38.128
 external_vlan: 102
+external_fqdn: "external.infra.mos.{{ root_domain | default(example.com) }}"
```
markgoddard: I don't think this one is used.

etc/kayobe/environments/ci-multinode/kolla.yml
```diff
@@ -5,7 +5,15 @@ kolla_enable_neutron_provider_networks: true
...
 
 # The multinode environment supports Manila but it is not enabled by default.
 # kolla_enable_manila: true
 # kolla_enable_manila_backend_cephfs_native: true
+
+kolla_internal_vip_address: "{{ internal_vip_address }}"
+kolla_internal_fqdn: "{{ internal_fqdn }}"
+
+kolla_external_vip_address: "{{ public_vip_address }}"
+kolla_external_fqdn: "{{ public_fqdn }}"
```
markgoddard: These variables are the newer alternatives to the <net>_vip_address and <net>_fqdn variables. If you want to use these it would be better to remove the others from networks.yml

etc/kayobe/environments/ci-multinode/kolla/globals.yml
```diff
@@ -25,3 +25,12 @@ octavia_provider_agents: "ovn"
 
 # Manila CephFS configuration
 manila_cephfs_filesystem_name: manila-cephfs
+
+# Designate configuration
+designate_ns_record:
+  - "ns1.mos.{{ root_domain | default(example.com) }}"
+  - "ns2.mos.{{ root_domain | default(example.com) }}"
+  - "ns3.mos.{{ root_domain | default(example.com) }}"
```
markgoddard:
```suggestion
  - "ns1.mos.{{ root_domain | default('example.com') }}"
  - "ns2.mos.{{ root_domain | default('example.com') }}"
  - "ns3.mos.{{ root_domain | default('example.com') }}"
```

markgoddard: Do these need to be resolvable?

markgoddard (changes requested):

etc/kayobe/environments/ci-multinode/kolla/globals.yml
jackhodgkiss: The name servers don't need to be resolvable. This would be important if we wanted/needed authoritative results however the authoritative names for `sms-lab.cloud` are the controllers on sms-lab which cannot reach the multinode controllers.

etc/kayobe/environments/ci-multinode/networks.yml
jackhodgkiss: Removed.

etc/kayobe/environments/ci-multinode/dns.yml
```diff
@@ -0,0 +1,8 @@
+---
+resolv_is_managed: true
+
+# List of DNS nameservers.
+resolv_nameservers:
+  - 192.168.39.4
+  - 192.168.39.5
+  - 192.168.39.6
```
markgoddard: Isn't there a bootstrapping issue here? DNS will be broken until designate is deployed.

Which hosts actually need to use Designate as their DNS server?

jackhodgkiss: None of the hosts within the multinode environment _require_ Designate as their DNS server.

If we set contents of `/etc/hosts` during the bootstrap phase within the terraform then the controllers and other nodes will have no issues.

The problem would be if someone was to use the multinode environment to deploy some service and used Designate to create the records. This would mean that all nodes within the environment wouldn't be aware of this change as the DNS is still point at Cloudflare/Google.

Maybe a warning to indicate that changes to dns.yml and host configure are required after successful deployment if the user requires certain nodes to have access to Designate such as Infra VMs or seed node.

The only other solution would be something such as DNSMASQ which would provide split DNS enabling everything but `mos.{{ root_domain }}` to go to Cloudflare/google whilst `mos.{{ root_domain` would be answered by `/etc/hosts` and eventually Designate.

markgoddard: I'd prefer to just document how to access the Designate DNS servers. Typically controllers and hypervisors would not need to access a service deployed on the cloud. The Ansible control host might be used for manual testing, and the seed runs tempest, which could potentially test designate.

etc/kayobe/environments/ci-multinode/networks.yml
```diff
@@ -76,22 +76,19 @@ internal_cidr: 192.168.37.0/24
...
 internal_allocation_pool_start: 192.168.37.3
 internal_allocation_pool_end: 192.168.37.254
-internal_vip_address: 192.168.37.2
 internal_vlan: 101
 
 # External network
 external_cidr: 192.168.38.0/24
 external_mtu: "{{ ansible_facts.default_ipv4.mtu - 50 }}"
 external_allocation_pool_start: 192.168.38.3
-external_allocation_pool_end: 192.168.38.128
```
Alex-Welsh: From external conversation: this removal was not intended

markgoddard (commented):
LGTM, let's just wait for the TF PR to complete.

markgoddard (approved):

PR #551, opened 2023-06-29
Add magnum container footer

scrungus:
the new cluster api driver requires helm to be installed in the containers. this is a temporary place for it to be installed until it is added to upstream kolla containers

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -339,7 +339,8 @@ kolla_build_blocks:
     RUN sed -i 's/"pc-q35-rhel8.5.0"/"pc-q35-*"/' /usr/share/qemu/firmware/50-edk2-ovmf-cc.json
     {% endif %}
     {% endraw %}
-
+  magnum_base_footer: |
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla.yml
```diff
@@ -339,7 +339,8 @@ kolla_build_blocks:
     RUN sed -i 's/"pc-q35-rhel8.5.0"/"pc-q35-*"/' /usr/share/qemu/firmware/50-edk2-ovmf-cc.json
     {% endif %}
     {% endraw %}
-
+  magnum_base_footer: |
+    RUN curl -fsL https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | head -n -1 | bash
```
mnasiadka: Kolla base image already has curlrc which has relevant flags in, no need to pass them here again.

markgoddard (approved):

markgoddard:
If you want this to be included in an image you'll need to build one and update the tag in etc/kayobe/kolla/globals.yml. Do you need help with that?

scrungus:
@markgoddard Should be all good, I've done it before. I will submit the new tag as part of an incoming PR with changes to CI too

PR #553, opened 2023-07-03
Fix overcloud host image promotion workflow

Alex-Welsh:
Allows parallel promotion now that the images can be built with the same tags, fixes a problem with the OS selection, and removes the host configure to speed up the process by a fair margin.


Alex-Welsh:
This works for now but I'm open to changing it in the future so that it draws tags from a branch rather than inputs

markgoddard (approved):

PR #556, opened 2023-07-06
[DEV-1553] added logrotate container to seed node

g0rgamesh:
blocked by: https://review.opendev.org/c/openstack/kayobe/+/887792


jovial (commented):
I can see this being pretty useful for custom containers, but shouldn't kolla ship some logrotation stuff for bifrost?

g0rgamesh:
> I can see this being pretty useful for custom containers, but shouldn't kolla ship some logrotation stuff for bifrost?

Kolla is not used by kayobe when deploying seed and it's containers.

markgoddard:
> I can see this being pretty useful for custom containers, but shouldn't kolla ship some logrotation stuff for bifrost?

I agree this would be better living in the kolla-ansible [bifrost role](https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/bifrost). I don't know whether it's worth the effort to reimplement there though.

MoteHue:
> Kolla is not used by kayobe when deploying seed and it's containers.

It isn't used for most containers (eg Pulp, Squid), but Kolla is used for deploying Bifrost on the seed.

etc/kayobe/seed.yml
```diff
@@ -137,6 +137,30 @@ seed_squid_container:
       - /srv/docker/squid/squid.conf:/etc/squid/squid.conf
       - /var/log/squid:/var/log/squid
 
+# Seed container running a cron service.
+seed_cron_container_enabled: true
+
+seed_cron_container:
+  cron:
+    container_name: cron
+    docker_registry: "{{ lookup('vars', admin_oc_net_name ~ '_ips')[groups.seed.0] }}:{{ pulp_port }}"
```
markgoddard: I don't think this is used.

etc/kayobe/seed.yml
```diff
@@ -137,6 +137,30 @@ seed_squid_container:
...
+    image: "{{ docker_registry ~ '/' if docker_registry else '' }}stackhpc/{{ kolla_base_distro }}-source-cron"
+    registry_username: "{{ pulp_username }}"
+    registry_password: "{{ pulp_password }}"
```
markgoddard: nor these

etc/kayobe/seed.yml
```diff
@@ -137,6 +137,30 @@ seed_squid_container:
...
       - /var/log/squid:/var/log/squid
 
+# Seed container running a cron service.
+seed_cron_container_enabled: true
+
+seed_cron_container:
+  cron:
+    container_name: cron
+    docker_registry: "{{ lookup('vars', admin_oc_net_name ~ '_ips')[groups.seed.0] }}:{{ pulp_port }}"
+    image: "{{ docker_registry ~ '/' if docker_registry else '' }}stackhpc/{{ kolla_base_distro }}-source-cron"
```
markgoddard:
```suggestion
    image: "{{ kolla_docker_registry ~ '/' if kolla_docker_registry else '' }}{{ kolla_docker_namespace }}/{{ kolla_base_distro }}-source-cron"
```

etc/kayobe/seed.yml
```diff
@@ -137,6 +137,30 @@ seed_squid_container:
...
+    registry_username: "{{ pulp_username }}"
+    registry_password: "{{ pulp_password }}"
+    pre: "{{ kayobe_config_path }}/containers/cron/pre.yml"
+    tag: "yoga-20230220T181235"
```
markgoddard, line 148: This tag is the main part that concerns me.

1. It won't necessarily be the same for each distro
2. It's in a different place in the config from the other kolla tags, so could easily get out of sync

markgoddard (changes requested):

g0rgamesh:
> > I can see this being pretty useful for custom containers, but shouldn't kolla ship some logrotation stuff for bifrost?
> 
> I agree this would be better living in the kolla-ansible [bifrost role](https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/bifrost). I don't know whether it's worth the effort to reimplement there though.

Why narrow it only for bifrost when logs on all containers should be rotated?

markgoddard:
> > > I can see this being pretty useful for custom containers, but shouldn't kolla ship some logrotation stuff for bifrost?
> > 
> > 
> > I agree this would be better living in the kolla-ansible [bifrost role](https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/bifrost). I don't know whether it's worth the effort to reimplement there though.
> 
> Why narrow it only for bifrost when logs on all containers should be rotated?

This PR only appears to cover bifrost services

g0rgamesh:
> > > > I can see this being pretty useful for custom containers, but shouldn't kolla ship some logrotation stuff for bifrost?
> > > 
> > > 
> > > I agree this would be better living in the kolla-ansible [bifrost role](https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/bifrost). I don't know whether it's worth the effort to reimplement there though.
> > 
> > 
> > Why narrow it only for bifrost when logs on all containers should be rotated?
> 
> This PR only appears to cover bifrost services

When other seed containers start to store their logs in `/var/lib/docker/volumes/kolla_logs/_data/` it will be a matter of just updating logrotate config.

markgoddard:
> When other seed containers start to store their logs in `/var/lib/docker/volumes/kolla_logs/_data/` it will be a matter of just updating logrotate config.

Looks like squid logs to files in `/var/log/squid/`.

FWIW this would be possible in kolla-ansible also. Either pass in a list of files as a variable, or allow providing additional logrotate config files.

etc/kayobe/seed.yml
g0rgamesh: Thanks, tests were OK with those modifications.

g0rgamesh:
> > When other seed containers start to store their logs in `/var/lib/docker/volumes/kolla_logs/_data/` it will be a matter of just updating logrotate config.
> 
> Looks like squid logs to files in `/var/log/squid/`.
> 
> FWIW this would be possible in kolla-ansible also. Either pass in a list of files as a variable, or allow providing additional logrotate config files.

Added squid logs volume mount.

etc/kayobe/seed.yml
g0rgamesh: Yes that was my concern as well, but as kolla is not used during this action I couldn't pull it from kolla configs.
Would just putting a filter for specific kolla globals.yml be approperiate?

markgoddard: I don't think that could work currently. globals.yml isn't necessarily valid YAML, or even templated to valid YAML

g0rgamesh: I guess for now we could document that tag should be changed during the deployment in that case.

etc/kayobe/containers/cron/files/logrotate.conf
```diff
@@ -0,0 +1,32 @@
...
+"/var/log/kolla/ironic/*.log"
+{
+}
+"/var/log/kolla/ironic-inspector/*.log"
+{
+}
+"/var/log/kolla/nginx/*.log"
+{
+}
+"/var/log/kolla/squid/*.log"
```
jovial, line 30: Little bit strange have this one under /var/log/kolla since it isn't a kolla service. Could we just bind mount it to /var/log/squid instead? Possibly needs different setting in place of `su root kolla`?

mnasiadka:
This is nice feature, but I think it would be better to leverage the common role in kolla-ansible to do that on the seed vm/infra vms.
Current deploy-containers functionality does not support any handlers for changed files (e.g. if you update log rotate.conf - it's not going to restart the container at all).

markgoddard:
> This is nice feature, but I think it would be better to leverage the common role in kolla-ansible to do that on the seed vm/infra vms. Current deploy-containers functionality does not support any handlers for changed files (e.g. if you update log rotate.conf - it's not going to restart the container at all).

I like the idea, but I think the common role would require a lot of changes to make it suitable - it bundles fluentd and kolla_toolbox, and the cron config is not customisable.

I think a change in bifrost to install cron would make sense and benefit other bifrost users.

markgoddard:
Moved to bifrost, backported to Antelope and Zed: https://review.opendev.org/q/I7c2ba2b263dd84641674b60aed8971b129aa846f

PR #559, opened 2023-07-12
Support TLS & HCP Vault PKI in the multinode env

Alex-Welsh:
Terraform PR: https://github.com/stackhpc/terraform-kayobe-multinode/pull/27

### DIFF
--- a/doc/source/configuration/vault.rst
+++ b/doc/source/configuration/vault.rst
@@ -36,6 +36,8 @@ Before beginning the deployment of vault for openstack internal TLS and backend 
 
   * Seed Node or a host to run the vault container on
   * Overcloud controller hosts to install second vault on
+  * Ansible Galaxy dependencies installed: ``kayobe control host bootstrap``
+  * Python dependencies installed: ``pip install -r kayobe-config/requirements.txt``
 
 Deployment
 ==========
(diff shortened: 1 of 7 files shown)


doc/source/configuration/vault.rst
```diff
@@ -36,6 +36,8 @@ Before beginning the deployment of vault for openstack internal TLS and backend
 
   * Seed Node or a host to run the vault container on
   * Overcloud controller hosts to install second vault on
+  * Ansible Galaxy dependencies installed: ``ansible-galaxy install ansible-galaxy install -r $KAYOBE_CONFIG_PATH/anisble/requirements.yml``
```
markgoddard: These should be installed using `kayobe control host bootstrap|upgrade`

doc/source/configuration/vault.rst
```diff
@@ -36,6 +36,8 @@ Before beginning the deployment of vault for openstack internal TLS and backend
...
+  * Python dependencies installed: ``pip install -r kayobe-config/requirements.txt``
```
markgoddard, line 40: This one's a bit awkward, I wish we had a cleaner way to pull it in.

markgoddard (changes requested):

doc/source/configuration/vault.rst
Alex-Welsh: post hook on control host bootstrap?

markgoddard: Could do. It also includes Kayobe though, and I'd be wary of overwriting an existing installation by surprise. Would ideally like something more "official". Will tried to do it years ago but gave up for reasons I forget. I think this is ok for now.

doc/source/configuration/vault.rst
Alex-Welsh: I've gone with bootstrap, since I don't really know how upgrade works. Can include options for both if needed

markgoddard (approved):

PR #561, opened 2023-07-14
Add support for configuring OpenSearch log retention policies

dougszumski:
There are a number of API driven operations that we wish to automate in OpenSearch. This commit adds support for configuring a log retention policy, if one doesn't exist.

The retention policy is created during overcloud post configure and can be customised via a new settings file: `opensearch.yml`.

In the future, this role will be expanded to cover more API driven operations. For example, loading of dashboards and  alerting on log signatures.


etc/kayobe/ansible/roles/opensearch/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- import_tasks: set_log_retention_policy.yml
+  when:
+    - kolla_enable_central_logging | bool
```
Alex-Welsh, line 4: Would this be useful when central logging is disabled but opensearch is not?

etc/kayobe/ansible/roles/opensearch/tasks/set_log_retention_policy.yml
```diff
@@ -0,0 +1,58 @@
...
+
+- name: Check if a log retention policy exists
+  become: true
+  uri:
+    url: "{{ opensearch_internal_endpoint }}/_plugins/_ism/policies/retention"
+    method: GET
+    status_code: 200, 404
+    return_content: yes
+  register: opensearch_retention_policy_check
+  run_once: true
```
Alex-Welsh, line 29: ditto

etc/kayobe/ansible/roles/opensearch/tasks/set_log_retention_policy.yml
```diff
@@ -0,0 +1,58 @@
...
+- name: Wait for OpenSearch to become ready
+  become: true
+  uri:
+    url: "{{ opensearch_internal_endpoint }}/_cluster/stats"
+    status_code: 200
+  register: result
+  until: result.get('status') == 200
+  retries: 30
+  delay: 2
+  run_once: true
```
Alex-Welsh, line 19: This should probably include changed_when: false, since it will never change anything but always runs.

dougszumski: I agree in general. I think in this case it's handled internally by the module. Eg. if the service is running, it looks like this:

image

If the service is down for a bit:

image

etc/kayobe/ansible/roles/opensearch/tasks/main.yml
dougszumski: Hmm, the only case I could think of would be if CloudKitty was enabled, had been patched to use daily indices, and central logging was off. I think it's very unlikely. Maybe I missed another use case?

There is another corner case to consider if an external ES/OS cluster is being used. I.e. you could have `enable_opensearch: False` and `enable_central_logging: True`. I think in that case, you would still likely want a retention policy.

dougszumski (commented):
Thanks for reviewing @Alex-Welsh

Alex-Welsh:
LGTM

Alex-Welsh (dismissed):

etc/kayobe/ansible/opensearch_configure.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+- hosts: localhost
+  tasks:
+    - include_role:
+        name: opensearch
+      when: kolla_enable_central_logging | bool
```
markgoddard: Not all deployments will be using opensearch.
```suggestion
      when: kolla_enable_opensearch | bool
```

etc/kayobe/ansible/roles/opensearch/tasks/main.yml
markgoddard: As before - use opensearch flag.

etc/kayobe/ansible/opensearch_configure.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+- hosts: localhost
```
markgoddard: Often the Ansible control host doesn't have access to the internal API network. For that reason we tend to run on one of the controllers during post configure tasks.

etc/kayobe/ansible/roles/opensearch/tasks/set_log_retention_policy.yml
```diff
@@ -0,0 +1,58 @@
...
+    msg: "Please set the variable {{ item.key }} in opensearch.yml. For example: {{ item.key }}: {{ item.value }} "
+  when: "{{ item.key }} is not defined"
+  with_dict:
+    - "opensearch_internal_endpoint": '"http://{{ kolla_internal_fqdn }}:9200"'
+    - "opensearch_log_index_pattern": '"flog-*"'
+
+- name: Wait for OpenSearch to become ready
+  become: true
+  uri:
+    url: "{{ opensearch_internal_endpoint }}/_cluster/stats"
```
markgoddard, line 13: Running on controllers should also (usually) avoid the issue of CA certs for internal TLS, since the vault playbooks add them to the trust store on controllers.

markgoddard (changes requested):
How does this relate to the upstream support proposed in https://review.opendev.org/c/openstack/kolla-ansible/+/867118?

etc/kayobe/ansible/opensearch_configure.yml
dougszumski: Thanks - I'm fine with changing it. I'd assumed that anyone using this config would be deploying the reference monitoring stack.

dougszumski:
> How does this relate to the upstream support proposed in https://review.opendev.org/c/openstack/kolla-ansible/+/867118?

Upstream is going to use the UI directly.

markgoddard:
> > How does this relate to the upstream support proposed in https://review.opendev.org/c/openstack/kolla-ansible/+/867118?
> 
> Upstream is going to use the UI directly.

That's a shame. Who wants to point and click? I see we've been over this in the patch though.

mnasiadka:
> > > How does this relate to the upstream support proposed in https://review.opendev.org/c/openstack/kolla-ansible/+/867118?
> > 
> > 
> > Upstream is going to use the UI directly.
> 
> That's a shame. Who wants to point and click? I see we've been over this in the patch though.

That's not true, there was a discussion in that patch why the approach is to set it once, and if the user wants to change it - he needs to change it in the UI.
Doug was persistent on his approach - which is at least suboptimal, I'm pretty sure if the approach would change - we can easily merge it.

markgoddard:
> > > How does this relate to the upstream support proposed in https://review.opendev.org/c/openstack/kolla-ansible/+/867118?
> > 
> > 
> > Upstream is going to use the UI directly.
> 
> That's a shame. Who wants to point and click? I see we've been over this in the patch though.

Maybe I missed some IRC chat, but what I don't understand is how we (StackHPC) had a disagreement over how to do this upstream and resolved it by doing the same thing downstream.

markgoddard (changes requested):
Discussion needs resolution.

dougszumski:
> 



> Maybe I missed some IRC chat, but what I don't understand is how we (StackHPC) had a disagreement over how to do this upstream and resolved it by doing the same thing downstream.



> > > > How does this relate to the upstream support proposed in https://review.opendev.org/c/openstack/kolla-ansible/+/867118?
> > > 
> > > 
> > > Upstream is going to use the UI directly.
> > 
> > 
> > That's a shame. Who wants to point and click? I see we've been over this in the patch though.
> 
> That's not true, there was a discussion in that patch why the approach is to set it once, and if the user wants to change it - he needs to change it in the UI. Doug was persistent on his approach - which is at least suboptimal, I'm pretty sure if the approach would change - we can easily merge it.

bbezak:
upstream change merged upstream - https://review.opendev.org/c/openstack/kolla-ansible/+/867118

PR #570, opened 2023-08-01
Support using local hashicorp consul/vault images

priteau:
This requires stackhpc.hashicorp 2.4.0.

### DIFF
--- a/etc/kayobe/ansible/vault-deploy-overcloud.yml
+++ b/etc/kayobe/ansible/vault-deploy-overcloud.yml
@@ -60,10 +60,15 @@
     - import_role:
         name: stackhpc.hashicorp.vault
       vars:
+        hashicorp_registry_url: "{{ overcloud_hashicorp_registry_url }}"
+        hashicorp_registry_username: "{{ overcloud_hashicorp_registry_username }}"
+        hashicorp_registry_password: "{{ overcloud_hashicorp_registry_password }}"
+        consul_docker_image: "{{ overcloud_consul_docker_image }}"
         consul_docker_tag: "{{ overcloud_consul_docker_tag }}"
         vault_config_dir: "/opt/kayobe/vault"
         vault_cluster_name: "overcloud"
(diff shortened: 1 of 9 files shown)


mnasiadka (dismissed):

etc/kayobe/ansible/vault-deploy-overcloud.yml
```diff
@@ -60,10 +60,15 @@
     - import_role:
         name: stackhpc.hashicorp.vault
       vars:
+        hashicorp_registry_url: "{{ overcloud_hashicorp_registry_url }}"
```
markgoddard, line 63: Would it make sense to do the same for the seed vault to keep them in sync?

priteau: I have updated the pull request to support local images for the seed. Can you please review again?

jovial (approved):
Looks reasonable to me. Only thought was that if we mirror the images from dockerhub directly, instead of pulling via ark, that we might end of using different versions of the images at different deployments. Not sure if they rebuild the images with the same tags?

priteau:
> Looks reasonable to me. Only thought was that if we mirror the images from dockerhub directly, instead of pulling via ark, that we might end of using different versions of the images at different deployments. Not sure if they rebuild the images with the same tags?

I followed the approach we use for Ceph images. From what I can see on Docker Hub, if we switched to patch-level version (e.g. 1.14.1 instead of 1.14) we should get the same images:

* 1.14.0 was pushed on June 21 (day of 1.14.0 release)
* 1.14.1 was pushed on July 25 (when it was released)
* 1.14 was updated to point to 1.14.1

mnasiadka:
I think it still would be better to sync those to Ark for internet-constrained environments (and counting docker hub pull limits)

markgoddard (approved):

markgoddard:
> I think it still would be better to sync those to Ark for internet-constrained environments (and counting docker hub pull limits)

Perhaps, but it could be done as a separate task.

PR #572, opened 2023-08-01
Bump OVS to version 2.17.7

MoteHue:


MoteHue:
Note: don't merge yet, still needs to be tested then promoted.

Alex-Welsh:
> Note: don't merge yet, still needs to be tested then promoted.

There's an annoyingly small link you can click on the right hand side of a PR to convert it to a draft:

image

PR #573, opened 2023-08-03
Fix backend API certificate generation

priteau:
Some dependencies for generating backend API certificates were missing but were undetected because they were fulfilled by other vault playbooks run earlier during the deployment process.

This was detected when playbook hosts were extended to include separate network hosts running some OpenStack services.

### DIFF
--- a/etc/kayobe/ansible/vault-generate-backend-tls.yml
+++ b/etc/kayobe/ansible/vault-generate-backend-tls.yml
@@ -1,10 +1,42 @@
 ---
+# Required for uri module to work with self-signed certificates and for systems to trust
+# the self-signed CA
+- name: Install CA on controllers
+  hosts: controllers
+  tasks:
+    - name: Copy the intermediate CA
+      copy:
+        src: "{{ kayobe_env_config_path }}/vault/OS-TLS-ROOT.pem"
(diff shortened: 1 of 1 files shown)


etc/kayobe/ansible/vault-generate-backend-tls.yml
```diff
@@ -1,10 +1,42 @@
 ---
+# Required for uri module to work with self-signed certificates and for systems to trust
+# the self-signed CA
+- name: Install CA on network hosts
```
markgoddard: Network hosts? Target is controllers

markgoddard: Maybe we need to expand the target hosts in vault-deploy-overcloud.yml?

priteau: > Network hosts? Target is controllers

My mistake, I forgot to edit when I removed `network` hosts from my changes.

markgoddard (approved):

etc/kayobe/ansible/vault-generate-backend-tls.yml
priteau: > Maybe we need to expand the target hosts in vault-deploy-overcloud.yml?

I have a preference for having explicit dependencies, otherwise we can run into issues when nodes are reprovisioned for example.

PR #583, opened 2023-08-16
Pin Consul and Vault to the latest MPL releases

priteau:


bbezak (approved):

etc/kayobe/inventory/group_vars/all/vault
```diff
@@ -10,13 +10,13 @@ seed_hashicorp_registry_password: "{{ stackhpc_docker_registry_password if stack
 seed_consul_docker_image: "{{ stackhpc_docker_registry ~ '/' if stackhpc_sync_hashicorp_images | bool else '' }}hashicorp/consul"
 
 # Seed Consul container image tag.
-seed_consul_docker_tag: "1.16"
+seed_consul_docker_tag: "1.16.1"
```
markgoddard, line 13: Do we know what will be the first BSL release? If it's 1.17 we could stick with 1.16 here.

bbezak: we know other way around - which are the last MPL ones:

https://www.hashicorp.com/license-faq#products-covered-by-bsl

bbezak: therefore we're assuming that 1.16.2 will be BSL rls, so pinning to 1.16.1

priteau: It's not clear which version will be the first BSL. We did talk about pinning to patch versions anyway, to ensure all our deployments are using the same versions.

markgoddard: I expect they'll bump the minor for BSL so that they can release MPL fixes on patch versions.

priteau: We could still merge this to get exactly the same container, like we did for Pulp in 3e60f9bd4bcf799a9824b85c3149a024b487629b?

markgoddard (approved):

PR #590, opened 2023-08-17
yoga: xena merge

priteau:
- Cap number of Pulp workers
- Fix documentation builds on Read the Docs

### DIFF
--- a/etc/kayobe/seed.yml
+++ b/etc/kayobe/seed.yml
@@ -112,8 +112,8 @@ seed_pulp_container:
     # s6-overlay-suexec starts as pid 1
     init: false
     env:
-      PULP_CONTENT_WORKERS: "{{ ansible_facts.processor_vcpus * 2 + 1 }}"
-      PULP_API_WORKERS: "{{ ansible_facts.processor_vcpus * 2 + 1 }}"
+      PULP_CONTENT_WORKERS: "{{ [ansible_facts.processor_vcpus * 2 + 1, 32] | min }}"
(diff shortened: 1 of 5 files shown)


etc/kayobe/seed.yml
```diff
@@ -112,8 +112,8 @@ seed_pulp_container:
     # s6-overlay-suexec starts as pid 1
     init: false
     env:
-      PULP_CONTENT_WORKERS: "{{ ansible_facts.processor_vcpus * 2 + 1 }}"
-      PULP_API_WORKERS: "{{ ansible_facts.processor_vcpus * 2 + 1 }}"
+      PULP_CONTENT_WORKERS: "{{ [ansible_facts.processor_vcpus * 2 + 1, 32] | min }}"
+      PULP_API_WORKERS: "{{ [ansible_facts.processor_vcpus * 2 + 1, 32] | min }}"
```
markgoddard, line 116: I realise this is just a sync, but this should be `max`, right?

priteau: The goal is to avoid running too many workers on hosts with many CPU cores, which is why we use `min`:

* on a host with 4 CPU core: `min(4 * 2 + 1, 32) = min(9, 32) = 9`
* on a host with 256 CPU cores: `min(256 * 2 + 1, 32) = min(513, 32) = 32`

priteau: The original uncapped formula would try running e.g. 513 workers on a 256-core host, which would fail due to resource exhaustion.

markgoddard: Of course. Brain not working after a 4am start :(

markgoddard (approved):

PR #591, opened 2023-08-18
Add Cloudkitty playbooks

technowhizz:


technowhizz:
@stackhpc/kayobe These are just standard cloudkitty playbooks. Would be nice to have in s-k-c

Alex-Welsh:
@technowhizz where are these coming from and what do they accomplish? Could you add a release note, and perhaps some documentation for what they are and how to use them?

technowhizz:
Came from @priteau or @markgoddard. I can add a release note.

markgoddard (changes requested):
I made an [attempt](https://github.com/stackhpc/ansible-collection-openstack/pull/30) at converting these playbooks into a role in the stackhpc.openstack collection. It's currently untested.

I think we might want to move the playbook that invokes the role to the openstack-config repo, since it fits there better than SKC.

markgoddard:
Thanks for proposing this. The playbooks have been converted to a role in the stackhpc.openstack collection, and support added to [openstack-config](https://github.com/stackhpc/openstack-config/pull/25).

PR #592, opened 2023-08-21
added some needed paths to lint ignore

g0rgamesh:


.yamllint
```diff
@@ -15,3 +15,8 @@ rules:
 ignore: |
   *etc/kayobe/environments/ci-multinode/secrets.yml
   *etc/kayobe/environments/ci-multinode/kolla/globals-tls-config.yml
+  *.automation/
+  *releasenotes/
+  *.github/
+  *.gitlab/
+  *.gitlab-ci.yml
```
markgoddard: Why do all of these need a `*` glob?

markgoddard (changes requested):

.yamllint
g0rgamesh: It works without them, removed.

markgoddard (approved):

PR #596, opened 2023-08-22
aufn/ceph: Update tenks.yml storage allocation

assumptionsandg:
Update storage allocation for Tenks.

Previous default allocations do not allow for an automatic deployment of an aufn and require manually increasing the storage capacity prior to deployment.


Alex-Welsh:
Has this been tested? It doesn't look like you're leaving yourself much room there for other things

MoteHue:
> Has this been tested? It doesn't look like you're leaving yourself much room there for other things

I've tested with an increase in the computes to 20GiB without issue. Haven't ever ran out of space on the storage nodes myself, have you had problems with the storage nodes too @assumptionsandg?

etc/kayobe/environments/aufn-ceph/tenks.yml
```diff
@@ -21,9 +21,9 @@ node_types:
     volumes:
       # There is a minimum disk space capacity requirement of 4GiB when using Ironic Python Agent:
       # https://github.com/openstack/ironic-python-agent/blob/master/ironic_python_agent/utils.py#L290
-      - capacity: 10GiB
+      - capacity: 20GiB
       # Ceph volume
-      - capacity: 10GiB
+      - capacity: 20GiB
```
Alex-Welsh, line 26: Why does the Ceph volume size need to be increased?

assumptionsandg:
Compute nodes are the main target. I seem to remember having issues with storage nodes in the past.

etc/kayobe/environments/aufn-ceph/tenks.yml
```diff
@@ -34,7 +34,7 @@ node_types:
     volumes:
       # There is a minimum disk space capacity requirement of 4GiB when using Ironic Python Agent:
       # https://github.com/openstack/ironic-python-agent/blob/master/ironic_python_agent/utils.py#L290
-      - capacity: 10GiB
+      - capacity: 20GiB
```
markgoddard: Combined storage requirements of all VMs is 220GB. SMS lab flavor says bare metal disks are 200GB, Matt checked one and found 233GB.

I'm not sure if these volumes are sparse, but if not they would consume the entire bare metal disk.

assumptionsandg:
I've pushed up a config known to work on CentOS Stream 8. The combined storage usage is 195GB in this configuration, This could potentially be improved further by slightly reducing the storage/compute rootdisk further.

Alex-Welsh:
> I've pushed up a config known to work on CentOS Stream 8. The combined storage usage is 195GB in this configuration, This could potentially be improved further by slightly reducing the storage/compute rootdisk further.

How far did you test it? I'd like to make sure it doesn't break quickly if you create a couple of glance images or logs stack up.

markgoddard (commented):
Would be nice to move this forward, one way or another.

Alex-Welsh:
I've cherry picked these commits along with `libvirt_vm_trust_guest_rx_filters: false` and the config works on Rocky 9. It might be worth adding the rx_filters variable and getting this pushed through

Alex-Welsh (dismissed):

markgoddard (approved):

PR #603, opened 2023-08-25
Magnum lb fix and focal repos bump

darmach:
Focal repos update - fix for unmet dependencies during container build.
Bump Magnum version - include fix for missing Octavi aload balancer health monitors.


mnasiadka:
needs reno

bbezak (approved):

PR #608, opened 2023-09-01
Add docs for Ceph RGWs with Cephadm

MoteHue:


MoteHue:
There may be more detail needed for the ingress services. I haven't deployed these myself, so have just taken the info given in the cephadm ansible collection readme.

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+
+Ceph RGWs require additional configuration for the following features:
+
+  * Support both S3 and Swift APIs.
+
+  * Authenticate user access via Keystone.
+
+  * Allow cross-project and public object access.
+
+The set of commands below configure all of these.
```
Alex-Welsh:
```suggestion
Ceph RGWs require additional configuration to:

  * Support both S3 and Swift APIs.

  * Authenticate user access via Keystone.

  * Allow cross-project and public object access.

The set of commands below configure all of these.
```

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+
+.. code:: yaml
+
+  # Append the following to cephadm_commands_post:
+  - "config set client.rgw rgw_content_length_compat true"
```
Alex-Welsh, line 340: Is there not a better way to set this config e.g. in a file somewhere?

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+
+  kayobe overcloud service deploy -kt ceph-rgw,keystone,haproxy,loadbalancer
+
+
+.. _RGWs-with-hyper-converged-Ceph:
+
+RGWs with hyper-converged Ceph
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you are using a hyper-converged Ceph setup (i.e. your OpenStack controllers
```
Alex-Welsh, line 387:
```suggestion
If you are using a hyperconverged Ceph setup (i.e. your OpenStack controllers
```

MoteHue: From googling, I've seen the term used both with and without hyphen. Personally, I think with is nicer to read :wink:

doc/source/configuration/cephadm.rst
MoteHue: Pretty sure that all cephadm configuration is done via commands like this. For example, the osd registration in the cephadm ansible collection: https://github.com/stackhpc/ansible-collection-cephadm/blob/380ad04ea7257c193784c61ce5c9b117ef3bee94/roles/cephadm/tasks/osds.yml#L12

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+
+RGWs are defined with the following:
+
+.. code:: yaml
+
+  cephadm_radosgw_services:
+ - id: myrgw
+   count_per_host: 1
+   spec:
+     rgw_frontend_port: 80
```
markgoddard: Indentation is off

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+  - "config set client.rgw rgw_enforce_swift_acls true"
+  - "config set client.rgw rgw_keystone_accepted_admin_roles 'admin'"
+  - "config set client.rgw rgw_keystone_accepted_roles 'member, Member, _member_, admin'"
+  - "config set client.rgw rgw_keystone_admin_domain Default"
+  - "config set client.rgw rgw_keystone_admin_password {{ secrets_ceph_rgw_keystone_password }}"
+  - "config set client.rgw rgw_keystone_admin_project service"
+  - "config set client.rgw rgw_keystone_admin_user 'ceph_rgw'"
+  - "config set client.rgw rgw_keystone_api_version '3'"
+  - "config set client.rgw rgw_keystone_token_cache_size '10000'"
+  - "config set client.rgw rgw_keystone_url https://{{ internal_net_vip_address }}:5000"
```
markgoddard: `internal_net_vip_address` is specific to your deployment. Better to use `kolla_internal_vip_address`, or even `kolla_internal_fqdn` which is more likely to work with internal TLS (the VIP may not be allowed on the cert).

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+  - "config set client.rgw rgw_keystone_verify_ssl false"
```
markgoddard, line 352: :(

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+pair).
+
+.. code:: yaml
+
+  cephadm_ingress_services:
+      - id: rgw.myrgw
+        spec:
+          frontend_port: 443
+          monitor_port: 1967
+          virtual_ip: 10.66.0.1/24
```
markgoddard: Should this be a different VIP than the Kolla one? Which subnet should it be on?

markgoddard (changes requested):

doc/source/configuration/cephadm.rst
MoteHue: I haven't actually deployed any ingress services so I'm not sure, I just took this example from the collection readme. https://github.com/stackhpc/ansible-collection-cephadm/blob/master/roles/cephadm/README.md

@cityofships git blame blames you ;), could you weigh in on this please? I suspect this is indeed the kolla public VIP?

doc/source/configuration/cephadm.rst
MoteHue: :( verify_ssl false?

doc/source/configuration/cephadm.rst
MoteHue: fixed

doc/source/configuration/cephadm.rst
MoteHue: Updated to use the fqdn

doc/source/configuration/cephadm.rst
markgoddard: Yes. Unfortunately it's probably fairly tricky to get the internal root CA cert into the Ceph containers and trusted.

doc/source/configuration/cephadm.rst
markgoddard: ping @cityofships

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+  - "config set client.rgw rgw_enforce_swift_acls true"
+  - "config set client.rgw rgw_keystone_accepted_admin_roles 'admin'"
+  - "config set client.rgw rgw_keystone_accepted_roles 'member, Member, _member_, admin'"
+  - "config set client.rgw rgw_keystone_admin_domain Default"
+  - "config set client.rgw rgw_keystone_admin_password {{ secrets_ceph_rgw_keystone_password }}"
+  - "config set client.rgw rgw_keystone_admin_project service"
+  - "config set client.rgw rgw_keystone_admin_user 'ceph_rgw'"
+  - "config set client.rgw rgw_keystone_api_version '3'"
+  - "config set client.rgw rgw_keystone_token_cache_size '10000'"
+  - "config set client.rgw rgw_keystone_url {{ kolla_internal_fqdn }}:5000"
```
markgoddard: Is this missing the `http(s)://` scheme?

MoteHue: Yes, have added it back

doc/source/configuration/cephadm.rst
cityofships: Sorry both, I missed your pings in the notifications flood somehow;

@markgoddard Definitely different from Kolla's VIP - unless you'd like keepaliveds fighting for that VIP. It's either Kolla making for an object store front end (HAproxy + keepalived) or Ceph's Ingress - e.g. for standalone clusters. Question on the VIP subnet - that's entirely up for the network design. It can be arbitrary really, outside of any Ceph's networks.

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,128 @@ should be used in the Kolla Manila configuration e.g.:
...
+
+RGWs are defined with the following:
+
+.. code:: yaml
+
+  cephadm_radosgw_services:
+    - id: myrgw
+      count_per_host: 1
+      spec:
+        rgw_frontend_port: 80
```
cityofships: `80` is rather unusual choice - does not correspond with an example from hyper-converged section too. Normally, web services like RGW have ports >1024 allocated.

MoteHue: I've changed to suggest port 8100 by default, as afaik this doesn't clash with any other services we deploy.

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,132 @@ should be used in the Kolla Manila configuration e.g.:
 
   manila_cephfs_filesystem_name: manila-cephfs
 
+RADOS Gateways
+--------------
+
+RGWs are defined with the following:
```
Alex-Welsh:
```suggestion
RADOS Gateways (RGWs) are defined with the following:
```

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,132 @@ should be used in the Kolla Manila configuration e.g.:
...
+
+Ceph RGWs require additional configuration to:
+
+  * Support both S3 and Swift APIs.
+
+  * Authenticate user access via Keystone.
+
+  * Allow cross-project and public object access.
+
+The set of commands below configure all of these.
```
Alex-Welsh, line 335: It would be nice to split these out into different sections so readers know which ones do which tasks. e.g. If I want authentication with keystone but not public object access, which ones do I use?

MoteHue: I could add the following comments, but IMO this just confuses things further. The keystone auth is only needed if S3+Swift is already enabled, and part of the object access requires keystone config too. 

It is my understanding that we would need all these options for RGWs to be used effectively in our deployments, so it would be better not to suggest only enabling some.

```
    # S3 and Swift Support
  - "config set client.rgw rgw_content_length_compat true"
  - "config set client.rgw rgw_enable_apis 's3, swift, swift_auth, admin'"
  - "config set client.rgw rgw_enforce_swift_acls true"
    # Keystone integration
  - "config set client.rgw rgw_keystone_accepted_admin_roles 'admin'"
  - "config set client.rgw rgw_keystone_accepted_roles 'member, Member, _member_, admin'"
  - "config set client.rgw rgw_keystone_admin_domain Default"
  - "config set client.rgw rgw_keystone_admin_password {{ secrets_ceph_rgw_keystone_password }}"
  - "config set client.rgw rgw_keystone_admin_project service"
  - "config set client.rgw rgw_keystone_admin_user 'ceph_rgw'"
  - "config set client.rgw rgw_keystone_api_version '3'"
  - "config set client.rgw rgw_keystone_token_cache_size '10000'"
  - "config set client.rgw rgw_keystone_url https://{{ kolla_internal_fqdn }}:5000"
  - "config set client.rgw rgw_keystone_verify_ssl false"
  - "config set client.rgw rgw_max_attr_name_len '1000'"
  - "config set client.rgw rgw_max_attr_size '1000'"
  - "config set client.rgw rgw_max_attrs_num_in_req '1000'"
    # Authenticate user access with Keystone
  - "config set client.rgw rgw_s3_auth_use_keystone true"
    # Allow cross-project and public object access
  - "config set client.rgw rgw_swift_account_in_url true"
  - "config set client.rgw rgw_swift_versioning_enabled true"

File: doc/source/configuration/cephadm.rst (hunk 7 above)
Comment (MoteHue) [reply doc/source/configuration/cephadm.rst]: I've written a bit to clarify this in the docs

File: doc/source/configuration/cephadm.rst [hunk 12]
```diff
@@ -308,6 +308,136 @@ should be used in the Kolla Manila configuration e.g.:
...
+HA with Ingress services
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ingress services are defined with the following. ``id`` should match the name
+(not id) of the RGW service to which ingress will point to. ``spec`` is a
+service specification required by Cephadm to deploy the ingress (haproxy +
+keepalived pair).
+
+Not that the ``virtual_ip`` here should be different than the Kolla VIP. The
+choice of subnet will be dependent on your deployment, but should be outside
```
cityofships: > should be outside of any other Ceph networks.

That's possibility, not a requirement

doc/source/configuration/cephadm.rst
```diff
@@ -308,6 +308,136 @@ should be used in the Kolla Manila configuration e.g.:
...
+
+HA with Ingress services
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ingress services are defined with the following. ``id`` should match the name
+(not id) of the RGW service to which ingress will point to. ``spec`` is a
+service specification required by Cephadm to deploy the ingress (haproxy +
+keepalived pair).
+
+Not that the ``virtual_ip`` here should be different than the Kolla VIP. The
```
cityofships: nit: Note

I would also emphasise that the VIP **must** be different to avoid conflict, not just **should**.

cityofships:
Just these two comments, otherwise LGTM 👍

doc/source/configuration/cephadm.rst
MoteHue: Ah ok, updated to reflect this

doc/source/configuration/cephadm.rst
MoteHue: Done

markgoddard (approved):

PR #610, opened 2023-09-01
Speed up OSD backfill/recovery in Quincy

priteau:
See https://www.spinics.net/lists/ceph-users/msg77216.html for details.

### DIFF
--- a/etc/kayobe/cephadm.yml
+++ b/etc/kayobe/cephadm.yml
@@ -87,7 +87,8 @@ cephadm_cluster_network: "{{ storage_mgmt_net_name | net_cidr }}"
 # post-deployment configuration, post commands run after the rest of the
 # post-deployment configuration.
 #cephadm_commands_pre:
-#cephadm_commands_post:
+cephadm_commands_post:
+  - "{% if cephadm_ceph_release == 'quincy' %}config set osd osd_mclock_cost_per_byte_usec_hdd 0.1{% endif %}"
 
 ###############################################################################
 # Kolla Ceph auto-configuration.
--- a/releasenotes/notes/ceph-quincy-recovery-speedup-93a8bf3310c6dafc.yaml
+++ b/releasenotes/notes/ceph-quincy-recovery-speedup-93a8bf3310c6dafc.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes slow backfill/recovery speed when using the Ceph Quincy release with
+    HDD storage.


etc/kayobe/cephadm.yml
```diff
@@ -87,7 +87,8 @@ cephadm_cluster_network: "{{ storage_mgmt_net_name | net_cidr }}"
 # post-deployment configuration, post commands run after the rest of the
 # post-deployment configuration.
 #cephadm_commands_pre:
-#cephadm_commands_post:
+cephadm_commands_post:
+  - "{% if cephadm_ceph_release == 'quincy' %}config set osd osd_mclock_cost_per_byte_usec_hdd 0.1{% endif %}"
```
mnasiadka, line 91: I would be cautious with setting this everywhere, this might impact normal client i/o - @cityofships do you remember what was a ,,safe'' value?
I would prefer us having a list of known Ceph issues on the wiki and workaround proposals instead of using static values for things like that assuming it will work properly on any cluster sizes and configurations.

markgoddard (dismissed):

markgoddard (commented):
I suppose a reno might be useful.

priteau:
> I suppose a reno might be useful.

Sorry, I always forget them in this repository…

etc/kayobe/cephadm.yml
priteau: > I would be cautious with setting this everywhere, this might impact normal client i/o - @cityofships do you remember what was a ,,safe'' value?
> I would prefer us having a list of known Ceph issues on the wiki and workaround proposals instead of using static values for things like that assuming it will work properly on any cluster sizes and configurations.

What about putting it behind a feature flag?

mnasiadka: A lot of code for handling a bug, which is already fixed upstream (and then we need to unset it on all environments after 17.2.7 gets released) - should we rather document the workaround on wiki?

Or maybe we should start thinking about a process for building Ceph container images on these occasions where we need slightly newer code that is not in the last release? (it's getting a bit annoying Ceph upstream release team does not care for such critical bugs and doesn't do ,,fast releases'' for those bugs).

I'm fine with the feature flag, if we prefer to go this way.

cityofships: > I would be cautious with setting this everywhere, this might impact normal client i/o - @cityofships do you remember what was a ,,safe'' value? I would prefer us having a list of known Ceph issues on the wiki and workaround proposals instead of using static values for things like that assuming it will work properly on any cluster sizes and configurations.

It was `0.4`. We've seen various client-side timeouts during backfilling after setting it as low as `0.1`.

cityofships: I'd say too it could be per-cluster setting rather than defaults; also not every minor Quincy version is affected.

markgoddard:
Seems like there's some opposition to this. Perhaps it should move to documentation?

priteau:
> Seems like there's some opposition to this. Perhaps it should move to documentation?

A new Quincy release should be available soon anyway…

PR #611, opened 2023-09-01
Fix for ci-aio automated deploy on non-lvm images

assumptionsandg:
Quick fix to prevent grow-root from crashing out on non-lvm based images.

### DIFF
--- a/etc/kayobe/ansible/growroot.yml
+++ b/etc/kayobe/ansible/growroot.yml
@@ -46,7 +46,7 @@
             name: "{% if os_family == 'RedHat' %}cloud-utils-growpart{% else %}cloud-guest-utils{% endif %}"
             state: present
             cache_valid_time: "{{ apt_cache_valid_time if os_family == 'Debian' else omit }}"
-            update_cache: "{{ True if os_family == 'Debian' else omit }}"
+            update_cache: "{{ true if os_family == 'Debian' else omit }}"
           become: true
 
         - name: Get root PV device
--- a/etc/kayobe/ansible/growroot.yml
+++ b/etc/kayobe/ansible/growroot.yml
@@ -27,57 +27,70 @@
     growroot_vg: "rootvg"
     # Don't assume facts are present.
     os_family: "{{ ansible_facts.os_family | default('Debian' if os_distribution == 'ubuntu' else 'RedHat') }}"
+    # LVM override
+    skip_lvm_check: false
 
   tasks:
-    - name: Check if growpart is installed
+    - name: Check LVM status
       shell:
-        cmd: type growpart
+        cmd: vgdisplay | grep -q lvm2
       changed_when: false
       failed_when: false
       check_mode: false
-      register: growpart_check
+      register: lvm_check
       become: true
 
-    - name: Ensure growpart is installed
-      package:
-        name: "{% if os_family == 'RedHat' %}cloud-utils-growpart{% else %}cloud-guest-utils{% endif %}"
-        state: present
-        cache_valid_time: "{{ apt_cache_valid_time if os_family == 'Debian' else omit }}"
-        update_cache: "{{ True if os_family == 'Debian' else omit }}"
-      become: True
-      when: growpart_check.rc != 0
+    - block:
+      - name: Check if growpart is installed
+        shell:
+          cmd: type growpart
+        changed_when: false
+        failed_when: false
+        check_mode: false
+        register: growpart_check
+        become: true
 
-    - name: Get root PV device
-      command: "pvs --select vg_name={{ growroot_vg }} --reportformat json"
-      register: pvs
-      become: True
-      changed_when: False
-      check_mode: false
+      - name: Ensure growpart is installed
+        package:
+          name: "{% if os_family == 'RedHat' %}cloud-utils-growpart{% else %}cloud-guest-utils{% endif %}"
+          state: present
+          cache_valid_time: "{{ apt_cache_valid_time if os_family == 'Debian' else omit }}"
+          update_cache: "{{ True if os_family == 'Debian' else omit }}"
+        become: True
+        when: growpart_check.rc != 0
 
-    - name: Fail if root PV device not found
-      fail:
-        msg: >
-          Expected LVM physical volume devices not found in volume group {{ growroot_vg }}
-      when: (pvs.stdout | from_json).report[0].pv | length == 0
+      - name: Get root PV device
+        command: "pvs --select vg_name={{ growroot_vg }} --reportformat json"
+        register: pvs
+        become: True
+        changed_when: False
+        check_mode: false
 
-    - name: Grow partition
-      command: "growpart {{ disk }} {{ part_num }}"
-      vars:
-        pv: "{{ pvs.stdout | from_json }}"
-        disk_tmp: "{{ pv.report[0].pv[0].pv_name[:-1] }}"
-        disk: "{{ disk_tmp[:-1] if disk_tmp[-1] == 'p' else disk_tmp }}"
-        part_num: "{{ pv.report[0].pv[0].pv_name[-1] }}"
-      become: True
-      failed_when: "growpart.rc != 0 and 'NOCHANGE' not in growpart.stdout"
-      changed_when: "'NOCHANGE' not in growpart.stdout"
-      register: growpart
+      - name: Fail if root PV device not found
+        fail:
+          msg: >
+            Expected LVM physical volume devices not found in volume group {{ growroot_vg }}
+        when: (pvs.stdout | from_json).report[0].pv | length == 0
 
-    - name: Grow LVM PV
-      command: "pvresize {{ disk }}"
-      vars:
-        pv: "{{ pvs.stdout | from_json }}"
-        disk: "{{ pv.report[0].pv[0].pv_name }}"
-      become: True
+      - name: Grow partition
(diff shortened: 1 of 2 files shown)


Alex-Welsh (dismissed):

Alex-Welsh:
Alternative option would be to update the growroot playbook to skip if no LVM is detected

mnasiadka:
+1 for that

markgoddard:
+1

etc/kayobe/ansible/growroot.yml
```diff
@@ -27,57 +27,70 @@
...
-      when: growpart_check.rc != 0
+    - block:
+      - name: Check if growpart is installed
+        shell:
+          cmd: type growpart
+        changed_when: false
+        failed_when: false
+        check_mode: false
+        register: growpart_check
+        become: true
```
Alex-Welsh: Is this necessary?

etc/kayobe/ansible/growroot.yml
```diff
@@ -27,57 +27,70 @@
...
-        pv: "{{ pvs.stdout | from_json }}"
-        disk: "{{ pv.report[0].pv[0].pv_name }}"
-      become: True
+      - name: Grow LVM PV
+        command: "pvresize {{ disk }}"
+        vars:
+          pv: "{{ pvs.stdout | from_json }}"
+          disk: "{{ pv.report[0].pv[0].pv_name }}"
+        become: True
+      when: lvm_check.rc != 1 or {{ skip_lvm_check }}
```
Alex-Welsh:
```suggestion
      when: lvm_check.rc != 1 or skip_lvm_check
```

etc/kayobe/ansible/growroot.yml
```diff
@@ -27,57 +27,70 @@
     growroot_vg: "rootvg"
     # Don't assume facts are present.
     os_family: "{{ ansible_facts.os_family | default('Debian' if os_distribution == 'ubuntu' else 'RedHat') }}"
+    # LVM override
+    skip_lvm_check: false
```
Alex-Welsh: Might be worth prefixing this with something like `growroot_` to make it a bit more specific

Alex-Welsh (changes requested):

etc/kayobe/ansible/growroot.yml
holliefae: I don't believe so, it seems the install growpart task passes even when cloud-guest-utils is already installed so there seems no need to skip it. Maybe there was another reason why that task was added?

etc/kayobe/ansible/growroot.yml
holliefae: My mistake!

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -67,7 +67,9 @@ source kayobe-env --environment ci-aio
 
 kayobe control host bootstrap
 
-kayobe playbook run etc/kayobe/ansible/growroot.yml
+if [ sudo vgdisplay | grep -q lvm2 ]; then
```
Alex-Welsh: No longer necessary

etc/kayobe/ansible/growroot.yml
holliefae: Done :)

Alex-Welsh (changes requested):
If you're changing `true` to `True`, you've missed a few instances.

etc/kayobe/environments/ci-aio/automated-setup.sh
holliefae: Removed.

etc/kayobe/ansible/growroot.yml
```diff
@@ -84,4 +84,4 @@
 #      when: "'NOCHANGE' not in growpart.stdout"
 # Commenting out the conditional because growpart is already triggered by cloud-init - hence it emits NOCHANGE
 # Cloud-Inits growpart implementation has a bug https://bugzilla.redhat.com/show_bug.cgi?id=2122575
-# PVresize is not being triggered
+# PVresize is not being triggered
```
Alex-Welsh: nit: you're now missing a newline at the end of this file

Alex-Welsh (dismissed):
LGTM but I think someone like @jackhodgkiss or @markgoddard need to verify the removal of `Check if growpart is installed`

Alex-Welsh:
Remember to squash these commits when you merge :slightly_smiling_face:

Alex-Welsh:
And ignore the tox errors for now. GH actions are universally having problems

etc/kayobe/ansible/growroot.yml
```diff
@@ -27,57 +27,60 @@
...
     os_family: "{{ ansible_facts.os_family | default('Debian' if os_distribution == 'ubuntu' else 'RedHat') }}"
+    # LVM override
+    growroot_skip_lvm_check: false
 
   tasks:
-    - name: Check if growpart is installed
+    - name: Check LVM status
       shell:
-        cmd: type growpart
+        cmd: vgdisplay | grep -q lvm2
```
markgoddard, line 36: FWIW, there may be cases where this is not installed, but I think this will do the right thing anyway.

etc/kayobe/ansible/growroot.yml
```diff
@@ -27,57 +27,60 @@
...
-        pv: "{{ pvs.stdout | from_json }}"
-        disk: "{{ pv.report[0].pv[0].pv_name }}"
-      become: True
+        - name: Grow LVM PV
+          command: "pvresize {{ disk }}"
+          vars:
+            pv: "{{ pvs.stdout | from_json }}"
+            disk: "{{ pv.report[0].pv[0].pv_name }}"
+          become: true
+      when: lvm_check.rc != 1 or growroot_skip_lvm_check
```
markgoddard: Why 1? Would it be safer to check `lvm_check.rc == 0`?

Also, should it be `and not growroot_skip_lvm_check | bool`?

Alex-Welsh: > Also, should it be `and not growroot_skip_lvm_check | bool`?

The variable was here for an option to always try to apply the config (and therefore error out on non-lvm images) preserving the original behaviour

markgoddard: Right, but based on the name we want to skip when it's true, or not skip when it's false

Alex-Welsh: Yep, you're right. I should really stop contradicting you at 5PM

holliefae: Added those fixes now

etc/kayobe/ansible/growroot.yml
```diff
@@ -27,57 +27,60 @@
...
-        pv: "{{ pvs.stdout | from_json }}"
-        disk: "{{ pv.report[0].pv[0].pv_name }}"
-      become: True
+        - name: Grow LVM PV
+          command: "pvresize {{ disk }}"
+          vars:
+            pv: "{{ pvs.stdout | from_json }}"
+            disk: "{{ pv.report[0].pv[0].pv_name }}"
+          become: true
+      when: lvm_check.rc == 0 or not growroot_skip_lvm_check | bool
```
markgoddard:
```suggestion
      when: lvm_check.rc == 0 and not growroot_skip_lvm_check | bool
```

markgoddard (changes requested):

etc/kayobe/ansible/growroot.yml
holliefae: Won't changing that condition prevent the playbook from running when lvm_check fails? The override variable allows the playbook to run even when lvm_check doesn't return 0.

markgoddard: Ah ok, I read that wrong. It should probably be called growroot_ignore_lvm_check, or actually skip the LVM check.

markgoddard (approved):

etc/kayobe/ansible/growroot.yml
markgoddard: Actually the logic is wrong. It should be 

```
when: lvm_check.rc == 0 or growroot_skip_lvm_check | bool
```

markgoddard (changes requested):
Logic incorrect

markgoddard (approved):

etc/kayobe/ansible/growroot.yml
```diff
@@ -27,57 +27,60 @@
...
     # Don't assume facts are present.
     os_family: "{{ ansible_facts.os_family | default('Debian' if os_distribution == 'ubuntu' else 'RedHat') }}"
+    # Ignore LVM check
+    growroot_ignore_lvm_check: false
 
   tasks:
-    - name: Check if growpart is installed
+    - name: Check LVM status
       shell:
-        cmd: type growpart
```
markgoddard, line 34: Hmm, looking at this again we've lost the growpart check. I think that was trying to avoid package installation during early system bootstrapping where possible, since we might not have network and proxy configuration in place.

markgoddard (changes requested):

etc/kayobe/ansible/growroot.yml
holliefae: Ah that makes sense. I removed it initially as i thought it's purpose was fulfilled by the install task.

markgoddard (approved):

PR #613, opened 2023-09-04
zed: upstream sync

markgoddard:
- Synchronise with kayobe stable/zed


Alex-Welsh:
should this be into `stackhpc/zed`?

PR #614, opened 2023-09-04
Reduce service workers for aufn-ceph & virtualised ci-multinode envs

markgoddard:
When running a multinode environment in VMs, we can easily run out of
memory on controllers. Reduce to 2 service/RPC workers in this case.
Also reduce RPC workers to 1 for the aufn-ceph environment.


Alex-Welsh (approved):
LGTM

PR #615, opened 2023-09-04
multinode: Run RefStack Tempest tests by default

markgoddard:
A full tempest test run takes a long time, and may not always be
necessary. Let's run RefStack by default, and make it easy to switch to
the full test suite.


priteau (approved):

priteau:
Related change: https://github.com/stackhpc/stackhpc-kayobe-config/pull/617

PR #629, opened 2023-09-08
Fix zed host image build workflow

Alex-Welsh:


Alex-Welsh:
kayobe-automation seems to need this patch to work for zed.
https://github.com/stackhpc/kayobe-automation/compare/main...fix-format
Not quite sure when the change was made but I'll make a proper fix and PR soon

markgoddard:
How much of this is still required?

Alex-Welsh:
I'm not sure. The host image build workflow changes are probably still required. I'll test it soon

.github/workflows/overcloud-host-image-build.yml
markgoddard, line 1: Should these fixes go back to yoga?

Alex-Welsh: It's running fine without them, probably because it uses the old CS8 runners

.github/workflows/overcloud-host-image-build.yml
```diff
@@ -34,14 +34,14 @@ jobs:
...
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           path: src/kayobe-config
 
       - name: Determine OpenStack release
         id: openstack_release
         run: |
-          BRANCH=$(awk -F'=' '/defaultbranch/ {print $2}' .gitreview)
+          BRANCH=$(awk -F'=' '/defaultbranch/ {print $2}' src/kayobe-config/.gitreview)
```
markgoddard, line 44: Isn't this a bug fix that would apply to yoga too?

markgoddard (approved):

.github/workflows/overcloud-host-image-build.yml
markgoddard: Merged, but please backport this fix if necessary.

PR #631, opened 2023-09-11
Add missing linux-firmware package in Rocky Linux 9

cityofships:


bbezak (approved):
approved - please try to fix it in DIB upstream as well

cityofships:
Does it have to wait for CI? Looks like there's no influence

cityofships:
> approved - please try to fix it in DIB upstream as well

https://review.opendev.org/c/openstack/diskimage-builder/+/894905

Root cause is the package list was copied from fedora-container element

PR #633, opened 2023-09-12
zed: remove symlinks

markgoddard:
Remove plugins and group_vars symlinks. These are no longer required in Zed.


Alex-Welsh:
Wrong target?

Alex-Welsh (approved):
LGTM

PR #637, opened 2023-09-13
Hostnames in prometheus

Alex-Welsh:
use prometheus_instance_label to have nice hostnames rather than IPs

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -103,5 +103,8 @@ prometheus_ceph_mgr_exporter_endpoints:
 {% endfor %}
 {% endif %}
 
+# Use inventory hostnames as labels
+prometheus_instance_label: "{% raw %}{{ ansible_facts.hostname }}{% endraw %}"
+
 #############################################################################
 
--- a/releasenotes/notes/prometheus-labels-1aebdfd8631a3406.yaml
+++ b/releasenotes/notes/prometheus-labels-1aebdfd8631a3406.yaml
@@ -1,0 +1,4 @@
+---
+upgrade:
+  - |
+    Instance labels in prometheus now use inventory hostnames rather than IPs.


technowhizz (dismissed):
LGTM. Ideally you should remove the .automation in a separate commit but thats just a nitpick

Alex-Welsh:
> LGTM. Ideally you should remove the .automation in a separate commit but thats just a nitpick

Ah, that's just me being bad at submodules

Alex-Welsh:
> > LGTM. Ideally you should remove the .automation in a separate commit but thats just a nitpick
> 
> Ah, that's just me being bad at submodules

Fixed

technowhizz (dismissed):

etc/kayobe/kolla/globals.yml
```diff
@@ -103,5 +103,8 @@ prometheus_ceph_mgr_exporter_endpoints:
 {% endfor %}
 {% endif %}
 
+# Use inventory hostnames as labels
+prometheus_instance_label: "{{ ansible_facts.hostname }}"
```
markgoddard: Needs to be wrapped in `{% raw %}`

markgoddard: Also this should be targeted at Zed because it's a breaking change.

Alex-Welsh: > Needs to be wrapped in {% raw %}

Done

Alex-Welsh: I thought this was a stackhpc/yoga change, from the description on the ticket: https://stackhpc.atlassian.net/browse/INFRA-143

Alex-Welsh: or it's available in yoga but we don't want it to be the default?

Alex-Welsh:
closing in favour of https://github.com/stackhpc/stackhpc-kayobe-config/pull/640

etc/kayobe/kolla/globals.yml
markgoddard: It's a breaking change, so we don't want to apply it by default on yoga. Individual deployments can still apply it on yoga if they wish.

PR #639, opened 2023-09-13
CI: Only sync old images after build

markgoddard:
The new format (without distro & type in the name) is only used from Zed.


Alex-Welsh (approved):
LGTM

PR #642, opened 2023-09-14
yoga: xena merge

markgoddard:
- Restore upstream mirrors in container image package repos
- CI: Only sync old images after build


Alex-Welsh (approved):
LGTM

PR #644, opened 2023-09-14
Update branch list for readthedocs configuration

priteau:
This will need to be merged into all newer branches.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,7 +13,7 @@ build:
     python: "3.7"
   jobs:
     post_checkout:
-      - git remote set-branches origin master stackhpc/yoga stackhpc/xena stackhpc/wallaby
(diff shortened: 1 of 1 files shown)


.readthedocs.yaml
```diff
@@ -13,7 +13,7 @@ build:
     python: "3.7"
   jobs:
     post_checkout:
-      - git remote set-branches origin master stackhpc/yoga stackhpc/xena stackhpc/wallaby
+      - git remote set-branches origin master stackhpc/2023.1 stackhpc/zed stackhpc/yoga stackhpc/xena stackhpc/wallaby
```
markgoddard, line 16: Do the older branches need to know about later ones?

priteau: I don't think so. Do you prefer the fix to be applied to each branch separately?

markgoddard: I don't mind too much, but AFAICT each branch lists the earlier branches, so I was wondering if there is actually an issue to be fixed.

priteau: I thought they didn't, I must have misread.

PR #646, opened 2023-09-15
Pin pulp version to 3.24.0

jackhodgkiss:
The 3.24 tag was recently updated to 3.24.1 which includes a version of pulp_container that causes client pulp container syncs to fail.


Alex-Welsh (dismissed):
LGTM

Alex-Welsh (approved):

PR #648, opened 2023-09-18
Set q35 hw machine type as default

darmach:
Changed nova.conf machine type to q35.

### DIFF
--- a/etc/kayobe/kolla/config/nova.conf
+++ b/etc/kayobe/kolla/config/nova.conf
@@ -1,0 +1,3 @@
+[libvirt]
+hw_machine_type = q35
+num_pcie_ports = 16
--- a/etc/kayobe/kolla/config/nova.conf
+++ b/etc/kayobe/kolla/config/nova.conf
@@ -1,3 +1,2 @@
 [libvirt]
 hw_machine_type = q35
-num_pcie_ports = 16
--- a/releasenotes/notes/nova-machine-type-q35-ab363c5f7f3f9807.yaml
+++ b/releasenotes/notes/nova-machine-type-q35-ab363c5f7f3f9807.yaml
@@ -1,0 +1,13 @@
+---
+upgrade:
+  - |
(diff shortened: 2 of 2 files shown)


etc/kayobe/kolla/config/nova.conf
```diff
@@ -0,0 +1,3 @@
+[libvirt]
+hw_machine_type = q35
+num_pcie_ports = 16
```
mnasiadka: That's going to be backported upstream (right @bbezak ?) - so we probably could set only hw_machine_type?

bbezak: master got merged today. backports just started - https://review.opendev.org/c/openstack/kolla-ansible/+/895373

bbezak: stable/yoga merged

bbezak (approved):

releasenotes/notes/nova-machine-type-q35-ab363c5f7f3f9807.yaml
```diff
@@ -0,0 +1,5 @@
+---
+upgrade:
+  - |
+    Configure Nova to use more modern 'q35' libvirt machine type rather than
+    'pc' which is considered legacy.
```
markgoddard, line 5: Will there be any impact on existing systems from changing this?

etc/kayobe/kolla/config/nova.conf
markgoddard: Probably we should have kept this since we can't assume kolla-ansible is up to date on client deployments.

releasenotes/notes/nova-machine-type-q35-ab363c5f7f3f9807.yaml
darmach: Most likely no - in yoga all instances should have type metadata already, during upgrade check will fail if not. When restarted new vms should use new type afaik.

PR #649, opened 2023-09-19
fix build of rocky and ubuntu images

bbezak:
recent commit broke rocky/ubuntu builds:
https://github.com/stackhpc/stackhpc-kayobe-config/commit/908d7513df8eed0cfb0aadced4cc9a34c78c7027#diff-12d955d16a7bd8d6dc7bae8b56c2e485f4445434f0fb67546ea09c9e8b9b504dR332
https://github.com/stackhpc/stackhpc-kayobe-config/blame/stackhpc/yoga/etc/kayobe/kolla.yml#L325C8-L325C46

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -322,7 +322,7 @@ kolla_build_blocks:
     {% endif %}
   base_ubuntu_package_sources_list: |
     RUN \
-        mv /etc/apt/sources.list{,.backup} && \
+        mv /etc/apt/sources.list /etc/apt/sources.list.backup && \
     {% for repo in stackhpc_ubuntu_focal_repos %}
         echo '{{ repo }}' >> /etc/apt/sources.list {% if not loop.last %} && \
     {% endif %}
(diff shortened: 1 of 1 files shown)


etc/kayobe/kolla.yml
```diff
@@ -329,7 +329,7 @@ kolla_build_blocks:
     {% endfor %}
   footer: |
     {% if stackhpc_kolla_clean_up_repo_mirrors | bool %}
-    {% if kolla_base_distro == 'centos' %}
+    {% if kolla_base_distro in ['centos', 'rocky'] %}
```
Alex-Welsh, line 332: From Michal
> `if base_package_type == 'rpm'`  would be probably better

But it might be better to make that a separate change

Alex-Welsh (dismissed):

etc/kayobe/kolla.yml
mnasiadka: But that would need to go as raw to Kolla-build

bbezak:
ubuntu is also broken:
    INFO:kolla.common.utils.base:[91mmv: missing destination file operand after '/etc/apt/sources.list{,.backup}'

markgoddard (approved):

bbezak:
(still testing it in https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/6233874225)
maybe we should run img build workflow when editing kolla.yml ?

bbezak:
build went fine

priteau (approved):

PR #653, opened 2023-09-21
Small Antelope fixes

assumptionsandg:
- ci-aio: Update to stackhpc/2023.1
- aufn/ceph: Update to stackhpc/2023.1
- stackhpc.yml update to Antelope repository.

### DIFF
--- a/etc/kayobe/environments/aufn-ceph/a-universe-from-nothing.sh
+++ b/etc/kayobe/environments/aufn-ceph/a-universe-from-nothing.sh
@@ -10,8 +10,8 @@
 set -eu
 
 BASE_PATH=~
-KAYOBE_BRANCH=stackhpc/yoga
-KAYOBE_CONFIG_BRANCH=stackhpc/yoga
+KAYOBE_BRANCH=stackhpc/2023.1
+KAYOBE_CONFIG_BRANCH=stackhpc/2023.1
 KAYOBE_ENVIRONMENT=aufn-ceph
 
 PELICAN_HOST="10.0.0.34 pelican pelican.service.compute.sms-lab.cloud"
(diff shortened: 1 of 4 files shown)


etc/kayobe/environments/aufn-ceph/a-universe-from-nothing.sh
```diff
@@ -10,8 +10,8 @@
 set -eu
 
 BASE_PATH=~
-KAYOBE_BRANCH=stackhpc/yoga
-KAYOBE_CONFIG_BRANCH=stackhpc/yoga
+KAYOBE_BRANCH=stackhpc/2023.1
+KAYOBE_CONFIG_BRANCH=stackhpc/2023.1
 KAYOBE_ENVIRONMENT=aufn-ceph
 
 PELICAN_HOST="10.0.0.34 pelican pelican.service.compute.sms-lab.cloud"
```
Alex-Welsh, line 17: There are more changes to make in this file.
We're moving away from ``virtualenv`` so the ``python3-virtualenv`` package needs to go.
so does ``virtualenv kayobe``
We should check the images we use (we're moving to the ark-hosted images for aufn in the future) and make sure we're not missing any packages.
Ideally we should be testing these changes before making them. With the 'wrong' branch at the top, it's clear this script is outdated. If we haven't tested it, I think we should have a big clear warning at the top of the file

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -12,8 +12,8 @@ if [ sudo vgdisplay | grep -q lvm2 ]; then
 fi
 
 BASE_PATH=~
-KAYOBE_BRANCH=stackhpc/zed
-KAYOBE_CONFIG_BRANCH=stackhpc/zed
```
Alex-Welsh, line 16: Again, big clear warning at the top of the file would be nice until we get actually test this

MoteHue: If you're adding warnings, perhaps put them in the docs for these environments? As they do just instruct you to download and run these scripts without ever opening them :wink:

Alex-Welsh: There's warnings for zed, which will get merged up soonish

markgoddard (commented):
How do we move this one forward? Seems like there are useful things worth merging.

Alex-Welsh (approved):

PR #655, opened 2023-09-21
Fix seed container image build

jackhodgkiss:
The argument list was missing -e which meant the first argument was interpreted as image regexes.


priteau (approved):
Sorry for missing xena. Also needs to be added to wallaby.

Alex-Welsh:
Too slow :wink: https://github.com/stackhpc/stackhpc-kayobe-config/pull/658

PR #663, opened 2023-09-27
zed: automate Kolla Ansible tag variable definitions

markgoddard:
This change reworks the way that Kolla image tags are defined, making the following improvements possible:

* Automatic promotion of images on merge (as is currently done for package repos)
* New CI job that checks that all specified tags are present in Test Pulp. This shows that they have been built, pushed to Ark, and also synced to Test Pulp on SMS lab
* Automatic definition of the <container>_tag variables for Kolla Ansible

The main visible change is that container image tags are now defined in the kolla_image_tags variable in etc/kayobe/kolla-image-tags.yml. This avoids the previous churn in the etc/kayobe/kolla/globals.yml file.

A new script, tools/kolla-images.py, understands the Kolla image and container relationships, and is able to display the tags used by each image. This output is parsed in the new container promotion workflows added in https://github.com/stackhpc/stackhpc-release-train/pull/193.

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -10,6 +10,9 @@ enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names }}{% endraw %}"
 # This is necessary for os migrations where mixed clouds might be deployed
 kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"
 
+# Convenience variable for base distro and version string.
+kolla_base_distro_and_version: "{% raw %}{{ kolla_base_distro }}-{{ kolla_base_distro_version }}{% endraw %}"
+
 # Dict of Kolla image tags to deploy for each service.
 # Each key is the tag variable prefix name, and the value is another dict,
 # where the key is the OS distro and the value is the tag to deploy.
@@ -17,12 +20,11 @@ kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %
 kolla_image_tags:
 {{ kolla_image_tags | to_nice_yaml | indent(width=4, first=true) }}
 
-openstack_tag: "{% raw %}{{ kayobe_image_tags['openstack'][kolla_base_distro] }}{% endraw %}"
-bifrost_tag: "{% raw %}{{ kayobe_image_tags['bifrost'][kolla_base_distro] }}{% endraw %}"
-ovn_tag: "{% raw %}{{ kayobe_image_tags['ovn'][kolla_base_distro] }}{% endraw %}"
-cloudkitty_tag: "{% raw %}{{ kayobe_image_tags['cloudkitty'][kolla_base_distro] }}{% endraw %}"
-neutron_tag: "{% raw %}{{ kayobe_image_tags['neutron'][kolla_base_distro] }}{% endraw %}"
-opensearch_tag: "{% raw %}{{ kayobe_image_tags['opensearch'][kolla_base_distro] }}{% endraw %}"
+# Variables defining which tag to use for each container's image.
+{{ lookup('pipe', 'python3 ' ~ kayobe_config_path ~ '/../../tools/kolla-images.py list-tag-vars') }}
+
+#############################################################################
+# RabbitMQ
 
 om_enable_rabbitmq_high_availability: true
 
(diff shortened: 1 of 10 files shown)


etc/kayobe/kolla/globals.yml
```diff
@@ -10,16 +10,20 @@ enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names or ansible_facts
...
-openstack_tag: "{% raw %}{{ kayobe_image_tags['openstack'][kolla_base_distro] }}{% endraw %}"
-ovn_tag: "{% raw %}{{ kayobe_image_tags['ovn'][kolla_base_distro] }}{% endraw %}"
+# Dict of Kolla image tags to deploy for each service.
+# Each key is the tag variable prefix name, and the value is another dict,
+# where the key is the OS distro and the value is the tag to deploy.
+# NOTE: This is defined in etc/kayobe/kolla-image-tags.yml.
+kolla_image_tags:
+{{ kolla_image_tags | to_nice_yaml | indent(width=4, first=true) }}
+
+{% for tag in kolla_image_tags %}
```
jovial:
```suggestion
{% for tag in kolla_image_tags %}
```

Couldn't we hardcode this list? That way we could get proper precedence:

```
nova_libvirt_tag:  "{% raw %}{{{ kolla_image_tags['nova-libvirt'] | default(nova_tag) | default(openstack_tag) }}{% endraw %}"
```

Then it just leave it permanently in the file. In my opinion it easier to understand what is going on.

jovial: I guess this isn't actually an issue if you had the full list of containers :+1: I was imagining we'd still have groups of containers like 'nova'.

markgoddard: That's an option. I was looking at writing a script that would generate the necessary variables based on kolla_image_tags, but maybe that's trying to be too "clever"?

I'd probably want to put the variables in a separate file and copy them to "declutter" it a bit though.

markgoddard: This version is still using the groups such as nova. The conflict resolution issues in the full image list prevented it from getting finished so I thought I'd try this hybrid approach.

tools/kolla-images.py
```diff
@@ -0,0 +1,365 @@
...
+
+def get_openstack_release() -> str:
+    """Return the OpenStack release."""
+    with open(get_abs_path(".gitreview"), "r") as f:
+        gitreview = f.readlines()
+    for line in gitreview:
+        if "=" not in line:
+            continue
+        key, value = line.split("=")
+        if key.strip().strip() == "defaultbranch":
```
priteau: What is the reason for calling `strip()` twice?

tools/kolla-images.py
```diff
@@ -0,0 +1,365 @@
...
+            value = value.strip().rstrip()
```
priteau: What is the reason for calling both `strip()` and `rstrip()`?

tools/kolla-images.py
```diff
@@ -0,0 +1,365 @@
...
+            prefix = "stable/"
+            assert value.startswith(prefix)
+            return value[len(prefix):]
```
priteau: Maybe add support for the `unmaintained/` prefix already?

tools/kolla-images.py
```diff
@@ -0,0 +1,365 @@
...
+
+def check_tags(base_distros: List[str], kolla_image_tags: KollaImageTags, registry: str, namespace: str):
+    """Check whether expected tags are present in container image registry."""
+    image_tags = get_tags(base_distros, kolla_image_tags)
+
+    missing = {}
+    for image, tags in image_tags.items():
+        for _ in range(3):
+            try:
+                output = subprocess.check_output(f"skopeo list-tags docker://{registry}/{namespace}/{image}", shell=True)
```
priteau, line 262: Is skopeo expected to be present?

priteau (changes requested):
Nice code 👍 Just a few minor comments inline. Also please fix merge conflicts.

tools/kolla-images.py
markgoddard: Sleep deprivation? Paranoia? I'll remove it.

tools/kolla-images.py
markgoddard: Just one strip should do the job.

tools/kolla-images.py
markgoddard: I'll take a look at this.

tools/kolla-images.py
markgoddard: Yes. Currently this is just used for local debugging. I could add a `type skopeo` to fail early if it's not installed.

tools/kolla-images.py
priteau: It may be required only if:

1) upstream kayobe-config changes `stable/yoga` to `unmaintained/yoga`. Would Yoga be marked eol instead?
2) we merge upstream kayobe-config into stackhpc-kayobe-config (we don't routinely do this at the moment)

tools/kolla-images.py
markgoddard: Done

jovial (approved):
Lets give a whirl. Seems like it will save some time :)

PR #664, opened 2023-09-27
[INFRA-351] enable ML2/OVN and DVR by default

g0rgamesh:


etc/kayobe/kolla.yml
```diff
@@ -541,7 +541,7 @@ kolla_build_customizations: "{{ kolla_build_customizations_common | combine(koll
 #kolla_enable_neutron:
 #kolla_enable_neutron_agent_ha:
 #kolla_enable_neutron_bgp_dragent:
-#kolla_enable_neutron_dvr:
+kolla_enable_neutron_dvr: yes
```
markgoddard: This flag is for neutron OVS DVR.

We need to set `neutron_ovn_distributed_fip: true` in `etc/kayobe/kolla/globals.yml`. See https://docs.openstack.org/kolla-ansible/latest/reference/networking/neutron.html#ovn-ml2-ovn

MoteHue:
I believe that DVR in the ML2/OVS sense is implicitly enabled when using ML2/OVN, so `kolla_enable_neutron_dvr: true` does not need to be set. We will need to set `neutron_ovn_distributed_fip: true` if we want to have the scalable north/south routing. 

Third question here helps to explain a bit: https://docs.openstack.org/neutron/latest/ovn/faq/index.html

etc/kayobe/kolla.yml
```diff
@@ -565,7 +565,7 @@ kolla_build_customizations: "{{ kolla_build_customizations_common | combine(koll
 #kolla_enable_openvswitch:
 #kolla_enable_osprofiler:
 #kolla_enable_outward_rabbitmq:
-#kolla_enable_ovn:
+kolla_enable_ovn: yes
```
markgoddard, line 598: We can now remove this flag from these files:

* `etc/kayobe/environments/aufn-ceph/kolla.yml`
* `etc/kayobe/environments/ci-builder/stackhpc-ci.yml`
* `etc/kayobe/environments/ci-multinode/kolla.yml`

markgoddard (changes requested):
This change needs to go into the stackhpc/zed branch as it's not backward compatible.

Please also add a clear release note with an upgrade section.

g0rgamesh:
> I believe that DVR in the ML2/OVS sense is implicitly enabled when using ML2/OVN, so `kolla_enable_neutron_dvr: true` does not need to be set. We will need to set `neutron_ovn_distributed_fip: true` if we want to have the scalable north/south routing.
> 
> Third question here helps to explain a bit: https://docs.openstack.org/neutron/latest/ovn/faq/index.html

Good catch, thank you.

etc/kayobe/kolla.yml
g0rgamesh: Thanks, will do.

mnasiadka:
> I believe that DVR in the ML2/OVS sense is implicitly enabled when using ML2/OVN, so `kolla_enable_neutron_dvr: true` does not need to be set. We will need to set `neutron_ovn_distributed_fip: true` if we want to have the scalable north/south routing. 
> 
> 
> 
> Third question here helps to explain a bit: https://docs.openstack.org/neutron/latest/ovn/faq/index.html

To be clear, kolla_enable_neutron_dvr is for ML2/OVS only

g0rgamesh:
> This change needs to go into the stackhpc/zed branch as it's not backward compatible.
> 
> Please also add a clear release note with an upgrade section.

Done.

markgoddard:
> > This change needs to go into the stackhpc/zed branch as it's not backward compatible.
> > Please also add a clear release note with an upgrade section.
> 
> Done.

Still targeting stackhpc/yoga

etc/kayobe/kolla.yml
markgoddard: and the other two?

g0rgamesh:
> > > This change needs to go into the stackhpc/zed branch as it's not backward compatible.
> > > Please also add a clear release note with an upgrade section.
> > 
> > 
> > Done.
> 
> Still targeting stackhpc/yoga

Yep, my bad: https://github.com/stackhpc/stackhpc-kayobe-config/pull/667

etc/kayobe/kolla.yml
g0rgamesh: This branch was created from master, I didn't notice PR was targeting Yoga by default.

markgoddard:
Wrong branch.

PR #678, opened 2023-09-29
Add Openstack Capacity to StackHPC Kayobe Config

assumptionsandg:
This patch adds [Openstack Capacity](https://github.com/stackhpc/os-capacity) to StackHPC Kayobe Config.

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/prometheus.yml.d/70-oscapacity.yml.j2
+++ b/etc/kayobe/kolla/config/prometheus/prometheus.yml.d/70-oscapacity.yml.j2
@@ -1,0 +1,9 @@
+{% raw %}
+scrape_configs:
+  - job_name: os-capacity
+    static_configs:
+        - targets:
+          - '{{ kolla_internal_vip_address | put_address_in_context('url') }}:9000'
+    scrape_interval: 15m
+    scrape_timeout: 10m
+{% endraw %}
--- a/releasenotes/notes/os-capacity-94006f03f16583e4.yaml
+++ b/releasenotes/notes/os-capacity-94006f03f16583e4.yaml
@@ -7,7 +7,9 @@ features:
     this change.
 upgrade:
   - |
-    To deploy Openstack Capacity, you must define Openstack
-    application credential variables: secrets_os_exporter_auth_url,
-    secrets_os_exporter_credential_id and secrets_os_exporter_
-    credential_secret as laid out in the 'Monitoring' documentation.
+    To deploy the Openstack Capacity Grafana dashboard, you must
+    define Openstack application credential variables:
+    ``secrets_os_exporter_auth_url``,
+    ``secrets_os_exporter_credential_id`` and 
+    ``secrets_os_exporter_credential_secret`` as laid out in the
+    'Monitoring' documentation.
--- a/releasenotes/notes/os-capacity-94006f03f16583e4.yaml
+++ b/releasenotes/notes/os-capacity-94006f03f16583e4.yaml
@@ -1,0 +1,13 @@
+---
+features:
+  - |
+    This patch adds Openstack Capacity metrics and exporters to
+    StackHPC Kayobe Config. This includes a deployment playbook,
+    Prometheus scrape jobs and HAProxy configurations to support 
+    this change.
+upgrade:
+  - |
+    To deploy Openstack Capacity, you must define Openstack
+    application credential variables: secrets_os_exporter_auth_url,
+    secrets_os_exporter_credential_id and secrets_os_exporter_
+    credential_secret as laid out in the 'Monitoring' documentation.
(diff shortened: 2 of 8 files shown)


Alex-Welsh:
I think we need two more things.

1. Either some documentation on what the playbook is an how to deploy it OR a hook to run it automatically
2. A release note

Alex-Welsh:
Current CI failures are just from a lack of space on SMS. As long as one of them passes, I think we an force push these changes through

mnasiadka:
I would vote for some documentation, especially we require defined variables in secrets

assumptionsandg:
> I would vote for some documentation, especially we require defined variables in secrets

Updated the monitoring docs to include information on how to define secrets.

Alex-Welsh:
Couple of tox docs errors but otherwise looks good

assumptionsandg:
> Couple of tox docs errors but otherwise looks good

Classic mistake :sweat_smile:

technowhizz:
We should merge the PR on smslab that is pending first and ensure all comments from there are addressed to avoid having two open PR's with different discussions @assumptionsandg

assumptionsandg:
Removed the playbook hook from service deploy to leave for manual deployment, as this exporter requires application credentials to be defined prior to deploying the exporter.

Alex-Welsh:
> We should merge the PR on smslab that is pending first and ensure all comments from there are addressed to avoid having two open PR's with different discussions @assumptionsandg

I agree that we should resolve all the conversations, but I think it would be best to merge this one first, then sync smslab/yoga with skc/yoga so we don't end up with any merge weirdness. Or we could cherry-pick this commit to the SMS PR so the hash is the same which would have the same effect. Either way, best to get this through asap

releasenotes/notes/os-capacity-94006f03f16583e4.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    This patch adds Openstack Capacity metrics and exporters to
+    StackHPC Kayobe Config. This includes a deployment playbook
+    as a hook and HAProxy configurations to support this change.
```
JohnGarbutt: I think we should note there is upgrade impact with this change, in that you need to add those new secrets to be make os-capacity work?

What happens when people start using this, and they are missing the secrets? Do we get alerts about missing scrape targets?

holliefae: Wouldn't the PrometheusTargetMissing alert pick this up if the exporter was improperly configured and crashed?

releasenotes/notes/os-capacity-94006f03f16583e4.yaml
```diff
@@ -0,0 +1,13 @@
...
+    This patch adds Openstack Capacity metrics and exporters to
+    StackHPC Kayobe Config. This includes a deployment playbook,
+    Prometheus scrape jobs and HAProxy configurations to support 
+    this change.
+upgrade:
+  - |
+    To deploy Openstack Capacity, you must define Openstack
+    application credential variables: secrets_os_exporter_auth_url,
+    secrets_os_exporter_credential_id and secrets_os_exporter_
+    credential_secret as laid out in the 'Monitoring' documentation.
```
Alex-Welsh:
```suggestion
    To deploy the Openstack Capacity Grafana dashboard, you must
    define Openstack application credential variables:
    ``secrets_os_exporter_auth_url``,
    ``secrets_os_exporter_credential_id`` and 
    ``secrets_os_exporter_credential_secret`` as laid out in the
    'Monitoring' documentation.
```

Alex-Welsh (dismissed):
There's a few small things we could change with the documentation, and you could add a link in the release note to the relevant section, but I think we're hitting diminishing returns on polishing this PR. I'm happy with it as is

releasenotes/notes/os-capacity-94006f03f16583e4.yaml
holliefae: HAProxy `HaproxyBackendDown` and `HaproxyServerDown` alerts get triggered when the auth file gets misconfigured. No Prometheus alerts get triggered and the exporter disappears form the Prometheus targets. I'll update the docs accordingly so people know what to look out for.

Alex-Welsh:
Something funky seems to have happened to your rebase. Might need a bit of git untangling

assumptionsandg:
> Something funky seems to have happened to your rebase. Might need a bit of git untangling

Not sure what happened with that rebase, should be good now though! :smile:

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/70-oscapacity.yml.j2
```diff
@@ -0,0 +1,9 @@
+{% raw %}
+scrape_configs:
+  - job_name: os-capacity
```
markgoddard, line 3: This will fail on a system with internal TLS. Requires this:
```
{% if kolla_enable_tls_internal | bool %}
    scheme: https
{% endif %}
```

markgoddard: Bumping this issue.

PR #682, opened 2023-10-03
Become during /opt/kayobe/vault creation

Alex-Welsh:
The vault deploy overcloud playbook can fail when trying to create /opt/kayobe/vault without privilege escalation

### DIFF
--- a/etc/kayobe/ansible/vault-deploy-overcloud.yml
+++ b/etc/kayobe/ansible/vault-deploy-overcloud.yml
@@ -43,6 +43,7 @@
       file:
         path: /opt/kayobe/vault
         state: directory
+      become: true
 
     - name: Template out TLS key and cert
       copy:


etc/kayobe/ansible/vault-deploy-overcloud.yml
```diff
@@ -43,6 +43,7 @@
       file:
         path: /opt/kayobe/vault
         state: directory
+      become: true
```
jovial, line 46: I guess this means the folder is owned by root, should explicitly set an owner and group?

markgoddard: I think that typically `/opt/kayobe` gets created during this task:
```yaml
        - name: Ensure global virtualenv directory exists
          file:
            path: "{{ virtualenv_path }}"
            state: directory
            owner: "{{ ansible_facts.user_uid }}"
            group: "{{ ansible_facts.user_gid }}"
            mode: 0755
          # Check whether the virtualenv directory is a subdirectory of the
          # global virtualenv directory.
          when: virtualenv.startswith(virtualenv_path)
          become: True
```

So it will be stack:stack 0755. I wonder what we require here? I suppose root:root 0700 would be most secure, if that works.

Alex-Welsh:
Long since stale, hasn't been required for over a year. Closing

PR #683, opened 2023-10-04
Add debugging info to tls deployment docs

Alex-Welsh:
Just adding a note about an error I've seen a couple of times. Not sure exactly what the cause of it is, might be related to admin endpoint deletion

### DIFF
--- a/doc/source/configuration/vault.rst
+++ b/doc/source/configuration/vault.rst
@@ -229,6 +229,18 @@ Enable the required TLS variables in kayobe and kolla
 
       kayobe overcloud service deploy
 
+   If VM provisioning fails with an error with this format:
+
+   .. code-block::
+
+      Unable to establish connection to http://<kolla internal vip/fqdn>:9696/v2.0/ports/some-sort-of-uuid: Connection aborted
+
+   Restart the nova-compute container on all hypervisors:
+
+   .. code-block::
+
+      kayobe overcloud host command run --command "docker restart nova_compute" --become --show-output -l compute
+
 Barbican integration
 ====================
 


doc/source/configuration/vault.rst
```diff
@@ -229,6 +229,18 @@ Enable the required TLS variables in kayobe and kolla
 
       kayobe overcloud service deploy
 
+   If VM provisioning fails with an error with this format:
```
markgoddard, line 232: Did this happen after completion of the service deploy? It seems odd that it's using an HTTP endpoint.

markgoddard:
> Just adding a note about an error I've seen a couple of times. Not sure exactly what the cause of it is, might be related to admin endpoint deletion

Did you delete the admin endpoints before running service deploy? Should that be a separate step rather than a warning?

doc/source/configuration/vault.rst
Alex-Welsh: > Did you delete the admin endpoints before running service deploy? Should that be a separate step rather than a warning?

> Did this happen after completion of the service deploy? It seems odd that it's using an HTTP endpoint.

I did delete the admin endpoints before the service deploy, and provisioning did not work after the service deploy. I thought it was very strange that it didn't pick up the changes and needed a manual restart.

markgoddard: Had the nova_compute containers been restarted during the service deploy?

Alex-Welsh: Not sure, the docker logs don't really go back far enough

markgoddard (approved):

PR #684, opened 2023-10-04
Fail container build workflow when no images build

Alex-Welsh:
QOL feature to fail if no images are built.

Moved the prune task earlier so that it a) always runs and b) will run before any images are built to hopefully stop it running out of space mid-build


Alex-Welsh:
Test workflows:
this should fail https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/6406674501
this should succeed https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/6406678539

jovial (approved):

markgoddard (approved):

PR #686, opened 2023-10-05
[INFRA-351] enable ML2/OVN and DVR by default

g0rgamesh:


markgoddard:
Not necessary, it will get merged from zed to 2023.1

PR #691, opened 2023-10-06
Bump Zed container image tags

Alex-Welsh:
I'm not quite sure where these tags got lost but they should've been merged in a while ago. We ran a full rebuild to bring in vulnerability mitigations to the base images.

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -12,8 +12,8 @@ kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %
 
 kayobe_image_tags:
   openstack:
-    rocky: zed-rocky-9-20230821T155947
-    ubuntu: zed-ubuntu-jammy-20230821T155947
+    rocky: zed-rocky-9-20230921T153510
+    ubuntu: zed-ubuntu-jammy-20230921T153510
   bifrost:
     rocky: zed-rocky-9-20230927T142529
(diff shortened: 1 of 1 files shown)


etc/kayobe/kolla/globals.yml
```diff
@@ -12,8 +12,8 @@ kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %
 
 kayobe_image_tags:
   openstack:
-    rocky: zed-rocky-9-20230821T155947
-    ubuntu: zed-ubuntu-jammy-20230821T155947
+    rocky: zed-rocky-9-20230921T153510
```
markgoddard, line 15: I'm a bit confused - the rocky 9 snapshots were only last week and are coming into zed via this open [PR](https://github.com/stackhpc/stackhpc-kayobe-config/pull/692/files)

markgoddard: It's probably not a huge issue - the main thing is that the host repos are updated for the vulnerability fixes

Alex-Welsh: Might've rebuilt them for a different reason. Either way, more up to date containers is more good right?

markgoddard: Sure.

markgoddard: Especially tested ones

markgoddard:
Ubuntu VMs failed to delete (due to zed upgrade?) but tempest passed.

PR #696, opened 2023-10-07
feat: add `cephadm` maintenance playbooks

jackhodgkiss:
Add two playbooks for entering and exiting [maintenance mode](https://docs.ceph.com/en/latest/dev/cephadm/host-maintenance/) for a given Ceph node.

```shell
kayobe playbook run $KAYOBE_CONFIG_PATH/ansible/cephadm-enter-maintenance.yml --limit ceph-mon-01

kayobe playbook run $KAYOBE_CONFIG_PATH/ansible/cephadm-exit-maintenance.yml --limit ceph-mon-01
```

Note these playbooks use [stackhpc.cephadm.commands](https://github.com/stackhpc/ansible-collection-cephadm) which will delegate the command to the first `mon` within your inventory. If this node is in maintenance you must specify `--cephadm_delegate_host` and provide another `mon`.

```
kayobe playbook run $KAYOBE_CONFIG_PATH/ansible/cephadm-exit-maintenance.yml --limit ceph-mon-01 --cephadm_delegate_host ceph-mon-02
```

**Note**: this relies on something such as [stackhpc/ansible-collection-cephadm/pull/109](https://github.com/stackhpc/ansible-collection-cephadm/pull/109) being merged with some additional changes.


Alex-Welsh:
> Note these playbooks use [stackhpc.cephadm.commands](https://github.com/stackhpc/ansible-collection-cephadm) which will delegate the command to the first mon within your inventory. If this node is in maintenance you must specify --cephadm_delegate_host and provide another mon.

Is there a way of determining the hosts that are not in maintenance and selecting one of them? That would make things a lot more simple.

jackhodgkiss:
> > Note these playbooks use [stackhpc.cephadm.commands](https://github.com/stackhpc/ansible-collection-cephadm) which will delegate the command to the first mon within your inventory. If this node is in maintenance you must specify --cephadm_delegate_host and provide another mon.
> 
> Is there a way of determining the hosts that are not in maintenance and selecting one of them? That would make things a lot more simple.

I suppose it is possible. However I see this as no different to how we handle controllers and VIP and intentionally avoid the VIP until the end.

My concern would be if a host is in maintenance the command gets trapped as it will proceed to authenticate with the cluster and silently fail so it would involve timeouts and other work arounds.

etc/kayobe/ansible/cephadm-enter-maintenance.yml
```diff
@@ -0,0 +1,32 @@
...
+
+        - name: Ensure active manager has been switched to another node
+          ansible.builtin.import_role:
+            name: stackhpc.cephadm.commands
+          vars:
+            cephadm_commands:
+              - "mgr fail {{ active_ceph_mgr }}"
+      when: '"Cannot stop active Mgr daemon" in cephadm_commands_result.results[0].stderr'
+
+    - name: Ensure host is in maintenance mode
```
Alex-Welsh, line 27: I'm not entirely sure how the cpehadm_commands role works, but it might be worth having one last play to ensure the new state is set, and erroring if not

etc/kayobe/ansible/cephadm-exit-maintenance.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Ensure a Ceph host has exited host maintenance
```
Alex-Welsh, line 2: ditto

Alex-Welsh:
> > > Note these playbooks use [stackhpc.cephadm.commands](https://github.com/stackhpc/ansible-collection-cephadm) which will delegate the command to the first mon within your inventory. If this node is in maintenance you must specify --cephadm_delegate_host and provide another mon.
> > 
> > 
> > Is there a way of determining the hosts that are not in maintenance and selecting one of them? That would make things a lot more simple.
> 
> I suppose it is possible. However I see this as no different to how we handle controllers and VIP and intentionally avoid the VIP until the end.
> 
> My concern would be if a host is in maintenance the command gets trapped as it will proceed to authenticate with the cluster and silently fail so it would involve timeouts and other work arounds.

Does it error gracefully if the node is in maintenance? If not it might be worth adding "precheck" task to verify

jackhodgkiss:
> > > > Note these playbooks use [stackhpc.cephadm.commands](https://github.com/stackhpc/ansible-collection-cephadm) which will delegate the command to the first mon within your inventory. If this node is in maintenance you must specify --cephadm_delegate_host and provide another mon.
> > > 
> > > 
> > > Is there a way of determining the hosts that are not in maintenance and selecting one of them? That would make things a lot more simple.
> > 
> > 
> > I suppose it is possible. However I see this as no different to how we handle controllers and VIP and intentionally avoid the VIP until the end.
> > My concern would be if a host is in maintenance the command gets trapped as it will proceed to authenticate with the cluster and silently fail so it would involve timeouts and other work arounds.
> 
> Does it error gracefully if the node is in maintenance? If not it might be worth adding "precheck" task to verify

I will have to check but I think ceph has a tendency to return 0 regardless of it being successful or not.

etc/kayobe/ansible/cephadm-enter-maintenance.yml
```diff
@@ -0,0 +1,32 @@
...
+      ansible.builtin.import_role:
+        name: stackhpc.cephadm.commands
+      vars:
+        cephadm_commands:
+          - "orch host maintenance enter {{ ansible_facts.nodename }}"
```
cityofships, line 32: Won't be possible for any host holding RGW services - gets:
```
WARNING: Removing RGW daemons can cause clients to lose connectivity.
Note: Warnings can be bypassed with the --force flag
```
Of course `--force` defeats the purpose of other checks and is not viable here.

jackhodgkiss:
@jackhodgkiss testing slack notifications.

markgoddard:
I reworked these into roles in the cephadm collection: https://github.com/stackhpc/ansible-collection-cephadm/pull/153. Once that merges I'll propose some playbooks in SKC.

PR #698, opened 2023-10-09
ci-aio: LVM conditonal fix

assumptionsandg:
Quick fix to avoid this syntax error during deployment.

+ '[' sudo vgdisplay                                                                                     
./deploy.sh: line 9: [: missing `]'                                                                      
+ grep -q lvm2 ']'                                                                                       
grep: ]: No such file or directory

### DIFF
--- a/etc/kayobe/environments/ci-aio/automated-setup.sh
+++ b/etc/kayobe/environments/ci-aio/automated-setup.sh
@@ -6,7 +6,7 @@ cat << EOF | sudo tee -a /etc/hosts
 10.205.3.187 pulp-server pulp-server.internal.sms-cloud
 EOF
 
-if [ sudo vgdisplay | grep -q lvm2 ]; then
+if [ "sudo vgdisplay | grep -q lvm2" ]; then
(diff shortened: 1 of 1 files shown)


etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -6,7 +6,7 @@ cat << EOF | sudo tee -a /etc/hosts
 10.205.3.187 pulp-server pulp-server.internal.sms-cloud
 EOF
 
-if [ sudo vgdisplay | grep -q lvm2 ]; then
+if [ "sudo vgdisplay | grep -q lvm2" ]; then
```
markgoddard: If you just want to test the exit code you can do this:
```suggestion
if sudo vgdisplay | grep -q lvm2; then
```

markgoddard: I think your change tests the string rather than executing the command.

holliefae: You're right it was passing the conditional on the string, thanks for the fix!

markgoddard (approved):

PR #699, opened 2023-10-09
DNM: CloudKitty  for fresh OpenSearch deployments

MoteHue:
This container tag brings in the fixes from this PR: https://github.com/stackhpc/cloudkitty/pull/57 which allows CloudKitty with the elasticsearch backend to be used with fresh OpenSearch. i.e for new deployments (Habrok, AZ Moldnal, etc.)

This PR should not be merged until we are happy with using the same patch for existing deployments after an ES->OS migration. (Or ideally, never at all if we want to integrate the proper opensearch v2 storage driver)


Alex-Welsh:
Is this still required/working?

MoteHue:
The proper solutions to this are making their way through upstream. IMO we should keep this around until that's fully merged and backported, just in case someone needs this in the meantime.

PR #702, opened 2023-10-10
Reduce ESP partition size in deployment image

cityofships:
It does not fit into 550MiB partition created by Ironic Python Agent. This is relevant when provisioning software RAID UEFI enabled servers.

https://github.com/openstack/ironic-python-agent/blob/stable/yoga/ironic_python_agent/raid_utils.py#L30

### DIFF
--- a/etc/kayobe/pulp-host-image-versions.yml
+++ b/etc/kayobe/pulp-host-image-versions.yml
@@ -3,6 +3,6 @@
 # These images must be in SMS, since they are used by our AIO CI runners
 stackhpc_centos_8_stream_overcloud_host_image_version: "yoga-20230525T095243"
 stackhpc_rocky_8_overcloud_host_image_version: "yoga-20230629T135322"
-stackhpc_rocky_9_overcloud_host_image_version: "yoga-20230929T133006"
+stackhpc_rocky_9_overcloud_host_image_version: "yoga-20231012T121552"
 stackhpc_ubuntu_focal_overcloud_host_image_version: "yoga-20230609T120720"
-stackhpc_ubuntu_jammy_overcloud_host_image_version: "yoga-20230609T120720"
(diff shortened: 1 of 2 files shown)


mnasiadka (dismissed):

markgoddard:
Do we need new host images with this change included?

cityofships:
> Do we need new host images with this change included?

Yes. Before that's available, one can either virt-customize local image or use IPA with `ESP_SIZE_MIB` modified to something bigger.

cityofships:
Building in: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/6495483910

priteau:
Is this an upstream issue too? DIB defaults to 550 MiB and IPA too, but the DIB image doesn't fit into what IPA creates? Maybe IPA should be just slightly bigger?

mnasiadka:
> Is this an upstream issue too? DIB defaults to 550 MiB and IPA too, but the DIB image doesn't fit into what IPA creates? Maybe IPA should be just slightly bigger?

I think it's rather the issue of our layout that we pass to DIB?

mnasiadka (approved):

mnasiadka:
I'll rebuild the images as part of https://github.com/stackhpc/stackhpc-kayobe-config/pull/721

cityofships:
> > Is this an upstream issue too? DIB defaults to 550 MiB and IPA too, but the DIB image doesn't fit into what IPA creates? Maybe IPA should be just slightly bigger?
> 
> I think it's rather the issue of our layout that we pass to DIB?

It's probably rounding / calculations error, I saw the ESP partition 0,1MB too big for MD created by IPA. It's easier to change here than a hard-coded value in IPA for us.

mnasiadka:
Merging as this is not tested in aio

PR #703, opened 2023-10-10
Add ethtool and pciutils to the host disk image

priteau:


Alex-Welsh:
Are these being added to fix a specific error, or as general debugging tools?

priteau:
As general debugging tools, they are useful to check for networking issues when we may not yet be able to install packages.

Alex-Welsh (approved):

cityofships (approved):

PR #704, opened 2023-10-10
Antelope image builds & CI passing

Alex-Welsh:


etc/kayobe/inventory/group_vars/overcloud/docker
markgoddard, line 1: I expect this will be required for all hosts running docker, not just overcloud. It could be added to etc/kayobe/docker.yml.

etc/kayobe/kolla.yml
```diff
@@ -81,11 +81,6 @@ kolla_ansible_source_version: "{{ stackhpc_kolla_ansible_source_version }}"
 # Docker namespace to use for Kolla images. Default is 'kolla'.
 kolla_docker_namespace: stackhpc
 
-# Whether docker should be configured to use an insecure registry for Kolla
-# images. Default is false, unless docker_registry_enabled is true and
-# docker_registry_enable_tls is false.
-kolla_docker_registry_insecure: "{{ 'https' not in stackhpc_repo_mirror_url }}"
```
markgoddard, line 87: https://review.opendev.org/c/openstack/kayobe/+/898091

etc/kayobe/environments/ci-builder/stackhpc-ci.yml
```diff
@@ -25,6 +25,7 @@ kolla_enable_octavia: true
 kolla_enable_opensearch: true
 kolla_enable_prometheus: true
 kolla_enable_redis: true
+kolla_enable_skyline: true
```
markgoddard: Let's decide as a group whether or not to support this.

etc/kayobe/kolla.yml
```diff
@@ -224,7 +219,7 @@ stackhpc_ubuntu_jammy_repos:
   - "deb {{ stackhpc_repo_ubuntu_jammy_url }} jammy-updates main universe"
   - "deb {{ stackhpc_repo_ubuntu_jammy_url }} jammy-backports main universe"
   - "deb {{ stackhpc_repo_ubuntu_jammy_security_url }} jammy-security main universe"
-  - "deb {{ stackhpc_repo_ubuntu_cloud_archive_url }} jammy-updates/{{ openstack_release }} main"
+  - "deb {{ stackhpc_repo_ubuntu_cloud_archive_url }} jammy-updates/antelope main"
```
markgoddard: I don't think this is too helpful. It shouldn't need to change between releases.

etc/kayobe/pulp.yml
```diff
@@ -102,7 +102,7 @@ stackhpc_pulp_deb_repos:
     distribution_name: "ubuntu-cloud-archive-"
     base_path: "ubuntu-cloud-archive/"
     components: "main"
-    distributions: "jammy-updates/{{ openstack_release }}"
+    distributions: "jammy-updates/antelope"
```
markgoddard: Again, this shouldn't need to change between releases.

I'd prefer that we annotate things that may need to change.

markgoddard (changes requested):

etc/kayobe/kolla.yml
Alex-Welsh: This was actually causing issues because `{{ openstack_release }}` resolves to `2023.1` rather than `antelope`

etc/kayobe/pulp.yml
Alex-Welsh: Same as https://github.com/stackhpc/stackhpc-kayobe-config/pull/704#discussion_r1356534594

etc/kayobe/environments/ci-builder/stackhpc-ci.yml
Alex-Welsh: Removed

etc/kayobe/inventory/group_vars/overcloud/docker
Alex-Welsh: Done

etc/kayobe/kolla.yml
markgoddard: Doh. Let's introduce a new variable, openstack_release_name. It can be annotated to state that it will need to be updated.

etc/kayobe/openstack.yml
```diff
@@ -5,6 +5,10 @@
 # Name of the current OpenStack release. Default is "2023.1".
 #openstack_release:
 
+# Codename of the current OpenStack release. Current release codename is
+# "antelope". This must be manually updated with each release.
+openstack_release_codename: "antelope"
```
Alex-Welsh: It was weirdly difficult to word this, and I'm still not exactly happy with it. Suggestions are welcome

markgoddard: I'd go with something like this:
```
# Codename of the current OpenStack release.
# NOTE(upgrade): Update to current release codename.
openstack_release_codename: "antelope"
```
Then in the wiki page we can say to grep for `NOTE(upgrade)`

Alex-Welsh: I like it. I'll start to add the note to other places too

Alex-Welsh: > I'd go with something like this:
> 
> ```
> # Codename of the current OpenStack release.
> # NOTE(upgrade): Update to current release codename.
> openstack_release_codename: "antelope"
> ```
> 
> Then in the wiki page we can say to grep for `NOTE(upgrade)`

Done

markgoddard (approved):

PR #712, opened 2023-10-13
Adapt RabbitMQ connection alert threshold

priteau:
The threshold of 1000 connections to a RabbitMQ server can be easily reached even on moderately-sized deployments.

On several deployments we saw around 25 connections per hypervisor and between 250 and 750 connections per controller. The threshold is based on twice these numbers to be conservative.


priteau:
This has been tested successfully on a production deployment.

markgoddard (approved):

PR #714, opened 2023-10-16
Seed image build fix antelope

Alex-Welsh:
Seed image builds were previously failing. This change cherry-picks the fix and adds tags for bifrost with newly built images


Alex-Welsh:
images with tag *-20231013T151957 are currently untested because RT runners are down, so no sync has taken place.

markgoddard (approved):

PR #715, opened 2023-10-17
Fix value of libvirt.hw_machine_type

priteau:
We are seeing the following entries in nova-compute.log:

    WARNING nova.virt.libvirt.utils [-] Invalid hw_machine_type config value q35

The format for this config option is `host-arch=machine-type`. For example: `x86_64=machinetype1,armv7l=machinetype2`.

### DIFF
--- a/etc/kayobe/kolla/config/nova.conf
+++ b/etc/kayobe/kolla/config/nova.conf
@@ -1,2 +1,2 @@
 [libvirt]
-hw_machine_type = q35
+hw_machine_type = x86_64=q35


darmach (commented):
That's interesting, should we correct also this example in the doc, or is it correct and in `openstack image set` we should set `hw_machine_type` to "just" `q35'?

https://docs.openstack.org/nova/latest/admin/hw-machine-type.html#configure

priteau:
> That's interesting, should we correct also this example in the doc, or is it correct and in `openstack image set` we should set `hw_machine_type` to "just" `q35'?
> 
> https://docs.openstack.org/nova/latest/admin/hw-machine-type.html#configure

I assume the documentation is correct, because the image also has an architecture which can be set via another property.

But when you set the default machine type via nova.conf, the expected format includes the architecture. I assume this is because a hypervisor can support multiple architectures if you are using QEMU with emulation. See https://docs.openstack.org/nova/latest/configuration/config.html#libvirt.hw_machine_type

mnasiadka (approved):

PR #716, opened 2023-10-19
OVN packages version bump to 22.12

darmach:
Need to bump to 22.12 to get to the same level as already deployed on many customers yoga-20231019T102525.


darmach:
Built new containers centos and rocky for ovn and openvswitch: yoga-20231019T102525

mnasiadka (approved):

PR #717, opened 2023-10-19
Fix custom sca policies remote checks

MoteHue:
The config option to allow for remote commands needs to be set on the agents, not the manager. This fixes many benchmarks being makred as not applicable.


Alex-Welsh:
Reno?

MoteHue:
Previous PR for context: https://github.com/stackhpc/stackhpc-kayobe-config/pull/711

etc/kayobe/ansible/wazuh-manager.yml
```diff
@@ -125,3 +115,40 @@
...
+- hosts: wazuh-agent
+  become: yes
+  become_user: root
+  tasks:
+    - name: Check if custom SCA policies directory exists
+      stat:
+        path: "{{ local_custom_sca_policies_path }}"
+      register: custom_sca_policies_folder
+      delegate_to: localhost
+      become: no
```
Alex-Welsh: Would it not be clearer to become on each task that is necessary, rather than become on everything and invert it here?

MoteHue:
> Reno?

The original PR has a reno, not sure if this also needs one as it just follows up on that?

etc/kayobe/ansible/wazuh-manager.yml
```diff
@@ -125,3 +115,40 @@
...
+
+    - name: Gather list of custom SCA policies
```
Alex-Welsh: Why run this task as root but not the last one?

etc/kayobe/ansible/wazuh-manager.yml
```diff
@@ -125,3 +115,40 @@
...
+      register: custom_sca_policies
+      when: custom_sca_policies_folder.stat.exists
+
+    - name: Allow Wazuh agents to execute commands in SCA policies sent from the Wazuh manager
+      blockinfile:
+        path: "/var/ossec/etc/local_internal_options.conf"
+        state: present
+        owner: wazuh
+        group: wazuh
+        block: |
```
Alex-Welsh: Slightly unnecessary to have this as a multiline string when you only have one line

etc/kayobe/ansible/wazuh-manager.yml
MoteHue: Now that I've introduces many more tasks into the playbook, I'd agree this would be clearer. Will update the PR accordingly.

etc/kayobe/ansible/wazuh-manager.yml
MoteHue: I just missed a become: no here tbh. Have updated the becomes now

etc/kayobe/ansible/wazuh-manager.yml
MoteHue: True, have corrected

jovial (dismissed):
Technically, I guess we should have a reno, but the chance of someone:
a) using this feature, and
b) using a commit in between when the feature landed and this bugfix
is low, so I guess we could let it slide.

etc/kayobe/ansible/wazuh-manager.yml
```diff
@@ -125,3 +116,38 @@
       service:
         name: wazuh-dashboard
         state: restarted
+
+- hosts: wazuh-agent
```
markgoddard: Should this play be in wazuh-agents.yml?

etc/kayobe/ansible/wazuh-manager.yml
```diff
@@ -9,7 +9,6 @@
     - generate-certs
 # Single node
 - hosts: wazuh-manager
-  become: yes
```
markgoddard: I trust that the wazuh-ansible roles are happy with this? If so a nice improvement.

MoteHue: I haven't tested from a fresh install, but didn't hit any issues when I ran through this to test the agent changes

etc/kayobe/ansible/wazuh-manager.yml
MoteHue: Yes seems sensible, I've moved them over

etc/kayobe/ansible/wazuh-manager.yml
MoteHue: Given a lot is done in the roles, perhaps this should be tested in a multinode...

etc/kayobe/ansible/wazuh-manager.yml
```diff
@@ -53,6 +42,7 @@
           when: custom_sca_policies.files | length > 0
 
         - name: Add custom policy definition(s) to the shared Agent config
+          becomd: yes
```
markgoddard: Spelling

etc/kayobe/ansible/wazuh-manager.yml
```diff
@@ -110,18 +101,7 @@
         - Restart wazuh-dashboard
 
     - name: Perform health check against filebeat
+      become: yes
       command: filebeat test output
       changed_when: false
-      become: true
       retries: 2
-
-  handlers:
```
markgoddard, line 118: These handlers are still used

etc/kayobe/ansible/wazuh-manager.yml
markgoddard: It's definitely worth testing. If you want to merge this soon it might be better to revert to using become and remove it in a separate PR. That would be cleaner than conflating these things anyway.

markgoddard (changes requested):

MoteHue:
> revert to using become and remove it in a separate PR

I'll do this, it would be good to get the actual fix in sooner

etc/kayobe/ansible/wazuh-manager.yml
MoteHue: Added back

etc/kayobe/ansible/wazuh-manager.yml
MoteHue: Removed

etc/kayobe/ansible/wazuh-manager.yml
MoteHue: Done

markgoddard (approved):

PR #719, opened 2023-10-20
Apply centos build customisations on rocky also

mnasiadka:


priteau:
What about also renaming the variable to `kolla_build_customizations_redhat`?

Alex-Welsh:
> What about also renaming the variable to `kolla_build_customizations_redhat`?

`kolla_build_customizations_rhel` would probably match the variable naming in the rest of the file best

mnasiadka:
> > What about also renaming the variable to `kolla_build_customizations_redhat`?
> 
> `kolla_build_customizations_rhel` would probably match the variable naming in the rest of the file best

I would go for kolla_build_customizations_rpm or _el - so which one?

priteau:
> > > What about also renaming the variable to `kolla_build_customizations_redhat`?
> > 
> > 
> > `kolla_build_customizations_rhel` would probably match the variable naming in the rest of the file best
> 
> I would go for kolla_build_customizations_rpm or _el - so which one?

`rpm` would make sense if the other variable was `deb` instead of `ubuntu`. We use a mix of `rhel` and `el`, I slightly prefer the former but feel free to pick `el` if you prefer.

priteau (approved):

PR #725, opened 2023-10-30
Fixes for Barbican with Vault plugin

MoteHue:
Currently gives the error below, as the default field is not being applied to the varriable.

```
TASK [Write barbican Approle ID to file if requested] *************************************************************************
Monday 30 October 2023  09:57:48 +0000 (0:00:00.075)       0:00:07.517 ********
fatal: [controller01 -> localhost]: FAILED! =>
  msg: |-
    The conditional check 'stackhpc_write_barbican_role_id_to_file | bool | default(false)' failed. The error was: error while evaluating conditional (stackhpc_write_barbican_role_id_to_file | bool | default(false)): 'stackhpc_write_barbican_role_id_to_file' is undefined

    The error appears to be in '/var/lib/home/stackhpc/mattc/src/kayobe-config/etc/kayobe/ansible/vault-deploy-barbican.yml': line 80, column 7, but may
    be elsewhere in the file depending on the exact syntax problem.

    The offending line appears to be:

        - name: Write barbican Approle ID to file if requested
          ^ here
```

### DIFF
--- a/doc/source/configuration/vault.rst
+++ b/doc/source/configuration/vault.rst
@@ -296,6 +296,7 @@ Configure Barbican
       [vault_plugin]
       vault_url = https://{{ kolla_internal_vip_address }}:8200
       use_ssl = True
+      ssl_ca_crt_file = /etc/ssl/certs/ca-bundle.trust.crt
       approle_role_id = {{ secrets_barbican_approle_role_id }}
       approle_secret_id = {{ secrets_barbican_approle_secret_id }}
       kv_mountpoint = barbican
--- a/doc/source/configuration/vault.rst
+++ b/doc/source/configuration/vault.rst
@@ -296,7 +296,7 @@ Configure Barbican
       [vault_plugin]
       vault_url = https://{{ kolla_internal_vip_address }}:8200
       use_ssl = True
-      ssl_ca_crt_file = /etc/ssl/certs/ca-bundle.trust.crt
+      ssl_ca_crt_file = {% raw %}{{ openstack_cacert }}{% endraw %}
       approle_role_id = {{ secrets_barbican_approle_role_id }}
       approle_secret_id = {{ secrets_barbican_approle_secret_id }}
       kv_mountpoint = barbican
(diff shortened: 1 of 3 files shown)


Alex-Welsh (dismissed):

MoteHue:
Also included another fix for the SSL with Barbican Vault Plugin. Needed to specify the cert path

doc/source/configuration/vault.rst
```diff
@@ -296,6 +296,7 @@ Configure Barbican
       [vault_plugin]
       vault_url = https://{{ kolla_internal_vip_address }}:8200
       use_ssl = True
+      ssl_ca_crt_file = /etc/ssl/certs/ca-bundle.trust.crt
```
markgoddard: It would be better to reference `openstack_cacert` since this is distro-specific. You'll need to wrap in `raw` since its a Kolla variable.

etc/kayobe/environments/ci-multinode/kolla/config/barbican.conf
```diff
@@ -7,6 +7,7 @@ enabled_secretstore_plugins=vault_plugin
 [vault_plugin]
 vault_url = https://{{ kolla_internal_vip_address }}:8200
 use_ssl = True
+ssl_ca_crt_file = /etc/ssl/certs/ca-bundle.trust.crt
```
markgoddard: As above

MoteHue:
I've also proposed another fix. The conditional check for the Barbican AppRole secret ID being present in vault would not work. Firstly, because the "when" statement was matching on it being present, when it should have been when it is not present. 

However, the module ``hashivault_approle_role_secret_list`` also does not return a useful output. It lists the vault-uuids for the secrets, rather than the actual approle id which we would want to match for.

I've instead changed to use a different module, which can check for the presence of the secret specifically.

doc/source/configuration/vault.rst
MoteHue: Done :)

etc/kayobe/environments/ci-multinode/kolla/config/barbican.conf
MoteHue: Done :)

markgoddard (approved):

PR #708, opened 2023-10-12
CI: split PR workflows, add paths

markgoddard:
- CI: Split .github/workflows/stackhpc-pull-request.yml into tox and aio workflows
- CI: Add paths to stackhpc-pull-request-tox-aio.yml workflow

This should reduce our CI overheads by only running aios when necessary. It will also make it easier to merge some changes.

PR #726, opened 2023-10-30
CI: Skip aio jobs based on changed files

markgoddard:
We don't always want to run all aio jobs - it can be wasteful of CI
resources and slows down merging changes.

This change uses the dorny/paths-filter action to detect changed files
and skip jobs as appropriate.

Note that we can't use the workflow-level paths attribute since this
would skip the workflow entirely, and would prevent us from making the
aio jobs required to pass (a skip counts as a pass).


markgoddard:
This is a tidied up version of https://github.com/stackhpc/stackhpc-kayobe-config/pull/708

.github/workflows/stackhpc-pull-request.yml
```diff
@@ -34,38 +57,55 @@ jobs:
...
+  # when the parent jobs completed successfully or were skipped. We pass an
+  # 'if' argument to the called workflow to allow running it conditionally.
+
   build-kayobe-image:
     name: Build Kayobe Image
+    needs:
+      - check-changes
     uses: ./.github/workflows/stackhpc-build-kayobe-image.yml
+    with:
+      if: ${{ needs.check-changes.outputs.aio == 'true' }}
```
Alex-Welsh, line 74: is `== 'true'` required here?

markgoddard: Unfortunately I believe it is - it's a string rather than a proper boolean.

Alex-Welsh (approved):

PR #727, opened 2023-10-30
Disable custom apt config for non-overcloud hosts

Alex-Welsh:
non-overcloud hosts e.g. the seed hypervisor are configured before the seed and its services are deployed. This means that they will look for the local package repo mirrors before they exist.

I'm marking this as a draft for now since my AUFN hasn't fully deployed. Currently unsure whether this affects Yoga and whether I should target it there. I need to do a yoga-zed upgrade test in an aufn at some point so I can test then if needed

### DIFF
--- a/etc/kayobe/apt.yml
+++ b/etc/kayobe/apt.yml
@@ -62,12 +62,18 @@ stackhpc_apt_repositories:
     signed_by: docker.asc
     architecture: amd64
 
-apt_repositories: "{{ stackhpc_apt_repositories }}"
+# Do not replace apt configuration for non-overcloud hosts. This can result in
+# errors if apt reconfiguration is performed before local repository mirrors
+# are deployed.
+apt_repositories: "{{ stackhpc_apt_repositories if 'overcloud' in group_names else [] }}"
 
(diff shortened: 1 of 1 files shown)


Alex-Welsh:
Currently blocked by https://github.com/stackhpc/stackhpc-kayobe-config/pull/728

markgoddard:
> non-overcloud hosts e.g. the seed hypervisor are configured before the seed and its services are deployed. This means that they will look for the local package repo mirrors before they exist.
> 
> I'm marking this as a draft for now since my AUFN hasn't fully deployed. Currently unsure whether this affects Yoga and whether I should target it there. I need to do a yoga-zed upgrade test in an aufn at some point so I can test then if needed

It does affect yoga and should go there first.

markgoddard (commented):
Change looks sensible. Could you add a reno?

Alex-Welsh:
> Change looks sensible. Could you add a reno?

Will do, but I'll save some faff and just open a new PR against yoga

Alex-Welsh:
New PR: https://github.com/stackhpc/stackhpc-kayobe-config/pull/729

PR #729, opened 2023-11-01
Disable custom apt config for non-overcloud hosts

Alex-Welsh:
non-overcloud hosts e.g. the seed hypervisor are configured before the seed and its services are deployed. This means that they will look for the local package repo mirrors before they exist.

### DIFF
--- a/etc/kayobe/apt.yml
+++ b/etc/kayobe/apt.yml
@@ -62,12 +62,18 @@ stackhpc_apt_repositories:
     signed_by: docker.asc
     architecture: amd64
 
-apt_repositories: "{{ stackhpc_apt_repositories }}"
+# Do not replace apt configuration for non-overcloud hosts. This can result in
+# errors if apt reconfiguration is performed before local repository mirrors
+# are deployed.
+apt_repositories: "{{ stackhpc_apt_repositories if 'overcloud' in group_names else [] }}"
 
 # Whether to disable repositories in /etc/apt/sources.list. This may be used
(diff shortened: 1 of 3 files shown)


markgoddard (commented):
I'm looking at this in `etc/kayobe/kolla/globals.yml` and thinking the Debian part is wrong (and should be removed):

```
# To work around issue of trying to install docker from
# empty pulp server, use upstream docker dnf repo on
# non-overcloud hosts
enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names or ansible_facts.os_family == 'Debian' }}{% endraw %}"
```

We include a docker repo in `etc/kayobe/apt.yml` for overcloud hosts, so don't need Kolla Ansible to install one.

Alex-Welsh:
> I'm looking at this in `etc/kayobe/kolla/globals.yml` and thinking the Debian part is wrong (and should be removed):
> 
> ```
> # To work around issue of trying to install docker from
> # empty pulp server, use upstream docker dnf repo on
> # non-overcloud hosts
> enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names or ansible_facts.os_family == 'Debian' }}{% endraw %}"
> ```
> 
> We include a docker repo in `etc/kayobe/apt.yml` for overcloud hosts, so don't need Kolla Ansible to install one.

Good point, I guess that just got missed when we enabled apt repo mirroring again

markgoddard (approved):

PR #732, opened 2023-11-01
Fix seed container image build

Alex-Welsh:
The argument list was missing -e which meant the first argument was interpreted as image regexes.


markgoddard (approved):
How many times has this been fixed now?

Alex-Welsh:
> How many times has this been fixed now?

Too many

PR #733, opened 2023-11-01
Ubuntu: add base repos to Kolla container image earlier

markgoddard:
Previously when building Kolla container images for Ubuntu it was common to get a dependency error during apt package installation, such as:

  The following packages have unmet dependencies:
    libc6-dev : Depends: libc6 (= 2.35-0ubuntu3.1) but 2.35-0ubuntu3.4 is to be installed

This is caused by the packages in the Ubuntu base image being newer than those in the release train repository snapshot.

This change fixes the issue by switching to the release train repositories earlier in the build process, before the first apt command runs.

It's possible that we may also need to pin the Ubuntu base container image, but that would be another thing to maintain, so let's try without for now since this approach is necessary and may be sufficient.

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -277,13 +277,19 @@ base_centos_repo_overrides_post_yum_rocky_list: "{{ stackhpc_rocky_9_repos + sta
 base_centos_repo_overrides_post_yum_list: "{{ base_centos_repo_overrides_post_yum_centos_list if kolla_base_distro == 'centos' else base_centos_repo_overrides_post_yum_rocky_list }}"
 stackhpc_yum_repos: "{{ stackhpc_centos_stream_repos if kolla_base_distro == 'centos' else stackhpc_rocky_9_repos }}"
 
-# List of repositories for Ubuntu focal.
-stackhpc_ubuntu_focal_repos:
+# List of base repositories for Ubuntu Focal.
+stackhpc_ubuntu_focal_base_repos:
   - "deb {{ stackhpc_repo_ubuntu_focal_url }} focal main universe"
   - "deb {{ stackhpc_repo_ubuntu_focal_url }} focal-updates main universe"
   - "deb {{ stackhpc_repo_ubuntu_focal_url }} focal-backports main universe"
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla.yml
```diff
@@ -320,9 +333,13 @@ kolla_build_blocks:
...
     {% endfor %}
     {% endif %}
+  # With the UCA keyring installed we can now add all repos.
   base_ubuntu_package_sources_list: |
+    {% if stackhpc_kolla_clean_up_repo_mirrors | bool %}
+    COPY sources.list.ubuntu /etc/apt/sources.list.backup
+    {% endif %}
     RUN \
-        mv /etc/apt/sources.list /etc/apt/sources.list.backup && \
+        rm /etc/apt/sources.list && \
```
Alex-Welsh, line 342: :disappointed:

Alex-Welsh (commented):
I like that we're going to have this resolved for our images, but it's possibly worth noting somewhere (docs maybe) that the problem will remain for places that aren't using our images & bifrost provisioning. I know some places e.g. eidf, this has come up before but I don't know exactly how common it is

markgoddard:
Tested and working for glance-api and nova-compute images.

markgoddard:
> I like that we're going to have this resolved for our images, but it's possibly worth noting somewhere (docs maybe) that the problem will remain for places that aren't using our images & bifrost provisioning. I know some places e.g. eidf, this has come up before but I don't know exactly how common it is

This is for the container images rather than the overcloud host images. It's only a problem if you use repository snapshots rather than upstream Ubuntu.

etc/kayobe/kolla.yml
markgoddard: There's definitely some magic going on here...

In base_header I'm replacing sources.list with a file containing the base Ubuntu repos. This is because we don't yet have the necessary keyring to use the UCA repo.

In base_ubuntu_package_sources_list we have the UCA keyring, so we can remove sources.list and build it up again including UCA.

Regarding the backup, Kolla provides the `docker/base/sources.list.ubuntu` file which is used by default on Ubuntu. I'm using it as the backup sources.list file on L339 that will be copied back into place at the end of the build in the `footer` section.

jovial (approved):

PR #734, opened 2023-11-01
Avoid facts for purge-command-not-found.yml

Alex-Welsh:
Gather facts was false but facts are used in the playbook.

### DIFF
--- a/etc/kayobe/ansible/purge-command-not-found.yml
+++ b/etc/kayobe/ansible/purge-command-not-found.yml
@@ -30,4 +30,4 @@
       become: true
       failed_when: false
       ignore_unreachable: true
-      when: ansible_facts.os_family == "Debian"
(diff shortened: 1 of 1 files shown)


etc/kayobe/ansible/purge-command-not-found.yml
```diff
@@ -11,7 +11,7 @@
 # by implementing this playbook separately for each group.
 - name: Purge command-not-found package
   hosts: seed-hypervisor:seed:overcloud:infra-vms
-  gather_facts: false
+  gather_facts: true
```
markgoddard: Gathering facts as the bootstrap user can be a bad idea if using fact caching because we later reference facts about the user.

Can we change to use `os_distribution` instead?

markgoddard (approved):

PR #747, opened 2023-11-06
add new var needed for pulp initial deployment

bbezak:


markgoddard (changes requested):
The way I was hoping this would work is that we'd set `seed_deploy_containers_registry_attempt_login: true` in seed.yml, then add a login task to the pulp post.yml tasks file.

PR #748, opened 2023-11-06
Playbook fix

grzegorzkoper:
Fixing the playbook as it was triggering docker with:
"The input device is not a TTY".

Fixing ansible-lint errors.


markgoddard:
We don't generally cherry-pick changes in SKC, instead we merge old branches into new ones. Let me do that now.

PR #754, opened 2023-11-07
Pin pulp.squeezer to 0.0.13

markgoddard:
From 0.0.14 pulp.squeezer requires the pulp_glue Python library to be installed.

Pin to 0.0.13 until we have a solution for installing pulp_glue.


jackhodgkiss (approved):
LGTM

PR #760, opened 2023-11-08
yoga rl9 in place docs

markgoddard:
Add in-place upgrade to RL9 migration

### DIFF
--- a/doc/source/operations/rocky-linux-9.rst
+++ b/doc/source/operations/rocky-linux-9.rst
@@ -617,8 +617,7 @@ Potential issues
 Full procedure
 --------------
 
-- It's good to inspect existing DNF packages and determine whether they are
-  really required.
+- Inspect existing DNF packages and determine whether they are really required.
 
 - Use the `migrate2rocky.sh
   <https://raw.githubusercontent.com/rocky-linux/rocky-tools/main/migrate2rocky/migrate2rocky.sh>`__
@@ -630,7 +629,9 @@ Full procedure
 
      sudo dnf module disable "*"
 
-- Migrate to NetworkManager:
+- Migrate to NetworkManager. This can be done using a manual process or with Kayobe.
+
+  The manual process is as follows:
 
   - Ensure that all network interfaces are managed by Network Manager:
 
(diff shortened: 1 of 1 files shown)


doc/source/operations/rocky-linux-9.rst
```diff
@@ -599,7 +599,84 @@ Wazuh manager
...
+       sudo systemctl enable NetworkManager
+       sudo systemctl start NetworkManager
+
+  - Migrate Ethernet connections to native NetworkManager configuration:
+
+    .. code:: console
+
+       sudo nmcli connection migrate
+
+  - Manually migrate non-Ethernet (bonds, bridges & VLAN subinterfaces) network interfaces to native NetworkManager.
```
markgoddard, line 655: If there's anything better/more explicit we can say here, that would be nice.

jovial: Is there no way to set `interfaces_use_nmconnection: true` and get kayobe to generate these files?

Edit: Just https://github.com/michaelrigart/ansible-role-interfaces/blob/master/vars/redhat-8.yml#L8 suggests that it might work

jovial (dismissed):

doc/source/operations/rocky-linux-9.rst
cityofships: @jovial Of course - just there's a problem if your main admin interface is on bond/bridge/vlan interface. The rest can definitely be done via Kayobe and should be smooth.

markgoddard: I've added a placeholder procedure for a Kayobe-based migration.

doc/source/operations/rocky-linux-9.rst
```diff
@@ -669,14 +680,36 @@ Full procedure
...
+    .. code:: console
 
-     sudo dnf install -y https://download.rockylinux.org/pub/rocky/9/BaseOS/x86_64/os/Packages/r/rocky-gpg-keys-9.2-1.6.el9.noarch.rpm \
-                         https://download.rockylinux.org/pub/rocky/9/BaseOS/x86_64/os/Packages/r/rocky-release-9.2-1.6.el9.noarch.rpm \
-                         https://download.rockylinux.org/pub/rocky/9/BaseOS/x86_64/os/Packages/r/rocky-repos-9.2-1.6.el9.noarch.rpm
-     sudo rm -rf /usr/share/redhat-logos
-     sudo dnf --releasever=9 --allowerasing --setopt=deltarpm=false distro-sync -y
-     sudo rpm --rebuilddb
-     sudo rpm -qa | grep el8 | xargs dnf remove
+       sudo rpm -qa | grep el8 | xargs dnf remove
```
cityofships: In fact `xargs` here will  fail with dnf interactively asking for confirmation. Original line was `xargs dnf -y remove` but this requires extra caution & prior inspection of the packages list.

markgoddard: Possibly the grep pattern could be better to match the `.el8.` bit, but I've reworked it to write to a file

jovial (approved):
LGTM

PR #764, opened 2023-11-09
docs: a couple of bits for the RL9 migration

markgoddard:


doc/source/operations/rocky-linux-9.rst
```diff
@@ -108,6 +110,7 @@ Elasticsearch/Kibana should be migrated to OpenSearch.
   docs <https://docs.openstack.org/kolla-ansible/yoga/reference/logging-and-monitoring/central-logging-guide-opensearch.html#migration>`__
 - If necessary, take a backup of the Elasticsearch data.
 - Ensure ``kolla_enable_elasticsearch`` is unset in ``etc/kayobe/kolla.yml``
+- If you have a custom Kolla Ansible inventory, ensure that it contains the ``opensearch`` and ``opensearch-dashboards`` groups. Otherwise, sync with the inventory in Kayobe.
```
mnasiadka, line 113: Wondering if we shouldn't think of a check upstream that would fail if you don't have required groups in your inventory

mnasiadka (approved):

doc/source/operations/rocky-linux-9.rst
markgoddard: We do have a precheck that should do it:

https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/service-precheck/tasks/main.yml

PR #775, opened 2023-11-13
Check if repos backup exists in kolla footer

MoteHue:
When a child service is build, such as neutron-infoblox-ipam-agent being a child of neutron-server, the fotoer is ran twice.

This means it fails as the repo backup has already been moved.

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -353,17 +353,17 @@ kolla_build_blocks:
     {% if kolla_base_distro in ['centos', 'rocky'] %}
     RUN \
         if [ -f /etc/yum.repos.d.backup/repos.tar.gz ]; then \
-          tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d
-          rm -rf /etc/yum.repos.d.backup/
+          tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d && \
+          rm -rf /etc/yum.repos.d.backup/ \
         fi && \
         if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/yum.repos.d; then \
           echo "Found repository mirror in Yum repositories"; \
           exit 1; \
-        fi && \
+        fi
     {% else %}
     RUN \
         if [ -f /etc/apt/sources.list.backup ]; then \
-          mv /etc/apt/sources.list.backup /etc/apt/sources.list
+          mv /etc/apt/sources.list.backup /etc/apt/sources.list \
         fi && \
         if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/apt/sources.list; then \
           echo "Found repository mirror in APT repositories"; \
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -352,18 +352,22 @@ kolla_build_blocks:
     {% if stackhpc_kolla_clean_up_repo_mirrors | bool %}
     {% if kolla_base_distro in ['centos', 'rocky'] %}
     RUN \
-        tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d && \
-        if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/yum.repos.d; then \
-           echo "Found repository mirror in Yum repositories"; \
-           exit 1; \
-        fi && \
-        rm -rf /etc/yum.repos.d.backup/
+        if [ -f /etc/yum.repos.d.backup/repos.tar.gz ]; then \
+          tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d && \
+          if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/yum.repos.d; then \
+            echo "Found repository mirror in Yum repositories"; \
+            exit 1; \
+          fi && \
+          rm -rf /etc/yum.repos.d.backup/
+        fi
     {% else %}
     RUN \
-        mv /etc/apt/sources.list.backup /etc/apt/sources.list && \
-        if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/apt/sources.list; then \
-           echo "Found repository mirror in APT repositories"; \
-           exit 1; \
+        if [ -f /etc/apt/sources.list.backup ]; then \
+          mv /etc/apt/sources.list.backup /etc/apt/sources.list && \
+          if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/apt/sources.list; then \
+            echo "Found repository mirror in APT repositories"; \
+            exit 1; \
+          fi
         fi
     {% endif %}
     {% endif %}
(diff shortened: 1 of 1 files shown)


MoteHue:
Example failure here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/6826625196/job/18568857784

etc/kayobe/kolla.yml
```diff
@@ -352,18 +352,22 @@ kolla_build_blocks:
...
+        fi
     {% else %}
     RUN \
-        mv /etc/apt/sources.list.backup /etc/apt/sources.list && \
-        if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/apt/sources.list; then \
-           echo "Found repository mirror in APT repositories"; \
-           exit 1; \
+        if [ -f /etc/apt/sources.list.backup ]; then \
+          mv /etc/apt/sources.list.backup /etc/apt/sources.list && \
+          if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/apt/sources.list; then \
```
jovial: Possibly move the check outside the conditional so that we still fail if the stackhpc mirrors are configured?

jovial:
I guess this also means we wouldn't use pulp in the images that build on an image where the footer has already run?

etc/kayobe/kolla.yml
MoteHue: It should still fail when the parent container image is built, as the footer would have ran here before.

jovial: Sorry, I meant this check:
```
          if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/yum.repos.d; then \
            echo "Found repository mirror in Yum repositories"; \
            exit 1; \
          fi && \
```

MoteHue: Done

etc/kayobe/kolla.yml
```diff
@@ -352,18 +352,22 @@ kolla_build_blocks:
...
+        if [ -f /etc/yum.repos.d.backup/repos.tar.gz ]; then \
+          tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d
+        fi && \
         if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/yum.repos.d; then \
-           echo "Found repository mirror in Yum repositories"; \
-           exit 1; \
+          echo "Found repository mirror in Yum repositories"; \
+          exit 1; \
         fi && \
         rm -rf /etc/yum.repos.d.backup/
```
jovial: probably still need this is the `if [ -f /etc/yum.repos.d.backup/repos.tar.gz ]; then \` block

MoteHue: I think you meant about the ``rm rf /etc/yum.repos.d.backup/`` here? I've moved it into the block, although given ``rm`` still runs even if the file doesn't exist, it probably would have been fine :wink:

jovial (dismissed):

etc/kayobe/kolla.yml
```diff
@@ -352,18 +352,22 @@ kolla_build_blocks:
     {% if stackhpc_kolla_clean_up_repo_mirrors | bool %}
     {% if kolla_base_distro in ['centos', 'rocky'] %}
     RUN \
-        tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d && \
+        if [ -f /etc/yum.repos.d.backup/repos.tar.gz ]; then \
+          tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d
+          rm -rf /etc/yum.repos.d.backup/
```
markgoddard:
```suggestion
          tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d && \
          rm -rf /etc/yum.repos.d.backup/ \
```

etc/kayobe/kolla.yml
```diff
@@ -352,18 +352,22 @@ kolla_build_blocks:
...
+        fi && \
         if grep -r '{{ stackhpc_repo_mirror_url }}' /etc/yum.repos.d; then \
-           echo "Found repository mirror in Yum repositories"; \
-           exit 1; \
+          echo "Found repository mirror in Yum repositories"; \
+          exit 1; \
         fi && \
```
markgoddard:
```suggestion
        fi
```

etc/kayobe/kolla.yml
```diff
@@ -352,18 +352,22 @@ kolla_build_blocks:
...
-        rm -rf /etc/yum.repos.d.backup/
     {% else %}
     RUN \
-        mv /etc/apt/sources.list.backup /etc/apt/sources.list && \
+        if [ -f /etc/apt/sources.list.backup ]; then \
+          mv /etc/apt/sources.list.backup /etc/apt/sources.list
```
markgoddard:
```suggestion
          mv /etc/apt/sources.list.backup /etc/apt/sources.list \
```

markgoddard (changes requested):

etc/kayobe/kolla.yml
```diff
@@ -353,7 +353,7 @@ kolla_build_blocks:
     {% if kolla_base_distro in ['centos', 'rocky'] %}
     RUN \
         if [ -f /etc/yum.repos.d.backup/repos.tar.gz ]; then \
-          tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d && \
```
markgoddard: We would hide a non-zero return code without &&. Is that expected?

MoteHue: I was seeing this error, which I thought might be being caused by using the && here. (I'll admit my bash scripting knowledge is patchy)
```
INFO:kolla.common.utils.neutron-server:Step 7/8 : RUN     if [ -f /etc/yum.repos.d.backup/repos.tar.gz ]; then       tar -xzf /etc/yum.repos.d.backup/repos.tar.gz -C /etc/yum.repos.d &&       rm -rf /etc/yum.repos.d.backup/     fi &&     if grep -r 'http://pulp-server.internal.sms-cloud:8080/' /etc/yum.repos.d; then       echo "Found repository mirror in Yum repositories";       exit 1;     fi
    INFO:kolla.common.utils.neutron-server: ---> Running in f9498830e462
    INFO:kolla.common.utils.neutron-server:[91m/bin/sh: -c: line 2: syntax error: unexpected end of file

MoteHue: I've restored the &&, I suspect the error is from missing semicolons due to this being pushed into an inline bash command. Will be able to test again once SMS is back

markgoddard (commented):
LGTM, please update once tested.

MoteHue:
> LGTM, please update once tested.

Tested and the builds got much further than before, so I believe this has fixed things. (The test build was cancelled after two hours due to not being needed).

Could hold off for a complete build test if that's preferable

markgoddard:
> > LGTM, please update once tested.
> 
> Tested and the builds got much further than before, so I believe this has fixed things. (The test build was cancelled after two hours due to not being needed).
> 
> Could hold off for a complete build test if that's preferable

Did it build the neutron infoblox image that was previously broken?

MoteHue:
> > > LGTM, please update once tested.
> > 
> > 
> > Tested and the builds got much further than before, so I believe this has fixed things. (The test build was cancelled after two hours due to not being needed).
> > Could hold off for a complete build test if that's preferable
> 
> Did it build the neutron infoblox image that was previously broken?

Yes, had a successful build here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/6863952531/job/18664728341

markgoddard (approved):

PR #787, opened 2023-11-15
CI: Don't fail fast on container image build job failure

markgoddard:
Previously if one of the container image build jobs (CS8, Ubuntu)
failed, the other would be cancelled. This is not necessarily helpful,
since the other job may complete successfully.

This change disables this fail fast behaviour.


MoteHue (approved):
This will be helpful, LGTM :)

PR #793, opened 2023-11-16
Configure SELinux in permissive mode on RL9 hosts

priteau:
This makes the host match the target configuration in Zed and avoids a
later reboot to change the SELinux mode.

Hosts that have SELinux disabled are rebooted only if
disable_selinux_do_reboot is changed to true.

### DIFF
--- a/etc/kayobe/inventory/group_vars/all/selinux
+++ b/etc/kayobe/inventory/group_vars/all/selinux
@@ -1,0 +1,9 @@
+---
+# Configure SELinux in permissive mode when configuring a Rocky Linux 9 host.
+selinux_state: "{{ 'permissive' if ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '9' else 'disabled' }}"
+
+# Do NOT reboot Rocky Linux 9 hosts to apply SELinux config changes. Operators
+# must opt-in by changing this variable when applying host configuration. This
+# is to avoid automatically rebooting hosts originally deployed with SELinux
+# disabled and which now need to be changed to permissive.
+disable_selinux_do_reboot: "{{ not (ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '9') }}"
--- a/releasenotes/notes/selinux-permissive-bb953d2cdcd7a545.yaml
+++ b/releasenotes/notes/selinux-permissive-bb953d2cdcd7a545.yaml
@@ -1,0 +1,7 @@
+---
+upgrade:
+  - |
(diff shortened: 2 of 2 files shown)


MoteHue (dismissed):
Would be nice to recommend this in the RL9 migration docs. Otherwise, LGMT :)

etc/kayobe/inventory/group_vars/overcloud/selinux
```diff
@@ -0,0 +1,7 @@
+---
+# Configure SELinux in permissive mode when configuring a Rocky Linux 9 host.
+selinux_state: "{{ 'permissive' if ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '9' else 'disabled' }}"
+
+# Do NOT reboot to apply SELinux config changes. Operators must opt-in by
+# changing this variable when applying host configuration.
+disable_selinux_do_reboot: False
```
markgoddard: I would prefer to change this only on RL9 hosts to avoid changing behaviour elsewhere.

etc/kayobe/inventory/group_vars/overcloud/selinux
markgoddard, line 1: Should this only apply to overcloud hosts, or would it be useful to apply everywhere?

priteau: I have moved it to `all` group variables which will apply to seed hypervisor, seed host and infra VMs too.

etc/kayobe/inventory/group_vars/all/selinux
```diff
@@ -0,0 +1,9 @@
+---
+# Configure SELinux in permissive mode when configuring a Rocky Linux 9 host.
+selinux_state: "{{ 'permissive' if ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '9' else 'disabled' }}"
```
Alex-Welsh, line 3: Shorter but maybe less clear? What do you think?
```suggestion
selinux_state: "{{ 'disabled' if disable_selinux_do_reboot else 'permissive' }}"
```

priteau: Thanks but I am not keen on this suggestion, as you have to understand the logic of the second variable to know what the first variable evaluates to.

priteau: Thanks but I am not keen on this suggestion, as you have to understand the logic of the second variable to know what the first variable evaluates to.

markgoddard (approved):

PR #794, opened 2023-11-17
Add more services to the rabbitmq-reset playbook

priteau:
This change adds services that we often deploy: Barbican, Blazar, CloudKitty, Designate, Manila and Octavia.

### DIFF
--- a/etc/kayobe/ansible/rabbitmq-reset.yml
+++ b/etc/kayobe/ansible/rabbitmq-reset.yml
@@ -1,6 +1,6 @@
 ---
 # Reset a broken RabbitMQ cluster.
-# Also restarts OpenStack services which may be broken.
+# Also restarts all OpenStack services using RabbitMQ.
 
 - name: Reset RabbitMQ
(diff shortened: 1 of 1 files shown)


etc/kayobe/ansible/rabbitmq-reset.yml
```diff
@@ -69,4 +69,4 @@
     # The following services can have problems if the cluster gets broken.
```
MoteHue: Nit: Given this is now a list of all the services we deploy which use RabbitMQ, the comment here and at the start of the file could be update to reflect this. (Right now it suggests only some of them imo)

markgoddard (dismissed):

etc/kayobe/ansible/rabbitmq-reset.yml
priteau: I have updated the comments.

markgoddard (approved):

MoteHue (approved):

PR #697, opened 2023-10-09
Various os_capacity fixes

assumptionsandg:
- Various typo fixes.
- Renamed `os_exporter` to `os_capacity`.
- Renamed `70-oscapacity.yml.j2` to `70-oscapacity.yml` to ensure correct templating.
- Added `70-oscapacity.yml` to .yamlint ignore list.
- Conditional flag `stackhpc_enable_os_capacity` added to prevent HAProxy config and Prometheus scrape targets being copied in before manual deployment.
- Conditional flag `stackhpc_os_capacity_openstack_verify` added where certificate verification can be overridden.
- Fixed interval errors

PR #795, opened 2023-11-17
Rename 70-oscapacity.yml.j2 to 70-oscapacity.yml

jovial:
This wasn't being picked up by kolla-ansible since the extension was not .yml


jovial:
Done in: https://github.com/stackhpc/stackhpc-kayobe-config/pull/697

PR #796, opened 2023-11-17
Add rekey playbook

Alex-Welsh:
Adding a playbook to rotate ssh keys in a cloud. 
Given a list of users (default: kolla, stack) the playbook will remove the existing ssh key from each user's `authorized_keys` file and replace it with a new one.


etc/kayobe/ansible/rekey-hosts.yml
```diff
@@ -0,0 +1,75 @@
+---
+- name: Rekey hosts
+  hosts: overcloud,seed,seed-hypervisor,infra-vms
+  gather_facts: false
+  vars:
+    new_key_type: ed25519
```
markgoddard: Nice idea, but will require a change to the `ssh_key_type` variable to make it work.

etc/kayobe/ansible/rekey-hosts.yml
```diff
@@ -0,0 +1,75 @@
...
+      delegate_to: localhost
+      run_once: true
+
+    - name: Locally promote new key (private)
+      command: "mv ~/.ssh/id_{{ new_key_type }}_new ~/.ssh/id_{{ new_key_type }}"
+      delegate_to: localhost
+      run_once: true
+
+    - name: Locally promote new key (public)
+      command: " mv ~/.ssh/id_{{ new_key_type }}_new.pub ~/.ssh/id_{{ new_key_type }}.pub"
```
markgoddard:
```suggestion
      command: "mv ~/.ssh/id_{{ new_key_type }}_new.pub ~/.ssh/id_{{ new_key_type }}.pub"
```

etc/kayobe/ansible/rekey-hosts.yml
```diff
@@ -0,0 +1,75 @@
...
+  hosts: overcloud,seed,seed-hypervisor,infra-vms
+  gather_facts: false
+  vars:
+    new_key_type: ed25519
+    ansible_ssh_common_args: "-o StrictHostKeyChecking=no"
+    rekey_users:
+      - stack
+      - kolla
+    existing_key_path: "~/.ssh/id_rsa"
+    rekey_remove_existing_key: true
```
markgoddard: I would set this to false by default - you don't want to get locked out. You could run again with it set to true once access is confirmed.

etc/kayobe/ansible/rekey-hosts.yml
markgoddard: We could assert on these two being equal, or just use `ssh_key_type`

markgoddard (changes requested):

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: `ssh.yml` variables actually solve a few different problems. I hadn't realised they existed. I'm working on rewriting the playbook to use those existing variables. Marking as draft for now as it's untested.

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: Done

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: Done

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: Tested, Done

Alex-Welsh:
I've changed my mind on this one, I'll revert back to the old style and just set `ssh_*_key_path` as the defaults. will update soon<sup>tm</sup>

Alex-Welsh:
I think I'm done fiddling with the format now. Again, I'm marking as draft until I get the chance to test it again.

Now by default it will rotate the keys used by kayobe/kolla-ansible but can also be used more generally by overriding the `existing/new_private/public_key_path` variables

Alex-Welsh:
Again, tested & working now

etc/kayobe/ansible/rekey-hosts.yml
```diff
@@ -0,0 +1,97 @@
+---
+# Playbook to rotate SSH keys across the cloud. By default it will rotate the
+# standard keys used by kayobe/kolla-ansible, but it can be configured for any
+# keys.
+
+- name: Rekey hosts
+  hosts: overcloud,seed,seed-hypervisor,infra-vms
+  gather_facts: false
+  vars:
+    ansible_ssh_common_args: "-o StrictHostKeyChecking=no"
```
markgoddard: Why do we need this?

etc/kayobe/ansible/rekey-hosts.yml
```diff
@@ -0,0 +1,97 @@
...
+      ansible.builtin.fail:
+        msg: "No existing public key file found. Check existing_public_key_path is set correctly."
+      when:
+        - not stat_result.stat.exists
+      delegate_to: localhost
+      run_once: true
+
+    - name: Generate a new SSH key
+      community.crypto.openssh_keypair:
+        path: "{{ existing_private_key_path }}_new"
```
markgoddard, line 56: Should this use new_private_key_path?

etc/kayobe/ansible/rekey-hosts.yml
```diff
@@ -0,0 +1,97 @@
...
+    - name: Remove old key from hosts
+      vars:
+        lookup_path: "{{ existing_public_key_path }}_old"
+      ansible.posix.authorized_key:
+        user: "{{ item }}"
+        state: absent
+        key: "{{ lookup('file', lookup_path) }}"
+      loop: "{{ rekey_users }}"
+      become: true
+      when: rekey_remove_existing_key | bool
```
markgoddard, line 117: Thinking about the flow here, should it be something like:

* run rekey-hosts.yml
* test access with new key
* remove old key from hosts

If we were to run this playbook again with `rekey_remove_existing_key` set to true, wouldn't it generate another new key?

Also, should we move the old key out of `.ssh` entirely to avoid it being picked up?

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: The key gets moved to `"{{ new_private_key_path }}"` later on. the only criteria for the path in that line is that it can't conflict with the existing key.

If it's purely `path: "{{ new_private_key_path }}"` then the new path cannot equal the existing path. 

`"{{ new_private_key_path }}_new"` would work in almost every scenario, except when 
`"{{ new_private_key_path }}_new" == "{{ existing_private_key_path }}"` (which is an incredibly niche situation but still)

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: I'll retest this bit soon. Unsure it it was required before I hadn't connected to the other hosts before or it legitimately needs it because the key is changing

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: If the playbook works, you should be able to just run it with `rekey_remove_existing_key=true` in the first place :wink: 
I do understand your point though and I'll split it up into blocks so you can run one part at a time

markgoddard: You could just add a tag to the last task

etc/kayobe/ansible/rekey-hosts.yml
markgoddard: The main thing I was thinking is if you change the key type then the temporary name would look wrong

Alex-Welsh: I see what you mean, however the user shouldn't ever see that temporary name, since it's immediately removed

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: I've added a block so I can check the file exists first

Alex-Welsh: Following your flow, you can now go:

1. `kayobe playbook run rekey-hosts.yml`
2. `ssh controller0`
3. `kayobe playbook run rekey-hosts.yml -t remove-key -e rekey_remove_existing_key=true`

etc/kayobe/ansible/rekey-hosts.yml
Alex-Welsh: It's unnecessary, removed

markgoddard (approved):

PR #798, opened 2023-11-23
feat: add pre-commit hooks

jackhodgkiss:
Add a simple playbook that can install [pre-commits](https://pre-commit.com/) and register a few simple hooks. This process is implemented as a hook that will run post control host bootstrap.

- check-yaml: perform basic yaml syntax linting
- end-of-file-fixer: identify and automatically fix missing newline
- trailing-whitespace: identify and automatically fix excessive white space
- ripsecrets: identify and prevent secrets from being committed to the branch

Note: this is currently configured as an opt-in feature and will only install and register hooks if `enable_pre_commit_hooks` is set to `true` in `install-pre-commit-hooks.yml`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,11 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/sirwart/ripsecrets
+    rev: v0.1.7
+    hooks:
+      - id: ripsecrets
--- a/doc/source/contributor/pre-commit.rst
+++ b/doc/source/contributor/pre-commit.rst
@@ -1,0 +1,38 @@
+================
+Pre-commit Hooks
+================
+
+StackHPC Kayobe configuration carries support for
+`pre-commit hooks <https://pre-commit.com/>`_ which simplify the use of git
+hooks enabling the identification and repairing of broken or poor code
+before committing.
+These hooks are designed to make working within SKC easier and less error prone.
+
+Currently the following hooks are provided:
+
+- ``check-yaml``: perform basic yaml syntax linting
+- ``end-of-file-fixer``: identify and automatically fix missing newline
+- ``trailing-whitespace``: identify and automatically fix excessive white space
+- ``ripsecrets``: identify and prevent secrets from being committed to the branch
+
+.. warning::
+   The hook ``ripsecrets`` is capable of preventing the accidental leaking of secrets
+   such as those found within `secrets.yml` or `passwords.yml`.
+   However if the secret is contained within a file on it's own and lacks a certain level
+   of entropy then the secret will not be identified as such as and maybe leaked as a result.
+
+Installation of `pre-commit` hooks is handled via the `install-pre-commit-hooks` playbook
+found within the Ansible directory.
+Either use `kayobe playbook run` or add the playbook as a hook within Kayobe config such as
+within `control-host-bootstrap/post.d`.
+Once done you should find `pre-commit` is available within the `kayobe` virtualenv.
+
+All that remains is the installation of the hooks themselves which can be accomplished either by
+running `pre-commit run` or using `git commit` when you have changes that need to be committed.
+This will trigger a brief installation process of the hooks which may take a few minutes.
+This a one time process and will not be required again unless new hooks are added or existing ones are updated.
+
+.. note::
+   Currently if you run ``pre-commit run --all-files`` it will make a series of changes to
+   release notes that lack new lines as well configuration files that ``check-yaml`` does not
+   approve of.
--- a/doc/source/contributor/pre-commit.rst
+++ b/doc/source/contributor/pre-commit.rst
@@ -23,9 +23,18 @@ Currently the following hooks are provided:
 
 Installation of `pre-commit` hooks is handled via the `install-pre-commit-hooks` playbook
 found within the Ansible directory.
-Either use `kayobe playbook run` or add the playbook as a hook within Kayobe config such as
+Either run the playbook manually or add the playbook as a hook within Kayobe config such as
 within `control-host-bootstrap/post.d`.
 Once done you should find `pre-commit` is available within the `kayobe` virtualenv.
+
+To run the playbook using the following command
+
+- ``kayobe playbook run ${KAYOBE_CONFIG_PATH}/ansible/install-pre-commit-hooks.yml``
+
+Whereas to run the playbook when control host bootstrap runs ensure it registered as symlink using the following command
+
+- ``mkdir -p ${KAYOBE_CONFIG_PATH}/hooks/control-host-bootstrap/post.d``
+- ``ln -s ${KAYOBE_CONFIG_PATH}/ansible/install-pre-commit-hooks.yml ${KAYOBE_CONFIG_PATH}/hooks/control-host-bootstrap/post.d/install-pre-commit-hooks.yml``
 
 All that remains is the installation of the hooks themselves which can be accomplished either by
 running `pre-commit run` or using `git commit` when you have changes that need to be committed.
--- a/etc/kayobe/ansible/install-pre-commit-hooks.yml
+++ b/etc/kayobe/ansible/install-pre-commit-hooks.yml
@@ -2,13 +2,15 @@
 - name: Install pre-commit hooks
   hosts: localhost
   gather_facts: false
+  vars:
+    pre_commit_version: 3.5.0
   tasks:
     - name: Install pre-commit hooks
       block:
         - name: Install pre-commit hooks into kayobe virtual env
           ansible.builtin.pip:
             name: pre-commit
-            version: 3.5.0
+            version: "{{ pre_commit_version  }}"
             virtualenv: "{{ lookup('ansible.builtin.env', 'VIRTUAL_ENV') }}"
           register: pip_install
 
--- a/etc/kayobe/ansible/install-pre-commit-hooks.yml
+++ b/etc/kayobe/ansible/install-pre-commit-hooks.yml
@@ -19,4 +19,3 @@
             creates: "{{ lookup('ansible.builtin.env', 'KAYOBE_CONFIG_PATH') }}/../../.git/hooks/pre-commit"
           args:
             chdir: "{{ lookup('ansible.builtin.env', 'KAYOBE_CONFIG_PATH') }}"
-      when: enable_pre_commit_hooks | default(false)
--- a/etc/kayobe/ansible/install-pre-commit-hooks.yml
+++ b/etc/kayobe/ansible/install-pre-commit-hooks.yml
@@ -10,10 +10,11 @@
             name: pre-commit
             version: 3.5.0
             virtualenv: "{{ lookup('ansible.builtin.env', 'VIRTUAL_ENV') }}"
+          register: pip_install
 
         - name: Register pre-commit hooks with git
           ansible.builtin.command:
             cmd: "{{ lookup('ansible.builtin.env', 'VIRTUAL_ENV') }}/bin/pre-commit install"
-            creates: "{{ lookup('ansible.builtin.env', 'KAYOBE_CONFIG_PATH') }}/../../.git/hooks/pre-commit"
           args:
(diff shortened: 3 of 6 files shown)


releasenotes/notes/add-pre-commit-hooks-07ce3b82bbe1d7a3.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Add playbook to install `pre-commit` hooks and register them with git.
```
priteau: I think you need double backticks throughout this file.

priteau (dismissed):

releasenotes/notes/add-pre-commit-hooks-07ce3b82bbe1d7a3.yaml
jackhodgkiss: Looking at the docs the examples use single back ticks https://docs.openstack.org/reno/latest/user/examples.html#input-file

```
fixes_command_line:
  - |
    This is a subsection. It requires setting `sections` in
```

priteau: > Looking at the docs the examples use single back ticks https://docs.openstack.org/reno/latest/user/examples.html#input-file
> 
> ```
> fixes_command_line:
>   - |
>     This is a subsection. It requires setting `sections` in
> ```

I think they got it wrong… And it isn't rendered so we don't see the result. Single backtick produces italic text with reno.

See https://docutils.sourceforge.io/docs/user/rst/quickref.html for details.

jackhodgkiss: Thanks for the information I have added the correct number of backticks.

releasenotes/notes/add-pre-commit-hooks-07ce3b82bbe1d7a3.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Add playbook to install ``pre-commit`` hooks and register them with git.
```
Alex-Welsh: Does pre-commit need to be `codeified`? It seems like just a regular word to me. I know it's also the name of the package but it works just as well as a generic description

priteau: Sure, we could remove them. Maybe Jack really meant to use italic for this one?

jackhodgkiss: It was only to add emphasis. Though I may use that more than I should.

Will clean up shortly.

priteau: In this case single backtick was correct, sorry!

Alex-Welsh (dismissed):

markgoddard (commented):
How are we expecting this to be used? Is it something that would be enabled for everyone in a given Kayobe config, or opted into per-person?

etc/kayobe/ansible/install-pre-commit-hooks.yml
```diff
@@ -0,0 +1,22 @@
...
+            version: 3.5.0
+            virtualenv: "{{ lookup('ansible.builtin.env', 'VIRTUAL_ENV') }}"
+
+        - name: Register pre-commit hooks with git
+          ansible.builtin.command:
+            cmd: "{{ lookup('ansible.builtin.env', 'VIRTUAL_ENV') }}/bin/pre-commit install"
+            creates: "{{ lookup('ansible.builtin.env', 'KAYOBE_CONFIG_PATH') }}/../../.git/hooks/pre-commit"
+          args:
+            chdir: "{{ lookup('ansible.builtin.env', 'KAYOBE_CONFIG_PATH') }}"
+      when: enable_pre_commit_hooks | default(false)
```
markgoddard: I think I'd prefer that this gets enabled by creating a symlink. That way we don't see the skipped tasks if it's not.

etc/kayobe/ansible/install-pre-commit-hooks.yml
```diff
@@ -0,0 +1,22 @@
...
+        - name: Install pre-commit hooks into kayobe virtual env
+          ansible.builtin.pip:
+            name: pre-commit
+            version: 3.5.0
+            virtualenv: "{{ lookup('ansible.builtin.env', 'VIRTUAL_ENV') }}"
+
+        - name: Register pre-commit hooks with git
+          ansible.builtin.command:
+            cmd: "{{ lookup('ansible.builtin.env', 'VIRTUAL_ENV') }}/bin/pre-commit install"
+            creates: "{{ lookup('ansible.builtin.env', 'KAYOBE_CONFIG_PATH') }}/../../.git/hooks/pre-commit"
```
markgoddard: Will this prevent pre-commit from being updated?

markgoddard (changes requested):
It would be nice to provide some simple docs on this in the contributor guide.

jackhodgkiss:
> How are we expecting this to be used? Is it something that would be enabled for everyone in a given Kayobe config, or opted into per-person?

I think having it is as an opt-in at least within SKC it the sensible move. If individual engineers want to enable it within their own checkout of the site config they can do and in situations where their is a single checkout shared enabling it would a group decision.

etc/kayobe/ansible/install-pre-commit-hooks.yml
jackhodgkiss: I don't believe so. Though I will investigate if this is the case.

etc/kayobe/ansible/install-pre-commit-hooks.yml
jackhodgkiss: To clarify. You want the `when` property removed so the playbook will run providing the the symlink exists within `hooks/control-host-bootstrap/post.d` which it currently is.

If so should I remove the symlink so it remains an opt-in.

I will ensure this behaviour and steps for enabling is documentated.

markgoddard: Correct

etc/kayobe/ansible/install-pre-commit-hooks.yml
jackhodgkiss: This appears to install the hook within `.git` directory and the presence of `creates` will mean any changes to such [template](https://github.com/pre-commit/pre-commit/blob/a9f19f4cc07d6fc3e0410bdc37a3d99c944760a7/pre_commit/resources/hook-tmpl) will be ignored. Could drop the `creates` to avoid this potential issue.

Alex-Welsh (dismissed):
My one lingering concern is that false positives could break things but they can be skipped with --no-verify so I think the benefits outweigh the risks

etc/kayobe/ansible/install-pre-commit-hooks.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Install pre-commit hooks
+  hosts: localhost
+  gather_facts: false
+  vars:
+    enable_pre_commit_hooks: false
```
markgoddard: Unused?

etc/kayobe/ansible/install-pre-commit-hooks.yml
markgoddard: I'd be tempted to drop it, or apply it only when the previous task is unchanged

doc/source/contributor/pre-commit.rst
```diff
@@ -0,0 +1,38 @@
...
+   such as those found within `secrets.yml` or `passwords.yml`.
+   However if the secret is contained within a file on it's own and lacks a certain level
+   of entropy then the secret will not be identified as such as and maybe leaked as a result.
+
+Installation of `pre-commit` hooks is handled via the `install-pre-commit-hooks` playbook
+found within the Ansible directory.
+Either use `kayobe playbook run` or add the playbook as a hook within Kayobe config such as
+within `control-host-bootstrap/post.d`.
+Once done you should find `pre-commit` is available within the `kayobe` virtualenv.
+
```
markgoddard, line 29: It's always nice to provide copy-pastable commands to avoid excessive brain usage

markgoddard (changes requested):

markgoddard (approved):

markgoddard (changes requested):
Actually perhaps this should go to antelope only?

Alex-Welsh:
@jackhodgkiss should we get this merged? I'm happy to rebase it on antelope for you if you're busy

etc/kayobe/ansible/install-pre-commit-hooks.yml
```diff
@@ -0,0 +1,22 @@
...
+  vars:
+    pre_commit_version: 3.5.0
+  tasks:
+    - name: Install pre-commit hooks
+      block:
+        - name: Install pre-commit hooks into kayobe virtual env
+          ansible.builtin.pip:
+            name: pre-commit
+            version: "{{ pre_commit_version  }}"
+            virtualenv: "{{ lookup('ansible.builtin.env', 'VIRTUAL_ENV') }}"
```
markgoddard: Is there an issue here that it forces us to use a virtualenv when developing SKC? e.g. I don't necessarily have one set up when writing docs.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/sirwart/ripsecrets
+    rev: v0.1.7
+    hooks:
+      - id: ripsecrets
```
markgoddard, line 11: This one seems to require a rust installation, and mine wasn't new enough.
```
[INFO] Installing environment for https://github.com/sirwart/ripsecrets.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/home/mark/.cargo/bin/cargo', 'install', '--bins', '--root', '/home/mark/.cache/pre-commit/repohp1cqnt3/rustenv-system', '--path', '.')
return code: 101
stdout: (none)
stderr:
      Installing ripsecrets v0.1.7 (/home/mark/.cache/pre-commit/repohp1cqnt3)
        Updating crates.io index
    error: failed to compile `ripsecrets v0.1.7 (/home/mark/.cache/pre-commit/repohp1cqnt3)`, intermediate artifacts can be found at `/home/mark/.cache/pre-commit/repohp1cqnt3/target`
    
    Caused by:
      package `clap v4.5.16` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.70.0
      Either upgrade to rustc 1.74 or newer, or use
      cargo update -p clap@4.5.16 --precise ver
      where `ver` is the latest version of `clap` supporting rustc 1.70.0
Check the log at /home/mark/.cache/pre-commit/pre-commit.log
```

etc/kayobe/ansible/install-pre-commit-hooks.yml
jackhodgkiss: I have updated this to support installation without a virtualenv if one isn't present.

.pre-commit-config.yaml
jackhodgkiss: I think this is a problem if `rustc` is already installed. It might prioritise the existing installation as opposed to running the one included within `/home/mark/.cache/pre-commit/repohp1cqnt3`

jackhodgkiss: If Rust toolchain is detected it will not install Rust for building of pre-commit hooks that have it as a build dependency [1].

Therefore, if an older version is installed it might not be compatible with specific hooks used.

However, ripsecrets v0.1.7 does work with rust 1.7.0 it only fails because Cargo does use the `cargo.lock` by default and requires `--locked` to be passed [2].

This [PR](https://github.com/pre-commit/pre-commit/pull/3163) proposes a simple change to pass `--locked` which I have tested and does work.

Maybe `cargo.conf` could force this [3]?

Not sure how you would like to proceed with this issue.

- [1] https://github.com/pre-commit/pre-commit/blob/main/pre_commit/languages/rust.py#L30
- [2] https://doc.rust-lang.org/cargo/commands/cargo-install.html#:~:text=The%20%2D%2Dlocked%20flag%20can,when%20the%20package%20was%20published.
- [3] https://doc.rust-lang.org/cargo/reference/config.html

markgoddard (approved):

PR #803, opened 2023-11-28
Fix OSD summary pie chart

jovial:
Only tested on antelope, Ceph Quincy on Ubuntu. Changes from:

image

to

image


etc/kayobe/kolla/config/grafana/dashboards/ceph/ceph_osds.json
```diff
@@ -539,18 +596,36 @@
...
     },
     {
-      "cacheTimeout": null,
-      "columns": [],
-      "datasource": "$datasource",
+      "datasource": {
+        "type": "prometheus",
+        "uid": "$datasource"
+      },
       "description": "The pie chart shows the various OSD sizes used within the cluster",
```
jovial, line 608: This is the panel that has changed. But re-exporting has updated multiple bits.

markgoddard (approved):
Picture looks good...

PR #807, opened 2023-11-29
Docs: Mention broken pxe boot_interface

jovial:


doc/source/operations/rocky-linux-9.rst
```diff
@@ -110,6 +110,29 @@ The error from NetworkManager was:
...
+  .. code-block:: shell
+
+      Failed to prepare to deploy: Could not link image http://192.168.1.1:8080/ipa.vmlinuz from /httpboot/master_images/99d5b4b4-0420-578a-a327-acd88c1f1ff6.converted to /tftpboot/d6673eaa-17a4-4cd4-a4e7-8e8cbd4fca31/deploy_kernel, error: [Errno 18] Invalid cross-device link: '/httpboot/master_images/99d5b4b4-0420-578a-a327-acd88c1f1ff6.converted' -> '/tftpboot/d6673eaa-17a4-4cd4-a4e7-8e8cbd4fca31/deploy_kernel'
+
+It is suggested that you migrate to iPXE.
+
+.. code-block:: yaml
+   :caption: ``$KAYOBE_CONFIG_PATH/kolla/config/bifrost/bifrost.yml``
+
+   enabled_boot_interfaces: "ipxe,pxe"
```
markgoddard: The default in ironic includes ipxe: https://opendev.org/openstack/ironic/src/branch/stable/yoga/ironic/conf/default.py#L103.

Bifrost has enabled_boot_interfaces empty by default, and kayobe doesn't override it, so we should be using ironic's default.

doc/source/operations/rocky-linux-9.rst
```diff
@@ -110,6 +110,29 @@ The error from NetworkManager was:
...
+   ipxe_efi_binary: "ipxe-snponly-x86_64.efi"
```
markgoddard: Is a fix required in bifrost for this?

doc/source/operations/rocky-linux-9.rst
jovial: In antelope they started using to variable to template the dnsmaq config file: https://github.com/openstack/bifrost/blob/stable/2023.1/playbooks/roles/bifrost-ironic-install/templates/dnsmasq.conf.j2#L106
but maybe in yoga we are OK.

doc/source/operations/rocky-linux-9.rst
jovial: I guess using snponly is optional, but it is quicker since it won't cycle through all your interfaces

doc/source/operations/rocky-linux-9.rst
jovial: I've removed this suggestion

markgoddard (approved):

PR #790, opened 2023-11-15
yoga host image selinux

markgoddard:
- RL9: Set SELinux to permissive mode in the standard overcloud host image
- RL9: Use Pulp repo snapshots when building overcloud host image

PR #808, opened 2023-11-29
Set selinux policy to permissive

jovial:
This will avoid the need for a reboot:

```
TASK [selinux : Ensure desired SELinux state] **********************************************************************************************************************[36/1977]
Wednesday 29 November 2023  12:45:39 +0100 (0:00:00.850)       0:01:15.944 ****                                                                                              
[WARNING]: Reboot is required to set SELinux state to 'permissive'
changed: [monitor1]                                                                   
                                                                                      
TASK [selinux : Abort SELinux configuration because reboot is disabled] *****************************************************************************************************
Wednesday 29 November 2023  12:45:41 +0100 (0:00:01.457)       0:01:17.401 **** 
fatal: [monitor1]: FAILED! => changed=false               
  msg: |-                                                                             
    SELinux state change requires a reboot, but selinux_do_reboot is false. Please run again with selinux_do_reboot set to true to reboot.

```

### DIFF
--- a/etc/kayobe/stackhpc-overcloud-dib.yml
+++ b/etc/kayobe/stackhpc-overcloud-dib.yml
@@ -23,7 +23,7 @@ stackhpc_overcloud_dib_name: "deployment_image"
 stackhpc_overcloud_dib_elements:
   - "{{ os_distribution }}-{% if os_distribution == 'rocky' %}container-stackhpc{% else %}minimal{% endif %}"
   - "cloud-init-datasources"
-  - "{% if os_distribution in ['centos', 'rocky'] %}disable-selinux{% endif %}"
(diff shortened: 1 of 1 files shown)


jovial:
Attempting to build this here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/7032155163

markgoddard:
I did something similar in #790. The main difference is I only set to permissive on RL9. This aligns with #793.

The main issue is that the image includes RL9.3, whereas our snapshots are still 9.2. I tried pinning the repos in the image but it didn't work.

jovial:
Closing as I wasn't aware of the other PR. Thought this would be a quick fix to save some reboots, but fixing to package snapshots seems like a bigger task.

PR #809, opened 2023-11-29
Add stop-openstack-services.yml playbook

Alex-Welsh:
Adds a new playbook to stop openstack services across the overcloud.

I was undecided as to whether this warrants its own playbook but I've needed to use it enough times now that I think other people might want it.

One function I was considering was a variable to change the `docker stop` to a `docker restart` which we could then redirect to in the case of the `rabbitmq-reset.yml` playbook.

### DIFF
--- a/etc/kayobe/ansible/stop-openstack-services.yml
+++ b/etc/kayobe/ansible/stop-openstack-services.yml
@@ -1,0 +1,29 @@
+---
+# Stops containers running OpenStack services
+
+- name: Stop OpenStack services
+  hosts: overcloud
+  become: true
+  gather_facts: false
+  vars:
+    - stop_service_list:
+        - "blazar"
+        - "barbican"
+        - "cinder"
+        - "cloudkitty"
+        - "designate"
+        - "glance"
+        - "heat"
+        - "horizon"
+        - "ironic"
+        - "keystone"
+        - "magnum"
+        - "manila"
+        - "neutron"
+        - "nova"
+        - "octavia"
+        - "placement"
+  tasks:
+    - name: Stop OpenStack services
+      shell: >-
+        docker ps -a | egrep '({{ stop_service_list | join('|') }})' | awk '{ print $NF }' | xargs docker stop
--- a/releasenotes/notes/stop-services-playbook-b85b53d1a7571009.yaml
+++ b/releasenotes/notes/stop-services-playbook-b85b53d1a7571009.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
+    Added the ``stop-openstack-services.yml`` playbook, which can be used to
+    stop OpenStack services across the overcloud.


etc/kayobe/ansible/stop-openstack-services.yml
```diff
@@ -0,0 +1,27 @@
...
+        - "heat"
+        - "ironic"
+        - "keystone"
+        - "magnum"
+        - "manila"
+        - "neutron"
+        - "nova"
+        - "octavia"
+  tasks:
+    # The following services use RabbitMQ.
```
MoteHue: This comment isn't needed here

releasenotes/notes/stop-services-playbook-b85b53d1a7571009.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Added the ``stop-openstack-services.yml`` playbook, which can be used to
+    stop openstack services across the overcloud.
```
MoteHue:
```suggestion
    stop OpenStack services across the overcloud.
```

Alex-Welsh: :angry:

Alex-Welsh: Done

etc/kayobe/ansible/stop-openstack-services.yml
Alex-Welsh: I blame the guy I copied it from

etc/kayobe/ansible/stop-openstack-services.yml
```diff
@@ -0,0 +1,26 @@
+---
+# Stops containers running OpenStack services
+
+- name: Stop OpenStack services
+  hosts: overcloud
+  become: true
+  gather_facts: false
+  vars:
+    - stop_service_list:
```
markgoddard, line 9: Needs horizon & glance

markgoddard (changes requested):

etc/kayobe/ansible/stop-openstack-services.yml
Alex-Welsh: Added

Alex-Welsh: Plus placement

mnasiadka:
Maybe a stupid question, but what's wrong with kolla-ansible stop? ;-)

markgoddard:
> Maybe a stupid question, but what's wrong with kolla-ansible stop? ;-)

Slooooow

mnasiadka:
> > Maybe a stupid question, but what's wrong with kolla-ansible stop? ;-)
> 
> Slooooow

Correct answer, should we rework that upstream as well?

markgoddard:
> > > Maybe a stupid question, but what's wrong with kolla-ansible stop? ;-)
> > 
> > 
> > Slooooow
> 
> Correct answer, should we rework that upstream as well?

How do you propose to do it?

mnasiadka:
> > > > Maybe a stupid question, but what's wrong with kolla-ansible stop? ;-)
> > > 
> > > 
> > > Slooooow
> > 
> > 
> > Correct answer, should we rework that upstream as well?
> 
> How do you propose to do it?
Maybe a similar playbook, just for those time constrained - but we would probably need a better mechanism for being able to tell which services need stopping - maybe a playbook with a tag per play? It's going to be slower than this, but probably faster than kolla-ansible stop currently.

Alex-Welsh:
> > > > > Maybe a stupid question, but what's wrong with kolla-ansible stop? ;-)
> > > > 
> > > > 
> > > > Slooooow
> > > 
> > > 
> > > Correct answer, should we rework that upstream as well?
> > 
> > 
> > How do you propose to do it?
> > Maybe a similar playbook, just for those time constrained - but we would probably need a better mechanism for being able to tell which services need stopping - maybe a playbook with a tag per play? It's going to be slower than this, but probably faster than kolla-ansible stop currently.

Playbook with a tag per play that adds all the services to a list, then runs the stop all at once? Would be a lot faster than trying to stop them all individually.

markgoddard (approved):

PR #810, opened 2023-11-30
Add placement-api in the rabbitmq-playbook

hkominos:
After rabbitmq is reset, the openstack services get restarted but placement_api does not because it is was not grepped.


priteau:
Placement doesn't use RabbitMQ.

PR #811, opened 2023-12-01
Remove rhel8 cis

Alex-Welsh:
RHEL 8 is no longer supported, so this change removes RHEL8-specific config for CIS hardening. The CIS playbook has been updated for RHEL 9 but is modified to warn the user that it is untested


MoteHue:
Will this not get merged upwards from Yoga? https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/yoga/etc/kayobe/ansible/cis.yml
Agreed that we can drop the rhel8 at that point

PR #816, opened 2023-12-05
Add a note about network interfaces changes

priteau:


doc/source/operations/rocky-linux-9.rst
```diff
@@ -242,6 +242,13 @@ Potential issues
 -  If you are using hyper-converged Ceph, please also note the potential issues
    in the Storage section below.
 
+-  Network interface names may change between CentOS Stream 8 and Rocky Linux
+   9, in which case you will need to update Kayobe configuration. Note that the
+   configuration should remain correct for hosts not yet migrated, otherwise
+   fact gathering may fail. For example, this can be done using ``host_vars``.
+   Once all hosts are migrated, the change can be moved to ``group_vars`` and
+   the temporary ``host_vars`` deleted.
```
Alex-Welsh: It may be better to suggest adding the migrated hosts to a new group and setting a group var, so you only have to set the config once and can move the hosts into the group when they're reprovisioned

markgoddard: I've tried to do this.

markgoddard (approved):

PR #819, opened 2023-12-06
Disable SELinux for Yoga environments

assumptionsandg:
Disable SELinux for Yoga CI environments to avoid need for reboot as seen in Zed/Antelope.

### DIFF
--- a/etc/kayobe/environments/ci-aio/globals.yml
+++ b/etc/kayobe/environments/ci-aio/globals.yml
@@ -59,7 +59,7 @@ os_release: >-
 ###############################################################################
 
 # Avoid a reboot.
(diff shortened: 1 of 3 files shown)


priteau:
We explicitly switched to permissive in stackhpc/yoga: https://github.com/stackhpc/stackhpc-kayobe-config/pull/793

Should multinode do the same to match test environments?

markgoddard:
> We explicitly switched to permissive in stackhpc/yoga: #793
> 
> Should multinode do the same to match test environments?

We should test the default config where possible. We're looking at changing the default in the RL9 image, which would make this more user friendly.

assumptionsandg:
Duplicate of #854 and #836

PR #831, opened 2023-12-11
Pin base image tag

bbezak:
Pin the OCI image tag used for the base-image of Kolla image
builds. This prevents packages in the image with
the latest tag getting in front of StackHPC release-train
package repositories. Ubuntu tag should be bumped when new
packages are available in StackHPC release-train.

### DIFF
--- a/etc/kayobe/kolla/kolla-build.conf
+++ b/etc/kayobe/kolla/kolla-build.conf
@@ -1,2 +1,11 @@
 [DEFAULT]
+{# Packages in the latest upstream Ubuntu base image can be ahead of our package repo #}
+{# snapshots, so pin to a specific tag. #}
+{# This tag should be updated when Ubuntu package repo snapshot versions are changed. #}
+{% if kolla_base_distro == 'ubuntu' %}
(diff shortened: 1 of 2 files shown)


bbezak:
tested here https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/7167964981

priteau (commented):
Do we need the same for Rocky Linux?

markgoddard:
> Do we need the same for Rocky Linux?

We should probably pin to the minor version, which we could get from stackhpc_pulp_repo_rocky_9_minor_version.

bbezak:
> > Do we need the same for Rocky Linux?
> 
> We should probably pin to the minor version, which we could get from stackhpc_pulp_repo_rocky_9_minor_version.

added, testing in https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/7180464635

markgoddard (commented):
Should it go to Zed first?

bbezak:
> Should it go to Zed first?

https://github.com/stackhpc/stackhpc-kayobe-config/pull/835

PR #834, opened 2023-12-12
Pin R9 base image tag

bbezak:


releasenotes/notes/pin-rocky9-baseimage-3d224df5b6f56b10.yaml
```diff
@@ -0,0 +1,6 @@
+fixes:
+  - |
+    Pin the OCI image tag used for the base-image of Rocky 9
+    Kolla image builds. This prevents packages in the image with
```
mnasiadka, line 4: I wouldn't say prevents, but meh.

mnasiadka (approved):

releasenotes/notes/pin-rocky9-baseimage-3d224df5b6f56b10.yaml
bbezak: yeah, however looks like RHEL is not updating their minor versions tags after release:
https://quay.io/repository/rockylinux/rockylinux?tab=tags&tag=latest

markgoddard (approved):

PR #835, opened 2023-12-12
Pin base image

bbezak:
Pin the OCI image tag used for the base-image of Kolla image
builds. This prevents packages in the image with
the latest tag getting in front of StackHPC release-train
package repositories. Ubuntu tag should be bumped when new
packages are available in StackHPC release-train.

### DIFF
--- a/etc/kayobe/kolla/kolla-build.conf
+++ b/etc/kayobe/kolla/kolla-build.conf
@@ -1,2 +1,11 @@
 [DEFAULT]
+{# Packages in the latest upstream Ubuntu base image can be ahead of our package repo #}
+{# snapshots, so pin to a specific tag. #}
+{# This tag should be updated when Ubuntu package repo snapshot versions are changed. #}
+{% if kolla_base_distro == 'ubuntu' %}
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla/kolla-build.conf
```diff
@@ -1,2 +1,11 @@
 [DEFAULT]
+{# Packages in the latest upstream Ubuntu base image can be ahead of our package repo #}
+{# snapshots, so pin to a specific tag. #}
+{# This tag should be updated when Ubuntu package repo snapshot versions are changed. #}
+{% if kolla_base_distro == 'ubuntu' %}
+base_tag = jammy-20231004
+{# Similarily pinning to Rocky 9 minor version used in our repos #}
```
m-bull:
```suggestion
{# Similarly pinning to Rocky 9 minor version used in our repos #}
```

bbezak: thx

markgoddard (approved):

PR #839, opened 2023-12-14
Ovn syntax

g0rgamesh:
… used in file


markgoddard:
Not required in antelope, it will get merged up from https://github.com/stackhpc/stackhpc-kayobe-config/pull/838.

PR #842, opened 2023-12-15
Disable os-capacity by default

priteau:
The HAProxy configuration for os-capacity is enabled by default and is causing alerts on every deployment until the exporter is deployed.

Disable it while its configuration is being finalised [1].

[1] https://github.com/stackhpc/stackhpc-kayobe-config/pull/697

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/prometheus.yml.d/70-oscapacity.yml
+++ b/etc/kayobe/kolla/config/prometheus/prometheus.yml.d/70-oscapacity.yml
@@ -1,3 +1,5 @@
+# yamllint disable-file
+---
 {% if stackhpc_enable_os_capacity | bool %}
 {% raw %}
 scrape_configs:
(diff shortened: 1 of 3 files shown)


bbezak (dismissed):

JohnGarbutt (dismissed):
That looks more correct to me!

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/70-oscapacity.yml
```diff
@@ -1,3 +1,4 @@
+{% if stackhpc_enable_os_capacity | bool %}
```
JohnGarbutt, line 3: Should this still end with .j2 to keep tox happy?

markgoddard (approved):

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/70-oscapacity.yml
priteau: kolla-ansible only reads `*.yml`: https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/prometheus/tasks/config.yml#L58-L66

JohnGarbutt (approved):
cool, that works for me then, good to know!

PR #853, opened 2023-12-22
Allow resolv.conf to be managed for Ubuntu AIO

assumptionsandg:
Fix Ubuntu AIOs in CI by allowing resolv.conf to be managed by Kayobe.

### DIFF
--- a/etc/kayobe/environments/ci-aio/stackhpc-ci.yml
+++ b/etc/kayobe/environments/ci-aio/stackhpc-ci.yml
@@ -8,9 +8,10 @@ kolla_docker_namespace: stackhpc-dev
 ###############################################################################
 # Network configuration.
 
-# Don't touch resolv.conf: use Neutron DNS for accessing Pulp server via
-# hostname.
-resolv_is_managed: false
+# Allow resolv.conf to be managed for Ubuntu
+resolv_is_managed: "{{ os_distribution == 'ubuntu' }}"
+resolv_nameservers: [ "8.8.8.8", "8.8.4.4" ]
+resolv_domain: "internal.sms-cloud"
 
 ###############################################################################
 # StackHPC configuration.


assumptionsandg:
Solved in #854 for CI

Turns out DNS config wasn't being applied as networkd config was being set for ens3 rather than enp3s0. Still broken for manual deployment AFAIK.

markgoddard:
Should this PR be closed?

markgoddard:
> Solved in #854 for CI
> 
> Turns out DNS config wasn't being applied as networkd config was being set for ens3 rather than enp3s0. Still broken for manual deployment AFAIK.

Why is it broken in a manual deployment?

assumptionsandg:
> > Solved in #854 for CI
> > Turns out DNS config wasn't being applied as networkd config was being set for ens3 rather than enp3s0. Still broken for manual deployment AFAIK.
> 
> Why is it broken in a manual deployment?

networkd configuration for enp3s0 isn't being templated in when deploying manually via the automated script, hence the need for resolv.conf to be managed by Kayobe in these cases as DNS breaks when the device is left unmanaged.

markgoddard:
I have now hit this in an Ubuntu VM and can see the issue you're trying to address.

etc/kayobe/environments/ci-aio/stackhpc-ci.yml
```diff
@@ -8,9 +8,10 @@ kolla_docker_namespace: stackhpc-dev
 ###############################################################################
 # Network configuration.
 
-# Don't touch resolv.conf: use Neutron DNS for accessing Pulp server via
-# hostname.
```
markgoddard, line 12: I think we don't actually make use of this - we write out an entry in /etc/hosts. I suggest we just remove all of this config.

holliefae: To clarify do you mean remove the resolv config entirely or just remove the hostname resolv_domain?

markgoddard: We can remove everything in this `Network configuration` section

holliefae: I'll close this one and make a pr.

assumptionsandg:
Continued in #920

PR #854, opened 2023-12-22
Fix Interface and SELinux errors in CI

assumptionsandg:
Fix Interface and SELinux errors in CI:

- Change default interface from ens3 to enp3s0.
- Disable SELinux until we have new host images.


markgoddard:
This does seem to fix the issue. Do we have any idea why the interface name changed? It looks like CI was last passing around November 17th(!), but I don't see a change in the host image used since then

Alex-Welsh (approved):

markgoddard (approved):

PR #856, opened 2023-12-28
Smartmon update

technowhizz:


markgoddard (commented):
Does this apply to Yoga? If so it should go there first.

dougszumski:
> Does this apply to Yoga? If so it should go there first.

bump @technowhizz

technowhizz:
@markgoddard, Changed for Yoga

markgoddard (approved):

PR #862, opened 2024-01-04
Updated the AiO Runner to point to ARC runner.

MaxBed4d:
Changed the 'runs-on' to now point to the ARC Runner.

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -70,7 +70,7 @@ jobs:
   all-in-one:
     name: All in one
     if: inputs.if
-    runs-on: [self-hosted, stackhpc-kayobe-config-aio]
+    runs-on: arc-release-train-runner
     permissions: {}
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -70,7 +70,7 @@ jobs:
   all-in-one:
     name: All in one
     if: inputs.if
-    runs-on: [self-hosted, stackhpc-kayobe-config-aio]
+    runs-on: arc-release-train-runner
```
markgoddard, line 73: I've installed the github app on this repo, but we'll also need a new runner group for aios defined in the ARC-Installer repo. Then this should reference the name of the new runner group.

markgoddard (changes requested):

.github/workflows/stackhpc-all-in-one.yml
MaxBed4d: Ok I will get onto provisioning them.

MaxBed4d:
Old and not needed.

PR #865, opened 2024-01-05
Reduce multinode lvm allocations

Alex-Welsh:
We're constantly running low on space in SMS, this PR goes with another in the multinode terraform to reduce the allocations across the board.

These sizes are tested and working for Rocky 9

### DIFF
--- a/etc/kayobe/environments/ci-multinode/inventory/group_vars/compute/lvm.yml
+++ b/etc/kayobe/environments/ci-multinode/inventory/group_vars/compute/lvm.yml
@@ -7,25 +7,25 @@ stackhpc_lvm_group_rootvg_lvs_extra:
 # StackHPC LVM Logical Volume (LV) configuration.
 
 # StackHPC LVM lv_swap LV size.
-stackhpc_lvm_lv_swap_size: 1g
+stackhpc_lvm_lv_swap_size: 500m
 
 # StackHPC LVM lv_root LV size.
-stackhpc_lvm_lv_root_size: 10g
+stackhpc_lvm_lv_root_size: 5g
 
 # StackHPC LVM lv_tmp LV size.
-stackhpc_lvm_lv_tmp_size: 10g
+stackhpc_lvm_lv_tmp_size: 5g
 
 # StackHPC LVM lv_var LV size.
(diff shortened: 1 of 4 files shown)


Alex-Welsh:
See also: https://github.com/stackhpc/terraform-kayobe-multinode/pull/40

etc/kayobe/environments/ci-multinode/inventory/group_vars/compute/lvm.yml
```diff
@@ -7,25 +7,25 @@ stackhpc_lvm_group_rootvg_lvs_extra:
...
-stackhpc_lvm_lv_root_size: 10g
+stackhpc_lvm_lv_root_size: 5g
 
 # StackHPC LVM lv_tmp LV size.
-stackhpc_lvm_lv_tmp_size: 10g
+stackhpc_lvm_lv_tmp_size: 5g
 
 # StackHPC LVM lv_var LV size.
-stackhpc_lvm_lv_var_size: 20g
+stackhpc_lvm_lv_var_size: 2g
```
markgoddard, line 19: That's quite a drop!

markgoddard (commented):
I assume you didn't test upgrades? That's when we might see the most usage in lv_docker due to multiple sets of containers. The aio upgrade job worked with a 45G disk, but it has fewer services running.

I think we'd need to merge this before the multinode one.

Alex-Welsh:
> I assume you didn't test upgrades? That's when we might see the most usage in lv_docker due to multiple sets of containers. The aio upgrade job worked with a 45G disk, but it has fewer services running.

Correct, I was just thinking it could be expanded whenever needed for that sort of thing

etc/kayobe/environments/ci-multinode/inventory/group_vars/compute/lvm.yml
Alex-Welsh: I'm not exactly sure why it was so large in the first place to be honest. Not a lot goes on in there from what I can see

Alex-Welsh:
This PR isn't really relevant any more with SMS gone. If it's still required for SMS2 I'll reopen it

PR #869, opened 2024-01-12
CI: Add upgrade tests

markgoddard:
Adds two upgrade test jobs: Ubuntu Jammy OVS and Rocky 9 OVN.
    
These jobs reuse the existing aio reusable workflow, which has a new 'upgrade' input parameter. When testing an upgrade, the previous release is first deployed, then the aio is upgraded to the current release before running RefStack tests.

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -55,6 +55,10 @@ on:
         description: Whether to run the workflow (workaround for required status checks issue)
         type: boolean
         default: true
+      upgrade:
+        description: Whether to perform an upgrade
+        type: boolean
+        default: false
     secrets:
       KAYOBE_VAULT_PASSWORD:
         required: true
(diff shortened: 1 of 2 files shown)


MoteHue:
Is it SMS capacity limits that are motivating just adding two tests? I'm wondering how big of an issue it would be to not cover OVN/OVS on both distributions. We do have customers on OVN Ubuntu for example, so issues could slip through the cracks here.

markgoddard:
> Is it SMS capacity limits that are motivating just adding two tests? I'm wondering how big of an issue it would be to not cover OVN/OVS on both distributions. We do have customers on OVN Ubuntu for example, so issues could slip through the cracks here.

Correct - ideally we would cover all cases. There may be an argument for nightly runs covering all combinations, if we can report the results somewhere sensible.

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -76,10 +80,27 @@ jobs:
...
       KAYOBE_IMAGE: ${{ inputs.kayobe_image }}
+      # NOTE(upgrade): Reference the PREVIOUS release here.
+      PREVIOUS_KAYOBE_IMAGE: ghcr.io/stackhpc/stackhpc-kayobe-config:stackhpc-yoga
+      # NOTE(upgrade): Reference the PREVIOUS release branch here.
+      PREVIOUS_BRANCH: stackhpc/yoga
     steps:
-      - uses: actions/checkout@v3
+      # If testing upgrade, checkout previous release, otherwise checkout current branch
+      - name: Checkout ${{ inputs.upgrade && 'previous release' || 'current' }} config
+        uses: actions/checkout@v3
```
Alex-Welsh:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -246,11 +275,62 @@ jobs:
...
           KAYOBE_AUTOMATION_SSH_PRIVATE_KEY: ${{ steps.ssh_key.outputs.ssh_key }}
 
+      # If testing upgrade, checkout the current release branch
+      # Stash changes to tracked files, and set clean=false to avoid removing untracked files.
+      - name: Stash config changes
+        run: git stash
+        if: inputs.upgrade
+
+      - name: Checkout current release config
+        uses: actions/checkout@v3
```
Alex-Welsh:
```suggestion
        uses: actions/checkout@v4
```

Alex-Welsh (commented):
couple of nits otherwise LGTM

Alex-Welsh (approved):
Might be worth an update to release train so it's a required check

PR #875, opened 2024-01-15
DNM: Env to trigger AiORunners. 

MaxBed4d:


.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -77,6 +77,10 @@ jobs:
       KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
       KAYOBE_IMAGE: ${{ inputs.kayobe_image }}
     steps:
+      - name: Checkout submodules
```
Alex-Welsh: Isn't this done in the following step?

MaxBed4d: This is to try and [workaround](https://github.com/actions/checkout/issues/758) the error below:

> Error: Input 'submodules' not supported when falling back to download using the GitHub REST API

Seemingly, when git tries to tarball git it uses an old one which doesn't work with submodules.

Alex-Welsh: can you try checkout/v4 and submodules: recursive?
https://github.com/actions/checkout

MaxBed4d: Will try checkoutv4

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -70,14 +70,18 @@ jobs:
...
     permissions: {}
     env:
       KAYOBE_ENVIRONMENT: ci-aio
       KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
       KAYOBE_IMAGE: ${{ inputs.kayobe_image }}
     steps:
-      - uses: actions/checkout@v3
+      - name: Install or Update Git.
+        run: |
+          sudo apt-get update && sudo apt-get install -y git && sudo apt-get install -y nodejs && sudo apt-get install -y unzip && sudo apt-get install -y nodejs && sudo apt-get install -y -f && sudo apt-get build-dep build-essential
```
Alex-Welsh: Perhaps try this action rather than installing everything manually: https://github.com/marketplace/actions/install-package

MaxBed4d: Tried it but it doenst use the correct package manager and would have run apt-get manually regardless, they even suggest not using their action for this case at the bottom of their document.

MaxBed4d: Fixed their action not being able to find the packages.

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -70,17 +70,35 @@ jobs:
...
+        uses: ConorMacBride/install-package@main
+        with:
+          apt: build-essential git unzip nodejs
+
+      - name: Upgrade python version
+        run: sudo apt-get -y install python3 pip python-is-python3
+
+      - name: Install wheel from pip
+        run: sudo pip install wheel        
+
```
markgoddard: Let's tidy this up a bit
```suggestion
      - name: Install Package
        uses: ConorMacBride/install-package@main
        with:
          apt: build-essential git unzip nodejs python3 python3-pip python3-wheel python-is-python3

```

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -103,10 +121,18 @@ jobs:
         run: terraform init
         working-directory: ${{ github.workspace }}/terraform/aio
 
+      - name: Create SSH key directory
+        run: mkdir -p ~/.ssh/
+
```
markgoddard: I don't think this is necessary

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -103,10 +121,18 @@ jobs:
...
       - name: Generate SSH keypair
         run: ssh-keygen -f id_rsa -N ''
         working-directory: ${{ github.workspace }}/terraform/aio
 
+      - name: Copy SSH keypair to ~/.ssh/
+        run: cp id_rsa* ~/.ssh/
+        working-directory: ${{ github.workspace }}/terraform/aio
+
+
```
markgoddard: I don't think this is necessary

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -209,32 +235,51 @@ jobs:
...
+      
+      # - name: Find and then test ~/.ssh/id_rsa to see if the file or the repo exists.
+      #   run: |
+      #     find . -type d -name "*ssh*" 2>/dev/null && \
+      #     find . -type f -name "*id_rsa*" 2>/dev/null && \
+      #     sudo ls -ltra ~/.ssh/id_rsa && \
+      #     sudo ls -ltra ~/.ssh/ && \
+      #   env:
+      #     KAYOBE_AUTOMATION_SSH_PRIVATE_KEY: ${{ steps.ssh_key.outputs.ssh_key }}
+
```
markgoddard: Please clean up the debugging before marking ready for review

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -209,32 +235,51 @@ jobs:
...
 
       - name: Run growroot
         run: |
-          sudo -E docker run -t --rm \
+          docker run -t --rm -u 1001 \
```
markgoddard: I don't think `-u 1001` is required. Same for others.

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -209,32 +235,51 @@ jobs:
...
             -v $(pwd):/stack/kayobe-automation-env/src/kayobe-config \
+            -v ~/.ssh:/.ssh \
```
markgoddard: I don't think this is required.  Same for others.

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -209,32 +235,51 @@ jobs:
...
       - name: Host configure
         run: |
-          sudo -E docker run -t --rm \
+          docker run -t --rm -u 1001 \
             -v $(pwd):/stack/kayobe-automation-env/src/kayobe-config \
+            -v ~/.ssh:/.ssh \
             -e KAYOBE_ENVIRONMENT -e KAYOBE_VAULT_PASSWORD -e KAYOBE_AUTOMATION_SSH_PRIVATE_KEY \
             $KAYOBE_IMAGE \
-            /stack/kayobe-automation-env/src/kayobe-config/.automation/pipeline/overcloud-host-configure.sh
+            /stack/kayobe-automation-env/src/kayobe-config/.automation/pipeline/overcloud-host-configure.sh -vvv
```
markgoddard: Please remove the `-vvv`

.github/workflows/stackhpc-build-kayobe-image.yml
```diff
@@ -86,6 +86,8 @@ jobs:
           build-args: |
              http_proxy=${{ inputs.http_proxy }}
              https_proxy=${{ inputs.https_proxy }}
+             KAYOBE_USER_UID=1001
+             KAYOBE_USER_GID=1001
```
markgoddard: It would be helpful to add a comment about why we're changing these.

markgoddard (changes requested):

.github/workflows/stackhpc-all-in-one.yml
MaxBed4d: I have previously tried this, however, I had some issues with installing some of the packages using the 'install-package main'. However, maybe it was because of a previous error.

.github/workflows/stackhpc-all-in-one.yml
MaxBed4d: Cleaned.

.github/workflows/stackhpc-all-in-one.yml
MaxBed4d: Removed.

.github/workflows/stackhpc-all-in-one.yml
MaxBed4d: Removed.

.github/workflows/stackhpc-all-in-one.yml
MaxBed4d: Removed.

.github/workflows/stackhpc-all-in-one.yml
MaxBed4d: Removed.

Alex-Welsh:
Is this PR still required since we're now using arc in our main branches?

markgoddard:
> Is this PR still required since we're now using arc in our main branches?

It shouldn't be necessary.

PR #876, opened 2024-01-15
vault: Support generating external TLS certificates for testing

markgoddard:
These should not generally be used in production, since the CA is self-signed.


markgoddard:
Would need additional Kayobe config to enable external TLS.

markgoddard:
Requires some changes in terraform-kayobe-multinode (on the external-tls branch) to be useful. Currently blocked on getting the CA cert into the rally container when running tempest.

markgoddard:
Merged in 2023.1

PR #880, opened 2024-01-17
AUFN-Ceph fixes

MoteHue:
Now that we're lacking in baremetals, AUFNs need to be run in VMs. The geneve network on SMS lab has MTU 1442, so networks need to be lowered to match this.

### DIFF
--- a/etc/kayobe/environments/aufn-ceph/a-universe-from-nothing.sh
+++ b/etc/kayobe/environments/aufn-ceph/a-universe-from-nothing.sh
@@ -87,7 +87,7 @@ kayobe seed vm provision
 kayobe seed host configure
 
 # Deploy local pulp server as a container on the seed VM
-kayobe seed service deploy --tags seed-deploy-containers --kolla-tags none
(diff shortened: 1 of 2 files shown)


cityofships (approved):

etc/kayobe/environments/aufn-ceph/a-universe-from-nothing.sh
```diff
@@ -87,7 +87,7 @@ kayobe seed vm provision
 kayobe seed host configure
 
 # Deploy local pulp server as a container on the seed VM
-kayobe seed service deploy --tags seed-deploy-containers --kolla-tags none
+kayobe seed service deploy --tags seed-deploy-containers --kolla-tags none -e deploy_containers_registry_attempt_login=False
```
markgoddard, line 90: I was hoping we would set this in our config, then add a login task to the pulp container post deploy tasks.

MoteHue: Something like this? https://github.com/stackhpc/stackhpc-kayobe-config/pull/891

PR #884, opened 2024-01-18
Stop deploying Docker registry

cityofships:
We don't use this one, do we?

### DIFF
--- a/etc/kayobe/docker-registry.yml
+++ b/etc/kayobe/docker-registry.yml
@@ -3,7 +3,7 @@
 # Docker registry configuration.
 
 # Whether a docker registry is enabled. Default is false.
-#docker_registry_enabled:
+docker_registry_enabled: false
 
 # Dict of environment variables to provide to the docker registry container.
 # This allows to configure the registry by overriding specific configuration


priteau (approved):

Alex-Welsh:
The comment says default is false. If that's true why do we need to set the var?

cityofships:
> The comment says default is false. If that's true why do we need to set the var?

Comment inaccurate? https://github.com/stackhpc/kayobe/blob/stackhpc/2023.1/ansible/roles/docker-registry/defaults/main.yml#L9

Perhaps there was a higher precedence variable that got removed at some point?

MoteHue:
> > The comment says default is false. If that's true why do we need to set the var?
> 
> Comment inaccurate? https://github.com/stackhpc/kayobe/blob/stackhpc/2023.1/ansible/roles/docker-registry/defaults/main.yml#L9
> 
> Perhaps there was a higher precedence variable that got removed at some point?

It's overridden in Kayobe group vars: https://github.com/stackhpc/kayobe/blob/54b9e92f5e1e5fc9ea026cee682f9ce1ce0ea794/ansible/inventory/group_vars/all/docker-registry#L6

cityofships:
> > > The comment says default is false. If that's true why do we need to set the var?
> > 
> > 
> > Comment inaccurate? https://github.com/stackhpc/kayobe/blob/stackhpc/2023.1/ansible/roles/docker-registry/defaults/main.yml#L9
> > Perhaps there was a higher precedence variable that got removed at some point?
> 
> It's overridden in Kayobe group vars: https://github.com/stackhpc/kayobe/blob/54b9e92f5e1e5fc9ea026cee682f9ce1ce0ea794/ansible/inventory/group_vars/all/docker-registry#L6

Yet the task has been firing anyway: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/7574482605/job/20628969896#step:7:976

markgoddard:
> > > > The comment says default is false. If that's true why do we need to set the var?
> > > 
> > > 
> > > Comment inaccurate? https://github.com/stackhpc/kayobe/blob/stackhpc/2023.1/ansible/roles/docker-registry/defaults/main.yml#L9
> > > Perhaps there was a higher precedence variable that got removed at some point?
> > 
> > 
> > It's overridden in Kayobe group vars: https://github.com/stackhpc/kayobe/blob/54b9e92f5e1e5fc9ea026cee682f9ce1ce0ea794/ansible/inventory/group_vars/all/docker-registry#L6
> 
> Yet the task has been firing anyway: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/7574482605/job/20628969896#step:7:976

The role asserts that the service is not running rather than skipping it - that's why it shows as unchanged

PR #886, opened 2024-01-19
Set OpenSearch heap size

priteau:
This is required for deployments using Rocky Linux 9. Without this
setting, OpenSearch would set heap size to 1 GB by default, which can
trigger alerts on production systems.


grzegorzkoper (dismissed):
+1

Alex-Welsh (dismissed):

JohnGarbutt (dismissed):
Oh my, this is very important.

(its a shame we don't default the openserach_heap_size to the es_heap_size upstream)

JohnGarbutt (commented):
Should we add a release note, this is quite a big deal and might require people to review their overrides?

markgoddard (dismissed):
Probably should have done this in kolla-ansible

markgoddard (approved):

PR #890, opened 2024-01-19
Add pvresize before lvextend

seunghun1ee:
Running pvresize before lvextend ensures that there is enough space for extension.

There are some cases where `lvextend` fails because initial usable space of a physical volume is not enough.
This leads the automated setup failure at package or python module installation stage.
As this happens earlier than kayobe setup, `kayobe playbook run etc/kayobe/ansible/growroot.yml` can't be used.
https://github.com/stackhpc/stackhpc-kayobe-config/blob/e1a64bdb28201fe8b00b3adca07d6d366439961e/etc/kayobe/ansible/growroot.yml#L87-L93 
Therefore I propose adding bash equivalent of the Ansible task above before `lvextend`.

The command gets the name of the physical volume by filtering the output of `sudo pvs --noheadings`.
`head -n 1` is used to take first line of output only then `awk '{print $1}'` is used to take first element of it, which is essentially the way the `growroot.yml` does to get PV name.

### DIFF
--- a/etc/kayobe/environments/ci-aio/automated-setup.sh
+++ b/etc/kayobe/environments/ci-aio/automated-setup.sh
@@ -7,8 +7,7 @@ cat << EOF | sudo tee -a /etc/hosts
 EOF
 
 if sudo vgdisplay | grep -q lvm2; then
-   disk=$(sudo pvs --noheadings | head -n 1 | awk '{print $1}')
-   sudo pvresize $disk
+   sudo pvresize $(sudo pvs --noheadings | head -n 1 | awk '{print $1}')
    sudo lvextend -L 4G /dev/rootvg/lv_home -r || true
    sudo lvextend -L 4G /dev/rootvg/lv_tmp -r || true
 fi
--- a/etc/kayobe/environments/ci-aio/automated-setup.sh
+++ b/etc/kayobe/environments/ci-aio/automated-setup.sh
@@ -7,6 +7,8 @@ cat << EOF | sudo tee -a /etc/hosts
 EOF
 
 if sudo vgdisplay | grep -q lvm2; then
+   disk=$(sudo pvs --noheadings | head -n 1 | awk '{print $1}')
+   sudo pvresize $disk
    sudo lvextend -L 4G /dev/rootvg/lv_home -r || true
    sudo lvextend -L 4G /dev/rootvg/lv_tmp -r || true
 fi
--- a/etc/kayobe/environments/ci-aio/automated-setup.sh
+++ b/etc/kayobe/environments/ci-aio/automated-setup.sh
@@ -69,8 +69,6 @@ source kayobe-env --environment ci-aio
 
 kayobe control host bootstrap
 
-kayobe playbook run etc/kayobe/ansible/growroot.yml
-
 kayobe overcloud host configure
 
 kayobe overcloud service deploy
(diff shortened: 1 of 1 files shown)


Alex-Welsh:
Growroot is no longer required if pvresize is being run early on, so you can delete [this line](https://github.com/stackhpc/stackhpc-kayobe-config/pull/890/files#diff-bc3d997969357b16390ebf2a9aa7c65f86030f4eb5fc5323194c095e5d483c46R72)

Alex-Welsh (dismissed):
LGTM but two notes:

1. Your change could be one line:
  `sudo pvresize $(sudo pvs --noheadings | head -n 1 | awk '{print $1}')`

2. It's a bit of a shame we don't use growroot anymore, but I think it's more important to keep this script clean and consistent than to try and have some complex conditional that does both a pvresize and growroot.

seunghun1ee:
> 1. Your change could be one line:
>    `sudo pvresize $(sudo pvs --noheadings | head -n 1 | awk '{print $1}')`

Yeah I agree. Looks like setting a variable is not needed here.

Alex-Welsh (dismissed):

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -7,6 +7,7 @@ cat << EOF | sudo tee -a /etc/hosts
 EOF
 
 if sudo vgdisplay | grep -q lvm2; then
```
markgoddard, line 9: I wonder why this is necessary when running this script, but not in CI.

Alex-Welsh: Different images? We're using the Ubuntu-20.04 image but CI uses overcloud-ubuntu-focal-*

Alex-Welsh: I've also seen it do different things when running things as different users in the past. Maybe it's something about when we run things as `ubuntu` vs `root` for example

markgoddard: We should switch to use the release train images.

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -67,8 +68,6 @@ source kayobe-env --environment ci-aio
 
 kayobe control host bootstrap
 
-kayobe playbook run etc/kayobe/ansible/growroot.yml
```
markgoddard: Please keep growroot - it does some additional things

etc/kayobe/environments/ci-aio/automated-setup.sh
markgoddard: Running through an aio on a CS8 release train image now. This part is required because we clone repos and create virtualenvs in /home, whereas in CI that's done on the CI runner.

markgoddard (changes requested):

etc/kayobe/environments/ci-aio/automated-setup.sh
seunghun1ee: Okay, the change is now reverted

markgoddard (approved):

seunghun1ee:
@markgoddard What do you think of also making PR with this change for `stackhpc/zed` and `stackhpc/2023.1`?

markgoddard:
> @markgoddard What do you think of also making PR with this change for `stackhpc/zed` and `stackhpc/2023.1`?

We merge changes forwards in this repo, so they will get fixed automatically. I can propose some merges.

seunghun1ee:
> > @markgoddard What do you think of also making PR with this change for `stackhpc/zed` and `stackhpc/2023.1`?
> 
> We merge changes forwards in this repo, so they will get fixed automatically. I can propose some merges.

Thanks Mark

PR #892, opened 2024-01-22
Try to run in container.

MaxBed4d:


MaxBed4d:
Old and isn't required anymore, was part of the initial deployment to get dind to work with AiO.

PR #904, opened 2024-01-31
Update Kolla container images for Ubuntu Jammy Zed

seunghun1ee:
Tested during Ubuntu jammy Yoga to Zed migration with tempest. No alert was fired at Prometheus.

### DIFF
--- a/etc/kayobe/kolla-image-tags.yml
+++ b/etc/kayobe/kolla-image-tags.yml
@@ -5,19 +5,19 @@
 kolla_image_tags:
   openstack:
     rocky-9: zed-rocky-9-20230921T153510
-    ubuntu-jammy: zed-ubuntu-jammy-20230921T153510
+    ubuntu-jammy: zed-ubuntu-jammy-20240129T151534
   bifrost:
     rocky-9: zed-rocky-9-20230927T142529
-    ubuntu-jammy: zed-ubuntu-jammy-20231101T132522
+    ubuntu-jammy: zed-ubuntu-jammy-20240129T151534
   ovn:
     rocky-9: zed-rocky-9-20230925T132313
-    ubuntu-jammy: zed-ubuntu-jammy-20230821T155947
+    ubuntu-jammy: zed-ubuntu-jammy-20240129T151534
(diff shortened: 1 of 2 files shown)


Alex-Welsh (changes requested):
New kolla tags need a release note

Alex-Welsh (dismissed):
Thanks, LGTM as long as AIOs pass

MoteHue:
Going to be using these next week given the OVN Jammy AIO passed, promotion is running here: https://github.com/stackhpc/stackhpc-release-train/actions/runs/7759396949/job/21163343101

etc/kayobe/kolla-image-tags.yml
```diff
@@ -5,19 +5,19 @@
 kolla_image_tags:
   openstack:
     rocky-9: zed-rocky-9-20230921T153510
-    ubuntu-jammy: zed-ubuntu-jammy-20230921T153510
+    ubuntu-jammy: zed-ubuntu-jammy-20240129T151534
   bifrost:
     rocky-9: zed-rocky-9-20230927T142529
-    ubuntu-jammy: zed-ubuntu-jammy-20231101T132522
+    ubuntu-jammy: zed-ubuntu-jammy-20240129T151534
```
Alex-Welsh: I don't think you need to define any of the tags that are the same as the default openstack tag

seunghun1ee: Should I remove all `ubuntu-jammy` entries that are same to default tag and leave only `rocky-9` tags?

Alex-Welsh: Yes

Alex-Welsh: Actually, it may be worth waiting until Will's PR can merge, since it will conflict with this one
https://github.com/stackhpc/stackhpc-kayobe-config/pull/910

seunghun1ee: It's done

Alex-Welsh (approved):

markgoddard (approved):

PR #905, opened 2024-01-31
Update docs to point out admin port change

technowhizz:
Points out the admin port change in keystone and mentions changing the cephadm.yml RGW config file as an example

### DIFF
--- a/doc/source/operations/upgrading.rst
+++ b/doc/source/operations/upgrading.rst
@@ -84,6 +84,16 @@ system. If you are using a Neutron plugin other than ML2/OVN, set
 For new deployments using OVN, see
 :kolla-ansible-doc:`reference/networking/neutron.html#ovn-ml2-ovn`.
 
+Keystone admin port change
+--------------------------
+
+The Keystone admin port has changed to be the same as the public port (5000).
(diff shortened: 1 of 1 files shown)


Alex-Welsh:
Does it delete the endpoint in openstack or does it just stop working? i.e. if you did an `openstack endpoint list`would it show up?

technowhizz:
Nope the endpoint is still there. It's the same port as the public (and internal) port

Alex-Welsh:
> Nope the endpoint is still there. It's the same port as the public (and internal) port

Might be safest to add a step to remove it in that case. I'm not 100% sure but I think we can remove all the admin endpoints once we're on Zed.

markgoddard:
> > Nope the endpoint is still there. It's the same port as the public (and internal) port
> 
> Might be safest to add a step to remove it in that case. I'm not 100% sure but I think we can remove all the admin endpoints once we're on Zed.

I started [this patch](https://review.opendev.org/c/openstack/kolla-ansible/+/904724) to remove the admin endpoint in bobcat, and was planning to backport to antelope. I'll see if I can push it along.

markgoddard (approved):

PR #906, opened 2024-02-01
Fix Wazuh docs for multiple environments

sjpb:


Alex-Welsh:
LGTM but we should merge it into `stackhpc/yoga` first then sync upwards

sjpb:
Ah bother, I'll start again!

Alex-Welsh:
You might be able to recover it with:
```bash
git rebase stackhpc/yoga
git push -f
```
then point the PR at `stackhpc/yoga`

But rebases are basically deterministic in my experience so who knows

PR #908, opened 2024-02-02
Zed: Update to Rocky 9.3 based kolla images

jovial:
These have been rebuilt using Rocky 9.3 snapshots.

Builds:

neutron: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/7757168800
others: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/7754948585

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -12,30 +12,25 @@ kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %
 
 kayobe_image_tags:
   openstack:
-    rocky: zed-rocky-9-20230921T153510
+    rocky: zed-rocky-9-20240202T105829
     ubuntu: zed-ubuntu-jammy-20230921T153510
   bifrost:
-    rocky: zed-rocky-9-20230927T142529
     ubuntu: zed-ubuntu-jammy-20230927T154443
   ovn:
-    rocky: zed-rocky-9-20230925T132313
     ubuntu: zed-ubuntu-jammy-20230821T155947
   cloudkitty:
-    rocky: zed-rocky-9-20231114T124701
     ubuntu: zed-ubuntu-jammy-20231114T124701
   neutron:
-    rocky: zed-rocky-9-20231115T094053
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla/globals.yml
```diff
@@ -12,22 +12,17 @@ kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %
...
-    rocky: zed-rocky-9-20230925T132313
     ubuntu: zed-ubuntu-jammy-20230821T155947
   cloudkitty:
-    rocky: zed-rocky-9-20231114T124701
     ubuntu: zed-ubuntu-jammy-20231114T124701
   neutron:
-    rocky: zed-rocky-9-20231115T094053
     ubuntu: zed-ubuntu-jammy-20231115T094053
   opensearch:
-    rocky: zed-rocky-9-20231214T095452
```
Alex-Welsh: Removing these lines will cause the tag lookup (lines 34-38) to break. I think in the past we've resolved this by either just duplicating the tag, or adding a default value to the lookup

etc/kayobe/kolla/globals.yml
```diff
@@ -12,22 +12,17 @@ kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %
...
     ubuntu: zed-ubuntu-jammy-20231214T095452
 
 openstack_tag: "{% raw %}{{ kayobe_image_tags['openstack'][kolla_base_distro] }}{% endraw %}"
```
jovial: Hmm, no default on these now :(

etc/kayobe/kolla/globals.yml
markgoddard: If only we had merged https://github.com/stackhpc/stackhpc-kayobe-config/pull/663 already we wouldn't need to worry about this :D

jovial: Cheers, I've add the default value.

jovial: > If only we had merged #663 already we wouldn't need to worry about this :D

Should we try and get that merged instead?

markgoddard (approved):

PR #914, opened 2024-02-07
Ubuntu Jammy 2023.1 kolla image upgrade

seunghun1ee:
OVS images are not included in this new tag `2023.1-ubuntu-jammy-20240129T151608`
Due to the issue where not all images can be built at the same time. (See https://github.com/stackhpc/kayobe/pull/239)


seunghun1ee:
Updated neutron tag to include OVS images.
Now this branch should have all Kolla images.

markgoddard (approved):

PR #916, opened 2024-02-07
Enable persistent journal in Rocky

cityofships:
This is a quick win to match Ubuntu's default behaviour - system's journal always persisted to disk and available after reboots.

### DIFF
--- a/etc/kayobe/ansible/rsyslog.yml
+++ b/etc/kayobe/ansible/rsyslog.yml
@@ -11,26 +11,26 @@
   become: yes
 
   tasks:
-   - name: Ensure rsyslog is installed
-     package:
+  - name: Ensure rsyslog is installed
+    package:
       name: rsyslog
       state: present
 
-   - name: Ensure rsyslog is started and enabled
-     systemd:
+  - name: Ensure rsyslog is started and enabled
+    systemd:
       state: started
       enabled: yes
       name: rsyslog
 
-   - name: Update rsyslog configuration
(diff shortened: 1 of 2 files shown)


markgoddard (commented):
Do we need to consider the impact this will have on the lv_log volume? It may not have been sized to allow for a persistent journal

markgoddard:
Probably also needs a release note

cityofships:
> Do we need to consider the impact this will have on the lv_log volume? It may not have been sized to allow for a persistent journal

Storage usage defaults to 10% of the filesystem and 4g hard limit - with 20g configured currently we should be safe.

Adding a reno shortly.

markgoddard (commented):
LGTM, but would appreciate a second opinion - are there any potential side-effects, and is this the right place in which to do it?

markgoddard:
Should we wait on Will's upstream Kayobe patch and backport that instead?

jovial:
For context, I tried submitting this upstream here: https://review.opendev.org/c/openstack/kayobe/+/909685 as the podman jobs were failing when this directory did not exist (and it seemed like a good idea)!

Alex-Welsh:
Upstream PR is now merged. Backport [here](https://github.com/stackhpc/kayobe/pull/291)

cityofships:
> Upstream PR is now merged. Backport [here](https://github.com/stackhpc/kayobe/pull/291)

Fair enough

PR #923, opened 2024-02-08
Override kolla_base_distro_version

MoteHue:
When running the kayobe-automation CI on a Zed/Antelope cloud running Ubuntu host images, these variable overrides cause the kolla tags to template incorrectly. For example, a Jammy host gets ``kolla_base_distro_and_version = rocky-jammy``.

Given we will not be performing any host OS migrations on Zed (or Antelope), we can remove these overrides and use the default values based on ``os_distribution``.

Bug explanation:

These variables end up with the following values, meaning the CI picks up rocky but keeps jammy.

```
kolla_base_distro: '{{ ansible_facts.distribution | lower }}'
kolla_base_distro_and_version: '{{ kolla_base_distro }}-{{ kolla_base_distro_version}}'
kolla_base_distro_version: jammy
```

This is because the variables are first templated under the Kayobe group_vars based on os_distribution, so the version maps to jammy. ``kayobe/ansible/inventory/group/vars/all/kolla``

```
kolla_base_distro: "{{ os_distribution }}"

kolla_base_distro_version_default_map: {
  "centos": "stream9",
  "debian": "bullseye",
  "rocky": "9",
  "ubuntu": "jammy",
}

kolla_base_distro_version: "{{ kolla_base_distro_version_default_map[kolla_base_distro] }}"
```

Then, the base distro and the final variable are overwritten in kolla globals kayobe-config/etc/kayobe/kolla/globals.yml

```
kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"

kolla_base_distro_and_version: "{% raw %}{{ kolla_base_distro }}-{{ kolla_base_distro_version }}{% endraw %}"
```

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -10,6 +10,10 @@ enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names }}{% endraw %}"
 # This is necessary for os migrations where mixed clouds might be deployed
 kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"
 
+# Use facts so this is determined correctly when the control host OS differs
+# from os_distribuition.
+kolla_base_distro_version: "{% raw %}{{ ansible_facts.distribution_major_version }}{% raw %}"
+
 # Convenience variable for base distro and version string.
 kolla_base_distro_and_version: "{% raw %}{{ kolla_base_distro }}-{{ kolla_base_distro_version }}{% endraw %}"
 
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -12,7 +12,7 @@ kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %
 
 # Use facts so this is determined correctly when the control host OS differs
 # from os_distribuition.
-kolla_base_distro_version: "{% raw %}{{ ansible_facts.distribution_major_version }}{% raw %}"
+kolla_base_distro_version: "{% raw %}{{ ansible_facts.distribution_major_version }}{% endraw %}"
 
 # Convenience variable for base distro and version string.
 kolla_base_distro_and_version: "{% raw %}{{ kolla_base_distro }}-{{ kolla_base_distro_version }}{% endraw %}"
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -12,7 +12,7 @@ kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %
 
 # Use facts so this is determined correctly when the control host OS differs
 # from os_distribuition.
-kolla_base_distro_version: "{% raw %}{{ ansible_facts.distribution_major_version }}{% endraw %}"
+kolla_base_distro_version: "{% raw %}{{ kolla_base_distro_version_default_map[kolla_base_distro] }}{% endraw %}"
 
 # Convenience variable for base distro and version string.
 kolla_base_distro_and_version: "{% raw %}{{ kolla_base_distro }}-{{ kolla_base_distro_version }}{% endraw %}"


jackhodgkiss (dismissed):
LGTM.

etc/kayobe/kolla/globals.yml
```diff
@@ -6,13 +6,6 @@
 # non-overcloud hosts
 enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names }}{% endraw %}"
 
-# kolla_base_distro must be set here to be resolvable on a per-host basis
-# This is necessary for os migrations where mixed clouds might be deployed
-kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"
-
-# Convenience variable for base distro and version string.
-kolla_base_distro_and_version: "{% raw %}{{ kolla_base_distro }}-{{ kolla_base_distro_version }}{% endraw %}"
```
jovial, line 14: I believe these are used for the new auto-generated kolla image tags i.e https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2023.1/etc/kayobe/kolla-image-tags.yml#L7

And in particular:

https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2023.1/etc/kayobe/kolla/globals.yml#L19

so may need further changes.

Alex-Welsh:
I think this brings up a wider question of perhaps supporting Jammy-based automation images. I can imagine this won't be the last time we see a conflict like this. I know it's yet another thing to support but it's a question of what will be more effort. Either that or whack-a-bug with distro conflicts.

MoteHue:
> I think this brings up a wider question of perhaps supporting Jammy-based automation images. I can imagine this won't be the last time we see a conflict like this. I know it's yet another thing to support but it's a question of what will be more effort. Either that or whack-a-bug with distro conflicts.

Imo supporting Ubuntu in kayobe-automation is the correct long-term fix for this issue. However, as you say the work to do this isn't small, so I'd proposed this PR here so that the immediate upgrade cycle of Z+A can continue without every Ubuntu customer needing to be fixed. 

I need to look the new kolla tags stuff first, but the comment for `kolla_base_distro` suggests it only needs to be set for the RL9 migrations. Perhaps rather than removal, just setting it to os_distribution would be a sufficient fix.

MoteHue:
So for more context, this error was hit when running a kolla config-diff using a Kayobe image built with a RL9 kayobe-automation but an Ubuntu Jammy target system. 

```
TASK [common : Ensuring config directories exist]
...
{{ common_tag }}: {{ openstack_tag }}: {{ kolla_image_tags['openstack'][kolla_base_distro_and_version] }}: 'dict object' has no attribute 'rocky-jammy'. 'dict object' has no attribute 'rocky-jammy'.
```

markgoddard:
Discussed with @MoteHue IRL. I think this is caused by Kayobe's globals.yml template (`ansible/roles/kolla-ansible/templates/kolla/globals.yml`) setting `kolla_base_distro_version` based on Kayobe's variable of the same name, which references `os_version`. We can avoid this by explicitly using facts for the version in SKC's globals.yml template:

```
kolla_base_distro_version: "{{ ansible_facts.distribution_major_version }}"
```

While this will lead to localhost seeing `kolla_base_distro_and_version` as `rocky-9`, IIUC this will never be referenced during deployment, and is only required to successfully template out localhost's hostvars.

MoteHue:
> Discussed with @MoteHue IRL. I think this is caused by Kayobe's globals.yml template (`ansible/roles/kolla-ansible/templates/kolla/globals.yml`) setting `kolla_base_distro_version` based on Kayobe's variable of the same name, which references `os_version`. We can avoid this by explicitly using facts for the version in SKC's globals.yml template:
> 
> ```
> kolla_base_distro_version: "{{ ansible_facts.distribution_major_version }}"
> ```
> 
> While this will lead to localhost seeing `kolla_base_distro_and_version` as `rocky-9`, IIUC this will never be referenced during deployment, and is only required to successfully template out localhost's hostvars.

I've tested this solution in Bristol CI, it works.

etc/kayobe/kolla/globals.yml
```diff
@@ -10,6 +10,10 @@ enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names }}{% endraw %}"
 # This is necessary for os migrations where mixed clouds might be deployed
 kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"
 
+# Use facts so this is determined correctly when the control host OS differs
+# from os_distribuition.
+kolla_base_distro_version: "{{ ansible_facts.distribution_major_version }}"
```
markgoddard: It needs to be wrapped in `{% raw %}` like the one above to use the version of the remote host rather than the Ansible control host.

MoteHue: Missed this as my testing was limited to just the kolla config diff, now done.

etc/kayobe/kolla/globals.yml
```diff
@@ -10,6 +10,10 @@ enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names }}{% endraw %}"
 # This is necessary for os migrations where mixed clouds might be deployed
 kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"
 
+# Use facts so this is determined correctly when the control host OS differs
+# from os_distribuition.
+kolla_base_distro_version: "{% raw %}{{ ansible_facts.distribution_major_version }}{% raw %}"
```
markgoddard:
```suggestion
kolla_base_distro_version: "{% raw %}{{ ansible_facts.distribution_major_version }}{% endraw %}"
```

markgoddard (changes requested):

markgoddard (dismissed):

etc/kayobe/kolla/globals.yml
```diff
@@ -10,6 +10,10 @@ enable_docker_repo: "{% raw %}{{ 'overcloud' not in group_names }}{% endraw %}"
 # This is necessary for os migrations where mixed clouds might be deployed
 kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"
 
+# Use facts so this is determined correctly when the control host OS differs
+# from os_distribuition.
+kolla_base_distro_version: "{% raw %}{{ ansible_facts.distribution_major_version }}{% endraw %}"
```
markgoddard: Hmm, seems this is wrong:

```
''dict object'' has no attribute ''ubuntu-22'''
```

I think we actually need to do this:
```
kolla_base_distro_version: "{% raw %}{{ kolla_base_distro_version_default_map[kolla_base_distro] }}{% endraw %}"
```

MoteHue: Made the change. Tested this does still fix the kolla config diffs. Let's hope CI passes here too now :crossed_fingers:

markgoddard (approved):

PR #924, opened 2024-02-09
zed: yoga merge

markgoddard:
- CI: Update aio jobs to run on ARC
- Support credentials for custom DNF repositories
- CI: Use Ark package repositories to install packages
- CI: Switch from sms-lab-release to openstack cloud
- CI: Update aio VM flavor, network and subnet for Leafcloud
- CI: Update GitHub action versions for node20
- Revert hashicorp/setup-terraform action to v2 due to failure
- CI: Revert to ens3 interface for Leafcloud
- Add Apt auth configuration
- CI: Increase flavor to en1.large - some jobs were failing Tempest jobs
- CI: Run container image build job on ARC cluster
- Add package repository credentials to container image build
- Don't verify Apt repo CA initially when using HTTPS in container build

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -38,7 +38,7 @@ on:
       vm_flavor:
         description: Flavor for the all-in-one VM
         type: string
-        default: en1.medium
+        default: en1.large
       vm_network:
         description: Network for the all-in-one VM
         type: string
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -70,16 +70,25 @@ jobs:
   all-in-one:
     name: All in one
     if: inputs.if
-    runs-on: [self-hosted, stackhpc-kayobe-config-aio]
+    runs-on: arc-skc-aio-runner
     permissions: {}
     env:
       KAYOBE_ENVIRONMENT: ci-aio
       KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
       KAYOBE_IMAGE: ${{ inputs.kayobe_image }}
     steps:
+      - name: Install Package
+        uses: ConorMacBride/install-package@main
+        with:
+          apt: build-essential git unzip nodejs python3 python3-pip python3-wheel python-is-python3
+
       - uses: actions/checkout@v4
         with:
            submodules: true
+
+      - name: Make sure dockerd is running and test Docker.
+        run: |
+          docker run --rm hello-world
 
       - name: Output image tag
         id: image_tag
@@ -209,11 +218,11 @@ jobs:
       # The same tag may be reused (e.g. pr-123), so ensure we have the latest image.
       - name: Pull latest Kayobe image
         run: |
-          sudo docker image pull $KAYOBE_IMAGE
+          docker image pull $KAYOBE_IMAGE
 
       - name: Run growroot
         run: |
-          sudo -E docker run -t --rm \
+          docker run -t --rm \
             -v $(pwd):/stack/kayobe-automation-env/src/kayobe-config \
             -e KAYOBE_ENVIRONMENT -e KAYOBE_VAULT_PASSWORD -e KAYOBE_AUTOMATION_SSH_PRIVATE_KEY \
             $KAYOBE_IMAGE \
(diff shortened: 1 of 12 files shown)


.github/workflows/stackhpc-check-tags.yml
```diff
@@ -20,14 +20,15 @@ jobs:
...
-    runs-on: [self-hosted, stackhpc-kayobe-config-aio]
+    runs-on: arc-skc-aio-runner
     permissions: {}
     env:
       KAYOBE_ENVIRONMENT: ci-aio
       KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
       KAYOBE_IMAGE: ${{ inputs.kayobe_image }}
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout
```
Alex-Welsh, line 35: Might need to install/update git before this. https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/7842242364/job/21400456106?pr=924#step:2:28

Alex-Welsh:
There I was thinking our days of TF errors were over :disappointed:

.github/workflows/stackhpc-check-tags.yml
markgoddard: Done

markgoddard:
This is mostly working now, just waiting on upgrade jobs. The check container image tags job needs some fixes because the stackhpc-dev container repositories on Ark are "push repositories". It's non-voting though, so I suggest we merge and fix that later.

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -168,7 +177,17 @@ jobs:
           OS_APPLICATION_CREDENTIAL_SECRET: ${{ secrets.OS_APPLICATION_CREDENTIAL_SECRET }}
 
       - name: Terraform Apply
-        run: terraform apply -auto-approve
+        run: |
```
Alex-Welsh, line 180: Since this is just a command, we could try this retry action:
https://github.com/nick-fields/retry

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -132,23 +127,19 @@ jobs:
...
-          kayobe seed host configure
-        env:
-          KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
+          pip install --user docker
 
-      - name: Prune local Kolla container images over 1 week old
+      - name: Configure localhost as a seed
         run: |
-          sudo docker image prune --all --force --filter until=168h --filter="label=kolla_version"
+          cat > src/kayobe-config/etc/kayobe/environments/ci-builder/inventory/hosts << EOF
```
Alex-Welsh, line 137: Is this not something we can do in SKC permanently?

etc/kayobe/kolla.yml
```diff
@@ -250,12 +250,26 @@ kolla_build_blocks:
         sed -i -e '/\[{{ repo.tag }}\]/,/^\[/ s/^\(mirrorlist *=.*\)/#\1/g' \
                -e '/\[{{ repo.tag }}\]/,/^\[/ s/^[# ]*\(baseurl *=.*\)/#\1/g' \
                -e '/\[{{ repo.tag }}\]/,/^\[/ s/^[# ]*\(metalink *=.*\)/#\1/g' \
+    {% if stackhpc_repo_mirror_username is truthy %}
+               -e '/\[{{ repo.tag }}\]/,/^\[/ s|^\(name.*\)|\1\nusername={{ stackhpc_repo_mirror_username }}|' \
+               -e '/\[{{ repo.tag }}\]/,/^\[/ s|^\(name.*\)|\1\npassword={{ stackhpc_repo_mirror_password }}|' \
```
Alex-Welsh, line 255: I'm not sure if there's a good way around it, but this could create some hard-to-debug errors if the password contains a `|`

.github/workflows/stackhpc-all-in-one.yml
Alex-Welsh: Just to clarify, I'm happy to merge as is. I don't think there's a huge benefit to using the custom action, but worth considering as an option.

markgoddard: I'm generally a bit wary of unofficial custom actions because they can have access to some sensitive data.

.github/workflows/stackhpc-container-image-build.yml
markgoddard: Possibly eventually. Currently the host image build workflow doesn't work in ARC, so we'd need to use the old style workflow which hasn't been tested with this.

etc/kayobe/kolla.yml
markgoddard: We could escape it, but a fix should go to stackhpc/yoga. @jovial had some valid concerns around the passwords being in lower layers of the images, even though we remove them, so we may need to revisit this.

Alex-Welsh (approved):
Couple of threads to follow up on but LGTM as a fix for our AIO issues

PR #926, opened 2024-02-09
Allow aio jobs to be cancelled (yoga)

MaxBed4d:
Allow aio jobs to be cancelled


MaxBed4d:
This is now successfully capable of cancelling any of the job's tasks upon cancellation request and if the pull request workflow is cancelled the AiO job will also be cancelled. 
The AiO job can be cancelled during executed steps.

Alex-Welsh (dismissed):
LGTM

markgoddard (approved):

markgoddard:
Verified that both the build-kayobe-image and aio workflows can be cancelled. After cancelling an aio workflow the VM instance is destroyed.

PR #932, opened 2024-02-09
Correct OpenSearch migration instructions

MoteHue:
Because we set ``kolla_enable_central_logging: true``, ``kolla_enable_elsticsearch`` is also set to true, as per the Kayobe defaults:

```
kolla_enable_elasticsearch: "{{ not kolla_base_distro == 'rocky' and (kolla_enable_central_logging | bool or kolla_enable_osprofiler | bool or kolla_enable_skydive | bool or kolla_enable_monasca | bool) }}"
```

Update the instructions to disbale Elasticsearch explicitly.

### DIFF
--- a/doc/source/operations/rocky-linux-9.rst
+++ b/doc/source/operations/rocky-linux-9.rst
@@ -161,7 +161,7 @@ Elasticsearch/Kibana should be migrated to OpenSearch.
 - Read the `Kolla Ansible OpenSearch migration
   docs <https://docs.openstack.org/kolla-ansible/yoga/reference/logging-and-monitoring/central-logging-guide-opensearch.html#migration>`__
 - If necessary, take a backup of the Elasticsearch data.
-- Ensure ``kolla_enable_elasticsearch`` is unset in ``etc/kayobe/kolla.yml``
+- Ensure ``kolla_enable_elasticsearch`` is set to false in ``etc/kayobe/kolla.yml``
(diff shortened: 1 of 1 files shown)


technowhizz (approved):
Lovely 👍

dougszumski:
> Because we set `kolla_enable_central_logging: true`, `kolla_enable_elsticsearch` is also set to true, as per the Kayobe defaults:
> 
> ```
> kolla_enable_elasticsearch: "{{ not kolla_base_distro == 'rocky' and (kolla_enable_central_logging | bool or kolla_enable_osprofiler | bool or kolla_enable_skydive | bool or kolla_enable_monasca | bool) }}"
> ```
> 
> Update the instructions to disbale Elasticsearch explicitly.

Hmm, shouldn't `kolla_enable_elasticsearch` evaluate to False for Rocky in the above? And for other distros, in Yoga, they should be using Elastic? 

See: https://review.opendev.org/c/openstack/kolla-ansible/+/875729

technowhizz:
We hit an issue with Ubuntu 😅

MoteHue:
> And for other distros, in Yoga, they should be using Elastic?

We need to run the OpenSearch migration in Yoga prior to upgrading to Zed, regardless of distro. At worst, these instructions give us a redundant instance of enable_opensearch that would be deleted when the default is changed to OpenSearch in Zed anyway. As Dawud said, the instructions are currently wrong for Ubuntu systems

dougszumski (approved):
Thanks for clarifying

PR #934, opened 2024-02-09
Bump stackhpc.kayobe_workflows to version 1.0.3

MoteHue:
We hit this error in a control host bootstrap otherwise.

```
ERROR! Failed to resolve the requested dependencies map. Could not satisfy the following requirements:
* stackhpc.kayobe_workflows:1.0.2 (direct request)
* stackhpc.kayobe_workflows:1.0.3 (dependency of "virtual collection Git repo")
```

### DIFF
--- a/etc/kayobe/ansible/requirements.yml
+++ b/etc/kayobe/ansible/requirements.yml
@@ -11,7 +11,7 @@ collections:
   - name: stackhpc.hashicorp
     version: 2.4.0
   - name: stackhpc.kayobe_workflows
-    version: 1.0.2
+    version: 1.0.3
 roles:
   - src: stackhpc.vxlan
   - name: ansible-lockdown.rhel8_cis


Alex-Welsh:
Does this only affect antelope?

MoteHue:
> Does this only affect antelope?

Ah good catch, I hit this in Zed too. Presumably Yoga is affected as well

markgoddard:
I'm not really sure what that error means. What is a "virtual collection Git repo"?

MoteHue:
I think we're hitting this issue: https://github.com/ansible/ansible/issues/79168
It isn't fixed until ansible core 2.15.5: https://github.com/ansible/ansible/pull/81606

markgoddard:
> I think we're hitting this issue: [ansible/ansible#79168](https://github.com/ansible/ansible/issues/79168) It isn't fixed until ansible core 2.15.5: [ansible/ansible#81606](https://github.com/ansible/ansible/pull/81606)

That looks like it's more about installing collections from source, but we're using galaxy here. I don't mind merging this, since it's not a controversial change, but I'm unsure why we're hitting it.

PR #935, opened 2024-02-09
Adds alerts for software raid failures

jovial:
See: https://github.com/prometheus/node_exporter/blob/master/docs/node-mixin/alerts/alerts.libsonnet

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/system.rules
+++ b/etc/kayobe/kolla/config/prometheus/system.rules
@@ -104,4 +104,24 @@ groups:
     annotations:
       summary: Host conntrack limit (instance {{ $labels.instance }})
       description: "The number of conntrack is approaching limit"
+
+  - alert: NodeRAIDDegraded
+    expr: |
+      node_md_disks_required{job="node",device!=""} - ignoring (state) (node_md_disks{state="active",job="node",device!=""}) > 0
+    for: "15m"
+    labels:
+      severity: critical
+    annotations:
+      description: "RAID array '{{ $labels.device }}' at {{ $labels.instance }} is in degraded state due to one or more disks failures. Number of spare drives is insufficient to fix issue automatically."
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla/config/prometheus/system.rules
```diff
@@ -104,4 +104,24 @@ groups:
...
+      severity: critical
+    annotations:
+      description: "RAID array '{{ $labels.device }}' at {{ $labels.instance }} is in degraded state due to one or more disks failures. Number of spare drives is insufficient to fix issue automatically."
+      summary: "RAID Array is degraded."
+
+  - alert: NodeRAIDDiskFailure
+    expr: |
+      node_md_disks{state="failed",job="node",device!=""} > 0
+    labels:
+      severity: warning
```
Alex-Welsh, line 122: Not sure whether this should be warning or critical. I can see the argument for it being a warning but critical might mean it gets caught by more filters

Alex-Welsh (dismissed):

etc/kayobe/kolla/config/prometheus/system.rules
dougszumski: For the sake of keeping in sync with the upstream source, I'm fine with leaving as-is. We at least have the critical alert above for when immediate action needs to be taken. Warnings should generally be things that don't need to wake people up in the night.

dougszumski (dismissed):

releasenotes/notes/adds-mdraid-alerts-309fb79e61389325.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Adds alerts for software raid f`ailures.
```
jovial:
```suggestion
    Adds alerts for software raid failures.
```

markgoddard (approved):

PR #940, opened 2024-02-13
CI: Fail aio job when Terraform reaches max attempts

markgoddard:
Previously the script exited 0 at the end of the loop, then failed in a later step.

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -158,13 +158,15 @@ jobs:
           for attempt in $(seq 5); do
               if terraform apply -auto-approve; then
                   echo "Created infrastructure on attempt $attempt"
-                  break
+                  exit 0
               fi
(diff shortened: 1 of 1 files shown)


.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -158,13 +158,15 @@ jobs:
...
                   echo "Created infrastructure on attempt $attempt"
-                  break
+                  exit 0
               fi
               echo "Failed to create infrastructure on attempt $attempt"
               sleep 10
               terraform destroy -auto-approve
               sleep 60
           done
+          echo "Failed to create infrastructure after $attempt attempts"
```
Alex-Welsh, line 168: $attempt is referenced here outside the loop. Does it persist?

Alex-Welsh: I've now done what I should've done in the first place and just tested it. It does indeed work

Alex-Welsh (approved):

PR #942, opened 2024-02-14
CI: Reduce aio flavor to en1.medium (8G)

markgoddard:
- Reduce Elasticsearch/OpenSearch heap size to 200m in ci-aio env
- Disable Heat in ci-aio env
- Reduce aio flavor to en1.medium


markgoddard:
This has been run 3 times, with 10 aios in each run. One failed, due to NTP not being synchronised on Ubuntu Jammy. This is an unrelated issue I've seen elsewhere. I think we're good to go.

mnasiadka (approved):

PR #943, opened 2024-02-14
CI: Use skc-ci-aio user for aio jobs

markgoddard:
This user only has read-only access to the package and container
repositories, so is safer than using the release-train-ci user which has
read/write permissions.

For the container image build job we can use the skc-ci-aio user to
access the package repositories, but must use the release-train-ci user
to push container images.

### DIFF
--- a/etc/kayobe/environments/ci-aio/stackhpc-ci.yml
+++ b/etc/kayobe/environments/ci-aio/stackhpc-ci.yml
@@ -19,8 +19,14 @@ resolv_is_managed: false
 # Build and deploy the development Pulp service repositories.
 # Use Ark's package repositories to install packages.
 stackhpc_repo_mirror_url: "{{ stackhpc_release_pulp_url }}"
-stackhpc_repo_mirror_username: "{{ stackhpc_docker_registry_username }}"
-stackhpc_repo_mirror_password: "{{ stackhpc_docker_registry_password }}"
+stackhpc_repo_mirror_username: "skc-ci-aio"
+stackhpc_repo_mirror_password: !vault |
+  $ANSIBLE_VAULT;1.1;AES256
+  31386366383365666135336331663635396237623139306362633933636233613765663731666338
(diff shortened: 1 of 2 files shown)


Alex-Welsh:
If ci-builder needs to be able to push, why not just use the release train user for it?

markgoddard:
> If ci-builder needs to be able to push, why not just use the release train user for it?

It uses the release train user as the docker registry user, but not for package repos.

markgoddard:
Fixed up trailing newline in password :facepalm:

markgoddard:
Looks like container access is not working. Doing a bit of local debugging, the skc-ci-aio user cannot pull images in the stackhpc-dev namespace, even though it is in the `container.namespace.consumers.stackhpc-dev` group in Pulp. If I add the `container.namespace.collaborators.stackhpc-dev` group it still does not work, but if I add `container.namespace.owners.stackhpc-dev` then it does. This is an improvement over using the release-train-ci user, which has  lot of roles assigned to it, however it is not ideal.

markgoddard:
> Looks like container access is not working. Doing a bit of local debugging, the skc-ci-aio user cannot pull images in the stackhpc-dev namespace, even though it is in the `container.namespace.consumers.stackhpc-dev` group in Pulp. If I add the `container.namespace.collaborators.stackhpc-dev` group it still does not work, but if I add `container.namespace.owners.stackhpc-dev` then it does. This is an improvement over using the release-train-ci user, which has lot of roles assigned to it, however it is not ideal.

This issue does not seem to affect the stackhpc namespace. The main difference is that it does not contain container-push repositories.

markgoddard:
Manually fixed permissions in Ark with the following command:

```
pulp container namespace role add --role container.containernamespace_consumer --group container.namespace.consumers.stackhpc-dev --name stackhpc-dev
```

This is necessary due to [RBAC changes](https://docs.pulpproject.org/pulp_container/rbac/migrating-perms-to-roles.html) in pulp_container 2.11

Alex-Welsh (approved):

PR #945, opened 2024-02-16
Stop warning about invalid group name characters

priteau:
This should silence the following Ansible warning [1]:

    [WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details

[1] https://docs.ansible.com/ansible/latest/reference_appendices/config.html#transform-invalid-group-chars


Alex-Welsh (approved):

priteau:
@markgoddard Could I please have your opinion on this change? While it could make us miss a deprecation warning, I don't believe the Ansible developers are planning to change this behaviour anytime soon given how much software out there uses invalid characters in group names?

markgoddard (approved):

PR #818, opened 2023-12-06
Fix growroot when using software raid

jovial:
Using @markgoddard's suggestion from: https://github.com/stackhpc/stackhpc-kayobe-config/pull/770#issuecomment-1814402925

PR #947, opened 2024-02-17
Fix growroot playbook for NVMe devices

priteau:
The `disk_tmp` variable uses a device path rather than a device name.

### DIFF
--- a/etc/kayobe/ansible/growroot.yml
+++ b/etc/kayobe/ansible/growroot.yml
@@ -77,7 +77,7 @@
           vars:
             pv: "{{ pvs.stdout | from_json }}"
             disk_tmp: "{{ pv.report[0].pv[0].pv_name[:-1] }}"
-            disk: "{{ disk_tmp[:-1] if disk_tmp[-1] == 'p' and disk_tmp[:4] == 'nvme' else disk_tmp }}"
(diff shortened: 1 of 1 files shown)


cityofships:
Somewhat duplicates https://github.com/stackhpc/stackhpc-kayobe-config/pull/818

etc/kayobe/ansible/growroot.yml
```diff
@@ -77,7 +77,7 @@
           vars:
             pv: "{{ pvs.stdout | from_json }}"
             disk_tmp: "{{ pv.report[0].pv[0].pv_name[:-1] }}"
-            disk: "{{ disk_tmp[:-1] if disk_tmp[-1] == 'p' and disk_tmp[:4] == 'nvme' else disk_tmp }}"
+            disk: "{{ disk_tmp[:-1] if disk_tmp[-1] == 'p' and disk_tmp[:9] == '/dev/nvme' else disk_tmp }}"
```
markgoddard, line 80: I ran the pvs command on my laptop and `pv_name` is `/dev/mapper/nvme0n1p3_crypt`. Presumably because it's an encrypted disk.

priteau: But I would think we are unlikely to encounter this naming scheme on one of our deployments when using our overcloud DIB image.

Could we merge this and fix more issues in follow-up changes?

markgoddard (approved):

PR #948, opened 2024-02-19
docs: guide for migrating to containerized libvirt in R8/R9 migration

bbezak:


doc/source/operations/rocky-linux-9.rst
```diff
@@ -388,6 +388,29 @@ The possible batches depend on a number of things:
 * available spare hypervisor capacity
 * sizes of groups of compatible hypervisors
 
+Transition to Containerized Libvirt
```
markgoddard: It would be better if this was inserted into the correct location within the existing procedure. We'd at least want to drain the node first.

markgoddard (changes requested):

doc/source/operations/rocky-linux-9.rst
bbezak: moved it before host configure and added note

markgoddard (approved):

PR #953, opened 2024-02-21
Make ci-builder to build both OVS and OVN images on container image build

seunghun1ee:
As upstream kayobe change https://review.opendev.org/c/openstack/kayobe/+/907721 was merged, a condition for using ``kolla_build_neutron_ovs`` was added.
``kolla_build_neutron_ovs`` is set to true for ``kayobe overcloud container image build`` when no regex is given

### DIFF
--- a/.github/workflows/stackhpc-container-image-build.yml
+++ b/.github/workflows/stackhpc-container-image-build.yml
@@ -171,9 +171,6 @@ jobs:
           if ${{ inputs.push }} == 'true'; then
             args="$args --push"
           fi
-          if [[ "${{ github.event.inputs.regexes }}" == "" ]]; then
-            args="$args -e kolla_build_neutron_ovs=true"
-          fi
           source venvs/kayobe/bin/activate &&
           source src/kayobe-config/kayobe-env --environment ci-builder &&
           kayobe overcloud container image build $args
--- a/.github/workflows/stackhpc-container-image-build.yml
+++ b/.github/workflows/stackhpc-container-image-build.yml
@@ -171,6 +171,9 @@ jobs:
           if ${{ inputs.push }} == 'true'; then
             args="$args --push"
           fi
+          if [[ ${{ github.event.inputs.regexes }} == "" ]]; then
+            args="$args -e kolla_build_neutron_ovs=true"
+          fi
           source venvs/kayobe/bin/activate &&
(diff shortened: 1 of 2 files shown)


.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -171,6 +171,9 @@ jobs:
           if ${{ inputs.push }} == 'true'; then
             args="$args --push"
           fi
+          if [[ "${{ github.event.inputs.regexes }}" == "" ]]; then
+            args="$args -e kolla_build_neutron_ovs=true"
```
markgoddard: Could you instead set this in `etc/kayobe/environments/ci-builder/stackhpc-ci.yml`?

seunghun1ee: Will it be something like

seunghun1ee: Will it be something like
```
kolla_build_neutron_ovs: {{ github.event.inputs.regexes == "" | default(omit) }}
```
?

markgoddard: No, Kayobe/Ansible can't see the GitHub variables. It can be `true` unconditionally.

seunghun1ee: Can it? I thought this needs to be omitted if user defines the regex.
Wouldn't it build all neutron images even though user doesn't include it in regex if ``kolla_build_neutron_ovs`` is set to true unconditionally?

markgoddard: If a regex is specified it will override the big one defined in kayobe.

https://opendev.org/openstack/kayobe/src/branch/master/ansible/container-image-build.yml#L14

seunghun1ee: I found https://opendev.org/openstack/kayobe/src/branch/master/kayobe/cli/commands.py#L1716-L1749
It shows https://opendev.org/openstack/kayobe/src/branch/master/ansible/inventory/group_vars/all/kolla#L209 is not used when regex is given by user.
I'll change to define ``kolla_build_neutron_ovs: true`` unconditionally at ``etc/kayobe/environments/ci-builder/stackhpc-ci.yml``

markgoddard (approved):

PR #957, opened 2024-02-22
Update HAProxy dashboard and alerts

dougszumski:
Kolla has switched to using the built in HAProxy metrics source [1].

The new metrics source is not directly compatible, and therefore we need to update dashboards and alerts relying on the old metrics.

In a few cases, equivalent metrics were not available, so the plots have been removed.

[1] https://review.opendev.org/c/openstack/kolla-ansible/+/877118

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/haproxy.rules
+++ b/etc/kayobe/kolla/config/prometheus/haproxy.rules
@@ -6,17 +6,8 @@ groups:
 - name: HAProxy
   rules:
 
-  - alert: HaproxyDown
-    expr: haproxy_up == 0
-    for: 1m
-    labels:
-      severity: critical
-    annotations:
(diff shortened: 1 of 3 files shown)


etc/kayobe/kolla/config/prometheus/haproxy.rules
```diff
@@ -6,17 +6,8 @@ groups:
 - name: HAProxy
   rules:
 
-  - alert: HaproxyDown
-    expr: haproxy_up == 0
```
dougszumski, line 10: cAdvisor should spot a container down - we no longer have this metric because the exporter is built in.

technowhizz (approved):
Can confirm that this works and the dashboards and alerts look good. Thank you @dougszumski

PR #962, opened 2024-02-27
Improved AIO deployment script

Alex-Welsh:
The automated AIO deployment script has been improved in several ways:

The password check is now the first command to run
No user intervention is required to deploy to a non-lvm image
The python3-venv package is installed on Ubuntu hosts
The purge-command-not-found playbook is automatically run before host configuration

The script should now work on Leafcloud images, without any intervention

### DIFF
--- a/etc/kayobe/environments/ci-aio/automated-setup.sh
+++ b/etc/kayobe/environments/ci-aio/automated-setup.sh
@@ -2,30 +2,35 @@
 
 set -eux
 
-cat << EOF | sudo tee -a /etc/hosts
-10.205.3.187 pulp-server pulp-server.internal.sms-cloud
-EOF
-
-if sudo vgdisplay | grep -q lvm2; then
-   sudo pvresize $(sudo pvs --noheadings | head -n 1 | awk '{print $1}')
-   sudo lvextend -L 4G /dev/rootvg/lv_home -r || true
-   sudo lvextend -L 4G /dev/rootvg/lv_tmp -r || true
-fi
-
 BASE_PATH=~
 KAYOBE_BRANCH=stackhpc/yoga
 KAYOBE_CONFIG_BRANCH=stackhpc/yoga
+KAYOBE_AIO_LVM=true
 
 if [[ ! -f $BASE_PATH/vault-pw ]]; then
     echo "Vault password file not found at $BASE_PATH/vault-pw"
     exit 1
 fi
 
+if sudo vgdisplay | grep -q lvm2; then
+   sudo pvresize $(sudo pvs --noheadings | head -n 1 | awk '{print $1}')
+   sudo lvextend -L 4G /dev/rootvg/lv_home -r || true
+   sudo lvextend -L 4G /dev/rootvg/lv_tmp -r || true
+elif $KAYOBE_AIO_LVM; then
+   echo "This environment is only designed for LVM images. If possible, switch to an LVM image.
+   To ignore this warning, set KAYOBE_AIO_LVM to false in this script."
+   exit 1
+fi
+
+cat << EOF | sudo tee -a /etc/hosts
+10.205.3.187 pulp-server pulp-server.internal.sms-cloud
+EOF
+
 if type dnf; then
     sudo dnf -y install git
 else
     sudo apt update
-    sudo apt -y install gcc git libffi-dev python3-dev python-is-python3
+    sudo apt -y install gcc git libffi-dev python3-dev python-is-python3 python3-venv
 fi
 
 cd $BASE_PATH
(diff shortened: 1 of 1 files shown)


etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -2,39 +2,42 @@
 
 set -eux
 
+if [[ ! -f $BASE_PATH/vault-pw ]]; then
```
markgoddard, line 10: Reference to BASE_PATH before definition will blow up with `set -u`

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -2,39 +2,42 @@
...
 
-if sudo vgdisplay | grep -q lvm2; then
-   sudo pvresize $(sudo pvs --noheadings | head -n 1 | awk '{print $1}')
-   sudo lvextend -L 4G /dev/rootvg/lv_home -r || true
-   sudo lvextend -L 4G /dev/rootvg/lv_tmp -r || true
-fi
-
 BASE_PATH=~
 KAYOBE_BRANCH=stackhpc/yoga
 KAYOBE_CONFIG_BRANCH=stackhpc/yoga
```
markgoddard: Move to the top, since they may require modification?

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -2,39 +2,42 @@
...
+pushd src
+[[ -d kayobe ]] || git clone https://github.com/stackhpc/kayobe.git -b $KAYOBE_BRANCH
+[[ -d kayobe-config ]] || git clone https://github.com/stackhpc/stackhpc-kayobe-config kayobe-config -b $KAYOBE_CONFIG_BRANCH
+popd
+
+if sudo vgdisplay | grep -q lvm2; then
+   sudo pvresize $(sudo pvs --noheadings | head -n 1 | awk '{print $1}')
+   sudo lvextend -L 4G /dev/rootvg/lv_home -r || true
+   sudo lvextend -L 4G /dev/rootvg/lv_tmp -r || true
+else
```
markgoddard: I'm not sure exactly at what point it fails, but this was done early to avoid filling the small default LVs.

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -2,39 +2,42 @@
...
+   rm $BASE_PATH/src/kayobe-config/etc/kayobe/environments/ci-aio/inventory/group_vars/controllers/lvm.yml
+   sed -i -e '/controller_lvm_groups/,+2d' $BASE_PATH/src/kayobe-config/etc/kayobe/environments/ci-aio/controllers.yml
```
markgoddard, line 45: An alternative argument is that running on a non-LVM (AKA non-standard) image is a user error and should fail without explicit intervention?

markgoddard (changes requested):

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: :facepalm: Fixed

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: Done

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: I was hoping for one `if` statement, so wanted this after the kayobe-config clone. I assumed it would be small enough (55MB for kayobe, 26MB for kayobe-config) but it may be too tight. I'll re-arrange.

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: A compromise: We `exit 1` early on and provide instructions for how to make the kayobe-config changes with the caveat that it isn't strictly supported.

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -2,39 +2,48 @@
...
+   sudo lvextend -L 4G /dev/rootvg/lv_home -r || true
+   sudo lvextend -L 4G /dev/rootvg/lv_tmp -r || true
+elif $KAYOBE_AIO_LVM; then
+   echo "This environment is only designed for LVM images. If possible, switch to an LVM image.
+   
+   While it is not recommended, the environment can be adapted by:
+   Removing the lvm.yml group vars file in the ci-aio environment
+   Unsetting controller_lvm_groups in controllers.yml in the ci-aio environment
+   setting KAYOBE_AIO_LVM to false in this script."
+   exit 1
```
markgoddard, line 22: You could argue that having taken the action to set `KAYOBE_AIO_LVM` to `false`, it would be reasonable to modify the config to unset controller_lvm_groups as you did before. Otherwise you'd need to use a custom branch.

markgoddard (approved):

PR #979, opened 2024-03-07
Move release note to correct place

technowhizz:
Moving the release note to the right folder


jovial (approved):
Thanks, not sure how I managed to do that :laughing:

PR #981, opened 2024-03-08
Use StackHPC downstream requirements fork

assumptionsandg:


etc/kayobe/kolla/kolla-build.conf
```diff
@@ -9,3 +9,8 @@ base_tag = jammy-20231004
 base_tag = 9.{{ stackhpc_pulp_repo_rocky_9_minor_version }}
 {% endif %}
 build_args = {{ kolla_build_args.items() | map('join', ':') | join(',') }}
+
+[openstack-base]
+type = url
```
markgoddard: Unless you provide it with the URL of a tarbacll, the type will need to be git.

etc/kayobe/kolla/kolla-build.conf
```diff
@@ -9,3 +9,8 @@ base_tag = jammy-20231004
...
+location = https://github.com/stackhpc/requirements
+reference = {{ stackhpc_kolla_source_version }}
```
markgoddard: We may decide to start using tags for the kolla version, at which point this would be wrong. It would be better to either make new variables for the requirements URL and version, or just hard code to `stackhpc/{{ openstack_release }}`

markgoddard (changes requested):

mnasiadka:
Maybe my question is utterly stupid, but why not just use kolla template overrides to override requirements?https://github.com/openstack/kolla/commit/19e40f160852a3519f3e58043b9b5555ca87f87d

markgoddard:
> Maybe my question is utterly stupid, but why not just use kolla template overrides to override requirements?[openstack/kolla@19e40f1](https://github.com/openstack/kolla/commit/19e40f160852a3519f3e58043b9b5555ca87f87d)

It looks like that isn't universally applied, so would require changes to kolla source. I suppose it would allow changes to be specific to each image, which may have pros and cons.

markgoddard (approved):

PR #983, opened 2024-03-11
Fix Tempest doc long line

Alex-Welsh:


tox.ini
```diff
@@ -6,7 +6,7 @@ skipsdist = True
 [testenv]
 install_command = pip install {opts} {packages}
 deps =
-    -c{env:TOX_CONSTRAINTS_FILE:https://releases.openstack.org/constraints/upper/yoga}
```
markgoddard: This should work if you've cleared your pip cache

Alex-Welsh: Reverted

markgoddard (approved):

PR #984, opened 2024-03-11
Yoga bump magnum tag

scrungus:


JohnGarbutt (dismissed):
This looks correct to me, and required for ECMWF

releasenotes/notes/bump-magnum-capi-helm-6723d89456e6a590.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
```
Alex-Welsh: nit: I think upgrade tags are for notes that users should be aware of when upgrading, like a variable name changing or something non-backwards-compatible. This should probably either be treated like a new feature or a fix to an old one

mnasiadka: Well - here's a description of categories: https://docs.openstack.org/reno/latest/user/usage.html#editing-a-release-note
I think upgrade is a bit better than feature (more people read upgrade section than others) - it might fit other, but then nobody cares

markgoddard: I agree with Alex here. Upgrade is for cases that may require user action or be backwards incompatible.

scrungus: done

markgoddard (approved):

scrungus:
https://github.com/stackhpc/stackhpc-release-train/actions/runs/8359424251/job/22882730344

PR #987, opened 2024-03-12
Bump magnum-capi-helm version

sd109:


releasenotes/notes/bump-magnum-capi-helm-7e4ad37d3d9eecce.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
```
Alex-Welsh, line 2: Upgrade is for notes that the user should be aware of when upgrading i.e. breaking changes. This is more of a feature (or fix if the update is to fix an issue)

markgoddard (approved):

PR #988, opened 2024-03-13
Add missing grafana plugins

technowhizz:
Adds missing grafana plugins

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -320,7 +320,9 @@ kolla_build_blocks:
     ADD additions-archive /
   grafana_plugins_install: |
     RUN grafana-cli plugins install vonage-status-panel \
-        && grafana-cli plugins install grafana-piechart-panel
+        && grafana-cli plugins install grafana-piechart-panel \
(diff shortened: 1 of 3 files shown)


etc/kayobe/kolla.yml
```diff
@@ -320,7 +320,9 @@ kolla_build_blocks:
     ADD additions-archive /
   grafana_plugins_install: |
     RUN grafana-cli plugins install vonage-status-panel \
-        && grafana-cli plugins install grafana-piechart-panel
+        && grafana-cli plugins install grafana-piechart-panel \
```
mnasiadka, line 323: Out of curiosity - why not do that upstream?

technowhizz: Why not add the plugins upstream?

These are the upstream plugins I am adding. If you mean why not add the skc plugins upstream then I think they're only unique to us and the ceph dashboard we use. 

Ultimately though it would be nice to be able to merge the plugin lists instead of having skc replace the upstream one. We may have a solution to that, that we will propose upstream. But in the meantime I think we need this :)

jovial (approved):
Seems sane to me

PR #989, opened 2024-03-14
Support allowed vulnerabilities in Trivy

assumptionsandg:
Support allowed vulnerabilities in Trivy per image and globally

### DIFF
--- a/etc/kayobe/trivy/allowed-vulnerabilities.yml
+++ b/etc/kayobe/trivy/allowed-vulnerabilities.yml
@@ -7,7 +7,7 @@
 # keystone_allowed_vulnerabilities:
 #   - CVE-2022-2447
 #
-# barbican-api_allowed_vulnerabilities:
+# barbican_api_allowed_vulnerabilities:
 #   - CVE-2023-31047
 
 global_allowed_vulnerabilities:
--- a/etc/kayobe/trivy/allowed-vulnerabilities.yml
+++ b/etc/kayobe/trivy/allowed-vulnerabilities.yml
@@ -1,0 +1,18 @@
+---
+###############################################################################
+# Trivy allowed vulnerabilities list
+
+# Example allowed vulnerabilities file setup
+#
+# keystone_allowed_vulnerabilities:
+#   - CVE-2022-2447
+#
+# barbican-api_allowed_vulnerabilities:
+#   - CVE-2023-31047
+
+global_allowed_vulnerabilities:
+  - CVE-2024-36039
+
+###############################################################################
+# Dummy variable to allow Ansible to accept this file.
+workaround_ansible_issue_8743: yes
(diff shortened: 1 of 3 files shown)


etc/kayobe/trivy.yml
```diff
@@ -0,0 +1,15 @@
...
+###############################################################################
+# Trivy allowed vulnerabilities list
+
+# Example allowed vulnerabilities file setup
+#
+# keystone_allowed_vulnerabilities:
+#   CVE-2022-2447
+#
+horizon_allowed_vulnerabilities:
+  CVE-2023-31047
```
markgoddard: Do we actually want to allow this one or was it just used for testing?

tools/scan-images.sh
```diff
@@ -34,6 +34,11 @@ touch image-scan-output/clean-images.txt image-scan-output/dirty-images.txt
 # generate a csv summary
 for image in $images; do
   filename=$(basename $image | sed 's/:/\./g')
+  imagename=$(echo $filename | cut -d "." -f 1)
+  rm .trivyignore && touch .trivyignore
+  for vulnerability in $(yq -e .$imagename'_allowed_vulnerabilities' $KAYOBE_CONFIG_PATH/trivy.yml); do
```
markgoddard: Does this do the right thing if an image does not have a list defined?

Does it work if the image name has dashes in? e.g. nova-compute

etc/kayobe/trivy.yml
markgoddard: Should this be a list? (add `- `)

etc/kayobe/trivy.yml
markgoddard, line 1: Unsure about the location of this file. I'd be tempted to move it somewhere else to avoid confusing operators, since it's not used by Kayobe.

Could be in the repo root, or etc/kayobe/trivy/allowed-vulnerabilities.yml

markgoddard (changes requested):

tools/scan-images.sh
holliefae: Doesn't seem to error on images not being defined.

Currently using underscores for consistency in the whitelists file, but can revert if necessary.

Alex-Welsh (dismissed):

etc/kayobe/trivy/allowed-vulnerabilities.yml
```diff
@@ -0,0 +1,18 @@
...
+# Example allowed vulnerabilities file setup
+#
+# keystone_allowed_vulnerabilities:
+#   - CVE-2022-2447
+#
+# barbican_api_allowed_vulnerabilities:
+#   - CVE-2023-31047
+
+global_allowed_vulnerabilities:
+  - CVE-2024-36039
```
markgoddard: Do we need this one now that the requirements change has merged?

holliefae: Moved to the example file setup, so people know `global_allowed_vulnerabilities` variable can be used in the future.

markgoddard (approved):

PR #1010, opened 2024-03-21
Ironic deployment guide documentation

assumptionsandg:
Ironic deployment documentation.

### DIFF
--- a/doc/source/configuration/ironic.rst
+++ b/doc/source/configuration/ironic.rst
@@ -1,0 +1,232 @@
+======
+Ironic
+======
+
+Ironic networking
+=================
+
+Ironic will require the workload provisioning and cleaning networks to be
+configured in ``networks.yml``
+
+The workload provisioning network will require an allocation pool for
+Ironic inspection and for Neutron, an example configuration is shown
+below.
+
+.. code-block:: yaml
+
+  # Workload provisioning network IP information.
+  provision_wl_net_cidr: "172.0.0.0/16"
+  provision_wl_net_allocation_pool_start: "172.0.0.4"
+  provision_wl_net_allocation_pool_end: "172.0.0.6"
+  provision_wl_net_inspection_allocation_pool_start: "172.0.1.4"
+  provision_wl_net_inspection_allocation_pool_end: "172.0.1.250"
+  provision_wl_net_neutron_allocation_pool_start: "172.0.2.4"
+  provision_wl_net_neutron_allocation_pool_end: "172.0.2.250"
+  provision_wl_net_neutron_gateway: "172.0.1.1"
+
+The cleaning network will also require a Neutron allocation pool.
+
+.. code-block:: yaml
+
+  # Cleaning network IP information.
+  cleaning_net_cidr: "172.1.0.0/16"
+  cleaning_net_allocation_pool_start: "172.1.0.4"
+  cleaning_net_allocation_pool_end: "172.1.0.6"
+  cleaning_net_neutron_allocation_pool_start: "172.1.2.4"
+  cleaning_net_neutron_allocation_pool_end: "172.1.2.250"
+  cleaning_net_neutron_gateway: "172.1.0.1"
+
+OpenStack Config
+================
+
+Overcloud Ironic will require a router to exist between the internal API
+network and the provision workload network, a way to achieve this is by
+using `OpenStack Config <https://github.com/stackhpc/openstack-config>`
+to define the internal API network in Neutron and set up a router with
+a gateway.
+
+It not necessary to define the provision and cleaning networks in this
+configuration as they will be generated during
+
+.. code-block:: console
+
+  kayobe overcloud post configure
+
+The openstack config file could resemble the network, subnet and router
+configuration shown below:
+
+.. code-block:: yaml
+
+  networks:
+    - "{{ openstack_network_intenral }}"
+  openstack_network_internal:
+    name: "internal-net"
+    project: "admin"
+    provider_network_type: "vlan"
+    provider_physical_network: "physnet1"
+    provider_segmentation_id: 458
+    shared: false
+    external: true
+
+  subnets:
+    - "{{ openstack_subnet_internal }}"
+  openstack_subnet_internal:
+    name: "internal-net"
+    project: "admin"
+    cidr: "10.10.3.0/24"
+    enable_dhcp: true
+    allocation_pool_start: "10.10.3.3"
+    allocation_pool_end: "10.10.3.3"
+
+  openstack_routers:
+    - "{{ openstack_router_ironic }}"
+
+  openstack_router_ironic:
+    - name: ironic
+      project: admin
+      interfaces:
+        - net: "provision-net"
+          subnet: "provision-net"
+          portip: "172.0.1.1"
+        - net: "cleaning-net"
+          subnet: "cleaning-net"
+          portip: "172.1.0.1"
+          network: internal-net
+
+To provision baremetal nodes in Nova you will also require setting a flavour
+speciifc to that type of baremetal host. You will need to replace the custom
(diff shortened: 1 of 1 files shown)


doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,230 @@
...
+
+.. code-block:: json
+
+  "conntrack_helpers": {
+    {"protocol": "udp",
+     "port": 69, 
+     "helper": "tftp"}
+  }
+
+Currently it's not possible to add this helper via the OpenStack CLI, to add
```
holliefae, line 156: Possibly Train related?

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,230 @@
...
+Node enrollment
+===============
+
+When nodes are defined in the inventory you can begin enrolling them by
+invoking the Kayobe commmand (Note that only the Redfish driver is supported
+by this command)
+
+.. code-block:: console
+
+  kayobe baremetal compute register
```
holliefae, line 224: Not available yet pending this upstream [patch](https://review.opendev.org/c/openstack/kayobe/+/909671)

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,232 @@
...
+Ironic
+======
+
+Ironic networking
+=================
+
+Ironic will require the workload provisioning and cleaning networks to be
+configured in ``networks.yml``
+
+The workload provisioning network will require an allocation pool for
```
dougszumski, line 11: It would helpful to clarify what these pools are used for. I.e. note that one pool is for statically assigned IPs that will be used by OpenStack Ironic services, and the other pools (Neutron ones) are for dynamically assigning IPs to hosts being provisioned / inspected. I like the idea of suggesting a /16 to start with, to leave room for growth.

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,232 @@
...
+        - net: "provision-net"
+          subnet: "provision-net"
+          portip: "172.0.1.1"
+        - net: "cleaning-net"
+          subnet: "cleaning-net"
+          portip: "172.1.0.1"
+          network: internal-net
+
+To provision baremetal nodes in Nova you will also require setting a flavour
+speciifc to that type of baremetal host. You will need to replace the custom
```
dougszumski, line 97: s/speciifc/specific

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,232 @@
...
+  cleaning_net_allocation_pool_start: "172.1.0.4"
+  cleaning_net_allocation_pool_end: "172.1.0.6"
+  cleaning_net_neutron_allocation_pool_start: "172.1.2.4"
+  cleaning_net_neutron_allocation_pool_end: "172.1.2.250"
+  cleaning_net_neutron_gateway: "172.1.0.1"
+
+OpenStack Config
+================
+
+Overcloud Ironic will require a router to exist between the internal API
```
dougszumski, line 42: I haven't seen a site use a neutron router for this yet. It's an interesting idea. Is there a specific reason for using one? Normally, at least in the past, we have avoided a dedicated router: https://docs.google.com/document/d/1H3hGYzJzieX8w7phxS3xmD6i6OEZgiHmBjvZ7EJEbU4/edit#heading=h.gyl3n9h885m

It would be helpful to start with why a router is required. Specifically, because the TFTP/HTTP server hosting the deploy images is bound exclusively to internal API network, rather than all possible networks the node might PXE boot on. When baremetal nodes PXE boot, they are given the location of the TFTP/HTTP server to fetch the images from. Since the cleaning/provisioning/inspection network is generally not the same as the internal API network, routing is required.

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,232 @@
...
+
+  kayobe overcloud post configure
+
+The openstack config file could resemble the network, subnet and router
+configuration shown below:
+
+.. code-block:: yaml
+
+  networks:
+    - "{{ openstack_network_intenral }}"
```
dougszumski, line 61: s/intenral/internal

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,232 @@
...
+  openstack_network_internal:
+    name: "internal-net"
+    project: "admin"
+    provider_network_type: "vlan"
+    provider_physical_network: "physnet1"
+    provider_segmentation_id: 458
```
dougszumski, line 67: nit: reference VLAN variable

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,232 @@
...
+  cleaning_net_neutron_allocation_pool_end: "172.1.2.250"
+  cleaning_net_neutron_gateway: "172.1.0.1"
+
+OpenStack Config
+================
+
+Overcloud Ironic will require a router to exist between the internal API
+network and the provision workload network, a way to achieve this is by
+using `OpenStack Config <https://github.com/stackhpc/openstack-config>`
+to define the internal API network in Neutron and set up a router with
```
dougszumski, line 45: Have you tested this configuration? I imagine you still need policy based routing here to prevent traffic bypassing the router? Eg. Node on cleaning network tries HTTP fetch of IPA image, request is routed to a controller, controller replies not via the router, but directly to the node via the cleaning interface. The node sees that the traffic came back from a different MAC, and ignores it.

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,232 @@
...
+    "resources:CUSTOM_<YOUR_BAREMETAL_RESOURCE_CLASS>": 1
+    "resources:VCPU": 0
+    "resources:MEMORY_MB": 0
+    "resources:DISK_GB": 0
+
+Enabling conntrack
+==================
+
+UEFI booting requires conntrack_helper to be configured on the Ironic neutron
+router, this is due to TFTP traffic being dropped due to being UDP. You will
```
dougszumski, line 122: This is another complexity that would go away if we didn't use the Neutron router. Assuming conntrack is already enabled on the controller.

dougszumski (commented):
Good start @assumptionsandg. I know you're still working on this, but I've added some comments.

doc/source/configuration/ironic.rst
```diff
@@ -0,0 +1,232 @@
+======
```
markgoddard, line 1: Thanks for sharing the information you have @assumptionsandg. I think we need to consider what the scope of this doc is. We have some general notes on ironic deployment in https://docs.google.com/document/d/1H3hGYzJzieX8w7phxS3xmD6i6OEZgiHmBjvZ7EJEbU4/edit that are possibly more detailed than this but not public. We might also consider whether a generic description of ironic config might be better placed in kayobe or kolla-ansible upstream docs.

Docs that might belong in SKC are generally either
1. our opinionated way of doing things, that wouldn't belong upstream
2. dependent on additional config in SKC

What was the original request for these docs about? Was it to cover the use of overcloud ironic to manage hypervisors? That could be considered under 1.

doc/source/configuration/ironic.rst
markgoddard, line 1: You'd need to reference this doc from configuration/index.rst in order for it to be included in the docs.

doc/source/configuration/ironic.rst
holliefae: I think the purpose was originally managing hypervisors but I wasn't sure whether to include all of the RAL documentation I had up to this point, I do still need to add the section about managing hypervisors after deployment.

Maybe it would make sense to split the deployment part out and propose that upstream? The openstack-config part might be too opinionated for upstream though.

assumptionsandg:
Incorporated into #1089

PR #1011, opened 2024-03-22
Add StackHPC Ironic tunings

jovial:


etc/kayobe/kolla/config/ironic/ironic-conductor.conf
```diff
@@ -0,0 +1,92 @@
...
+[agent]
+# Use iPXE server for direct deploy source
+image_download_source = http
+# Reduce load by only collecting logs on failure
+deploy_logs_collect = on_failure
+
+[console]
+socat_address = {% raw %}{{ api_interface_address }}{% endraw %}
+
+[conductor]
```
dougszumski, line 7: Duplicate section

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
```diff
@@ -0,0 +1,92 @@
...
+automated_clean=true
```
dougszumski, line 8: I saw at some sites we have this cleaning tweak from the Ironic docs:

```
[deploy]/erase_devices_priority=0
[deploy]/erase_devices_metadata_priority=0
[conductor]/clean_step_priority_override=deploy.erase_devices_express:5
```

I know cleaning is very site specific. Probably safer to leave the above out and default to scrub. Perhaps worth linking to the Ironic Doc?

dougszumski (commented):
Nice!

etc/kayobe/kolla/config/global.conf
```diff
@@ -0,0 +1,4 @@
+[oslo_messaging_rabbit]
+rpc_conn_pool_size = 300
+rpc_thread_pool_size = 2048
+rpc_response_timeout = 3600
```
JohnGarbutt: I am not sure we want this being global?

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
```diff
@@ -0,0 +1,92 @@
...
+rpc_response_timeout = 360
+
+[agent]
+# Use iPXE server for direct deploy source
+image_download_source = http
+# Reduce load by only collecting logs on failure
+deploy_logs_collect = on_failure
+
+[console]
+socat_address = {% raw %}{{ api_interface_address }}{% endraw %}
```
JohnGarbutt: Yeah, this is duplicate to. Do we need it?

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
```diff
@@ -0,0 +1,92 @@
...
+
+[conductor]
+automated_clean=true
+# We have busy conductors failing to heartbeat
+# Default is 10 secs
+heartbeat_interval = 30
+# Default is 60 seconds
+heartbeat_timeout = 360
+sync_local_state_interval = 360
```
JohnGarbutt: I think there is a pthread config we can use here as well now?

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
JohnGarbutt: I would actually like our default to be fail if secure erase fails, with a note on how to work around that.

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
```diff
@@ -0,0 +1,92 @@
...
+
+# Normally this is 100. We see eventlet threads
+# not making much progress, to for saftey reduce
+# this by half, should leave work on rabbit queu
+workers_pool_size = 50
+# Normally this is 8, keep it same
+period_max_workers = 8
+
+# Stop power state sync
```
JohnGarbutt: I actually think this is a bad move now, disabling placement and power stuff.
We should consider just increasing the interval a bit.

Probably 120 like the ones below?

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
```diff
@@ -0,0 +1,92 @@
...
+power_failure_recovery_interval = 0
+# Stop checking for orphan allocations for now
+check_allocations_interval = 0
+
+# Wait much longer before provision timeout check, to reduce background load
+# The default is 60 seconds
+check_provision_state_interval = 120
+check_rescue_state_interval = 120
+
+[database]
```
JohnGarbutt, line 34: I think we have some of this in upstream kolla now, from the work Pierre did with failovers?

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
```diff
@@ -0,0 +1,92 @@
...
+# to get a connection out the pool that has already timed out
+connection_recycle_time = 1800
+# By default this is 30 seconds, but as we reduce
+# the pool overflow, some people will need to wait longer
+pool_timeout = 60
+
+[deploy]
+shred_random_overwrite_iterations = 0
+shred_final_overwrite_with_zeros = false
+continue_if_disk_secure_erase_fails = true
```
JohnGarbutt: Ah, @dougszumski this is part of what you were talking about?

etc/kayobe/kolla/config/nova/nova-compute-ironic.conf
```diff
@@ -2,3 +2,25 @@
...
 host = {{ kolla_nova_compute_ironic_static_host_name | mandatory('You must set a static host name to help with service failover. See the operations documentation, Ironic section.') }}
 {% endif %}
+# Don't limit the number of concurrent builds for the nova ironic compute
+# service.
+max_concurrent_builds = 35
+
+force_config_drive = True
+
+# Disable power state sync interval (default is 600 seconds)
+sync_power_state_interval = -1
```
JohnGarbutt: I am not sure we want this still.

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
jovial: done

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
jovial: This one:

```
# Run the health check heartbeat thread through a native
# python thread by default. If this option is equal to False
# then the health check heartbeat will inherit the execution
# model from the parent process. For example if the parent
# process has monkey patched the stdlib by using
# eventlet/greenlet then the heartbeat will be run through a
# green thread. This option should be set to True only for the
# wsgi services. (boolean value)
#heartbeat_in_pthread = false
```
I guess that should only be set on ironic-api. Was there some other tuning options too?

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
jovial: removed

etc/kayobe/kolla/config/global.conf
jovial: removed

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
jovial: removed

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
jovial: I have removed the options that are in upstream kolla (although the values are different to the ones we had in the first version of this PR)

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
jovial: I explicitly set a value for erase_devices as the priroity is dependent on the hardware manager in use. What do you think?

etc/kayobe/kolla/config/nova/nova-compute-ironic.conf
jovial: Remvoed

jovial:
I've removed the bits that were flagged and have made pieces conditional on having Ironic enabled. Please review again :pray:

etc/kayobe/kolla/config/nova/nova-compute-ironic.conf
```diff
@@ -2,3 +2,16 @@
 [DEFAULT]
```
navaati: Hey, that should probably be outside of the `{% if %}` now that there is other stuff under it.

mnasiadka:
any chance we move forward on this?

etc/kayobe/kolla/config/nova/nova-compute-ironic.conf
jovial: done

Alex-Welsh:
@JohnGarbutt or @dougszumski Any chance you could re-review this? It looks like exactly the sort of opinionated defaults we should be setting in SKC, but I don't have the subject knowledge to review it myself

bbezak:
bump on that PR as it is very useful @jovial @dougszumski @JohnGarbutt - or we can build on top of https://github.com/stackhpc/stackhpc-kayobe-config/pull/1729 @jackhodgkiss

etc/kayobe/kolla/config/ironic/ironic-conductor.conf
JohnGarbutt: Coming back to this, I think we should move to erase_devices_express, now I understand it better. I think we need to make the default good out the box with our pre-built IPA.

jovial: Express will fallback to metadata erase if hardware assisted erase fails i.e it could leave data on the disk. Wouldn't it be better to default to a secure form of erasing i.e erase_devices?  We'd then allow users to opt into a faster potentially non-secure metadata erase if they accepted that security trade off.

jovial:
Closing in favor of https://github.com/stackhpc/stackhpc-kayobe-config/pull/2229/

PR #1016, opened 2024-04-03
Firewall config

Alex-Welsh:
Testing changes in CI because my AIO died.


etc/kayobe/compute.yml
```diff
@@ -126,22 +126,57 @@
...
 
 # Whether to install and enable firewalld.
-#compute_firewalld_enabled:
+compute_firewalld_enabled: true
 
 # A list of zones to create. Each item is a dict containing a 'zone' item.
-#compute_firewalld_zones:
+compute_firewalld_zones: "{{ compute_firewalld_zones_default | union(compute_firewalld_zones_extra) | unique | select }}"
+compute_firewalld_zones_extra: []
+compute_firewalld_zones_default: "{% set network_zones = [] %}{% for network in network_interfaces %}{% set _ = network_zones.append({'zone': network | net_zone}) %}{% endfor %}{{ network_zones }}"
```
markgoddard: I think you need to apply the filtering in here because `select` will treat `{'"zone": None}` as truthy.

markgoddard: IMO it would be more readable to use a multiple lines, e.g.

```
compute_firewalld_zones_default: >-
  {% set network_zones = [] %}
  {% for network in network_interfaces %}
  {% set _ = network_zones.append({'zone': network | net_zone}) %}
  {% endfor %}
  {{ network_zones }}
```

etc/kayobe/compute.yml
```diff
@@ -126,22 +126,57 @@
 # Compute node firewalld configuration.
 
 # Whether to install and enable firewalld.
-#compute_firewalld_enabled:
+compute_firewalld_enabled: true
 
 # A list of zones to create. Each item is a dict containing a 'zone' item.
-#compute_firewalld_zones:
+compute_firewalld_zones: "{{ compute_firewalld_zones_default | union(compute_firewalld_zones_extra) | unique | select }}"
+compute_firewalld_zones_extra: []
```
markgoddard: Move extra vars below defaults with an empty line to help with merge conflicts?

etc/kayobe/compute.yml
```diff
@@ -126,22 +126,57 @@
...
+        zone: "{{ public_net_name | net_zone}}"
+        network: "{{ public_net_name }}"
+    enabled: true
+  # GENEVE
+  - rules:
+      - port: 6081/udp
+        state: enabled
+        zone: "{{ tunnel_net_name | net_zone}}"
+        network: "{{ tunnel_net_name }}"
+    enabled: "{{ ('geneve' in (kolla_neutron_ml2_type_drivers + kolla_neutron_ml2_tenant_network_types)) | bool }}"
```
markgoddard: nit: bool unnecessary here, result is boolean

etc/kayobe/compute.yml
```diff
@@ -126,22 +126,57 @@
...
+
+compute_firewalld_rules: "{{ compute_firewalld_rules_default | selectattr('enabled', 'true') | map(attribute='rules') | flatten | selectattr('network', 'in', network_interfaces) | selectattr('zone') | union(compute_firewalld_rules_extra) | unique | select }}"
+compute_firewalld_rules_extra: []
+compute_firewalld_rules_default:
+  # Common
+  - rules:
+      - service: ssh
+        state: enabled
+        zone: "{{ provision_oc_net_name | net_zone}}"
+        network: "{{ provision_oc_net_name }}"
```
markgoddard: admin_oc_net_name

etc/kayobe/compute.yml
```diff
@@ -126,22 +126,57 @@
...
+      - service: dhcpv6-client
+        state: disabled
+        zone: "{{ public_net_name | net_zone}}"
```
markgoddard: nit: whitespace

etc/kayobe/compute.yml
```diff
@@ -126,22 +126,57 @@
...
+        network: "{{ public_net_name }}"
+      - service: ssh
```
markgoddard: In an aio environment this would conflict with the admin_oc_net_name ssh rule

etc/kayobe/controllers.yml
```diff
@@ -134,23 +134,106 @@
...
+        network: "{{ provision_wl_net_name }}"
+        state: enabled
+      - service: tftp
+        zone: "{{ provision_wl_net_name | net_zone }}"
+        network: "{{ provision_wl_net_name }}"
+        state: enabled
+      - port: 8089/tcp
+        zone: "{{ provision_wl_net_name | net_zone }}"
+        network: "{{ provision_wl_net_name }}"
+        state: enabled
```
markgoddard: Only necessary when overcloud ironic is in use

etc/kayobe/controllers.yml
```diff
@@ -134,23 +134,106 @@
...
+        network: "{{ public_net_name }}"
+        state: enabled
+    enabled: "{{ kolla_enable_designate | bool }}"
+  # GENEVE
+  - rules:
+      - port: 6081/udp
+        zone: "{{ tunnel_net_name | net_zone }}"
+        network: "{{ tunnel_net_name }}"
+        state: enabled
+    enabled: "{{ ('geneve' in (kolla_neutron_ml2_type_drivers + kolla_neutron_ml2_tenant_network_types)) | bool }}"
```
markgoddard: Strictly only necessary on network hosts - it's possible to make these separate. Could check the host is in the 'network' group?

etc/kayobe/seed.yml
```diff
@@ -155,23 +155,101 @@ seed_extra_containers: {}
...
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      - service: ntp
+        zone: "{{ provision_oc_net_name | net_zone }}"
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      - service: ntp
+        zone: "{{ switch_mgmt_net_name | net_zone }}"
+        network: "{{ switch_mgmt_net_name }}"
+        state: enabled
```
markgoddard: I think this one might be environment specific

etc/kayobe/kolla.yml
```diff
@@ -365,7 +365,7 @@ kolla_build_customizations: "{{ kolla_build_customizations_common | combine(koll
 
 # Dict mapping Kolla Dockerfile ARG names to their values.
 kolla_build_args:
-  node_exporter_version: "1.5.0" # kolla has 1.4.0
+  node_exporter_version: "1.5.0"  # kolla has 1.4.0
```
markgoddard: ?

markgoddard (commented):
I've done a quick first pass. I think we need to consider whether these things are set by default or need to be opted into (like the LVM config). In the latter case we might provide some variables that can be referenced to enable the firewall config.

Alex-Welsh:
> I've done a quick first pass. I think we need to consider whether these things are set by default or need to be opted into (like the LVM config). In the latter case we might provide some variables that can be referenced to enable the firewall config.

My opinion would be that we could disable it for Antelope and enable it for Caracal. I'll look into making a single bool to enable/disable all the rules at once.

etc/kayobe/compute.yml
Alex-Welsh: Done

etc/kayobe/compute.yml
Alex-Welsh: Done

etc/kayobe/compute.yml
Alex-Welsh: Done

etc/kayobe/compute.yml
Alex-Welsh: Done

etc/kayobe/compute.yml
Alex-Welsh: Done

etc/kayobe/compute.yml
Alex-Welsh: Removed

etc/kayobe/kolla.yml
Alex-Welsh: Removed

etc/kayobe/controllers.yml
Alex-Welsh: Done

etc/kayobe/seed.yml
Alex-Welsh: Removed

etc/kayobe/kolla.yml
Alex-Welsh: Unremoved, I remember now: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/8649973848/job/23717449450?pr=1016#step:5:22

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,37 @@
...
+This will configure the standard set of firewalld rules on controller hosts.
+Rule definitions are automatically added according to group membership.Rule
+sets exist for the following groups:
+
+* Controllers - ``stackhpc_controller_firewalld_rules``
+* Compute - ``stackhpc_compute_firewalld_rules``
+* Storage - ``stackhpc_storage_firewalld_rules``
+* Monitoring - ``stackhpc_monitoring_firewalld_rules``
+* Wazuh Manager Infrastructure VM - ``stackhpc_infra_vm_firewalld_rules``
+* Ansible Control host Infrastructure VM - ``stackhpc_infra_vm_firewalld_rules``
```
markgoddard: Wrong variable names

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,369 @@
...
+stackhpc_firewalld_zones: |
+  {% set network_zones = [] %}
+  {% for network in network_interfaces %}
+  {% if network | net_zone is not none %}
+  {% set _ = network_zones.append({'zone': network | net_zone }) %}
+  {% endif %}
+  {% endfor %}
+  {{ network_zones }}
+
+stackhpc_firewalld_rules: |
```
markgoddard, line 27: Interesting approach

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,369 @@
+---
+#########################################
+# Stackhpc default firewall configuration
+#########################################
+
+stackhpc_firewalld_zones: |
+  {% set network_zones = [] %}
+  {% for network in network_interfaces %}
+  {% if network | net_zone is not none %}
```
markgoddard: I think this would work and would catch an empty string:
```
{% if network | net_zone %}
```

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,369 @@
...
+
+stackhpc_firewalld_rules: |
+  {{  (stackhpc_controller_firewalld_rules if 'controllers' in group_names else []) +
+      (stackhpc_compute_firewalld_rules if 'compute' in group_names else []) +
+      (stackhpc_storage_firewalld_rules if 'storage' in group_names else []) +
+      (stackhpc_monitoring_firewalld_rules if 'monitoring' in group_names else []) +
+      (stackhpc_seed_firewalld_rules if 'seed' in group_names else []) +
+      (stackhpc_seed_hypervisor_firewalld_rules if 'seed_hypervisor' in group_names else []) +
+      (stackhpc_wazuh_manager_infra_vm_firewalld_rules if 'wazuh_manager' in group_names else []) +
+      (stackhpc_ansible_control_infra_vm_firewalld_rules if 'ansible_control' in group_names else []) }}
```
markgoddard:
```suggestion
      (stackhpc_seed_hypervisor_firewalld_rules if 'seed-hypervisor' in group_names else []) +
      (stackhpc_wazuh_manager_infra_vm_firewalld_rules if 'wazuh-manager' in group_names else []) +
      (stackhpc_ansible_control_infra_vm_firewalld_rules if 'ansible-control' in group_names else []) }}
```

markgoddard: Alternatively could check `inventory_hostname == localhost`

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,369 @@
...
+
+###############################################################################
+# Controller firewalld rules
+
+stackhpc_controller_firewalld_rules: |
```
markgoddard: I think you need to remove newlines:
```suggestion
stackhpc_controller_firewalld_rules: >-
```

etc/kayobe/networks.yml
```diff
@@ -59,6 +59,12 @@
 # hosts
 #cleaning_net_name:
 
+# Name of the network used to manage octavia loadbalancers
+octavia_net_name: ""
+
+# Name of the network used to manage network switches
+switch_mgmt_net_name: ""
```
markgoddard: Can remove this and rules referring to it

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,37 @@
...
+``etc/kayobe/inventory/group_vars/all/firewall``. This allows configuration
+variables to be overridden on a per-group or per-host basis (which would not be
+possible for an "extra variable" in ``etc/kayobe/*.yml``). This configuration
+is not used by default, and must be actively opted into. This can be done as
+follows:
+
+.. code-block:: yaml
+   :caption: ``etc/kayobe/controllers.yml``
+
+   controller_firewalld_zones: "{{ stackhpc_firewalld_zones }}"
```
markgoddard, line 37: Better add the enable flag and default zone here too

markgoddard (changes requested):
reno pls :)

etc/kayobe/networks.yml
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Both appear to work. I've gone with folded because it makes more logical sense to me

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Neato

doc/source/configuration/firewall.rst
Alex-Welsh: Done

doc/source/configuration/firewall.rst
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Given that all the zones could just be set to one universal variable, I wanted to do the same for rules to keep it nice and clean for the user

etc/kayobe/environments/ci-multinode/kolla/inventory/group_vars/all
```diff
@@ -0,0 +1,6 @@
+---
+
+# Just adding for testing
+enable_firewalld: true
+firewalld_zones: "{{ stackhpc_firewalld_zones }}"
+firewalld_rules: "{{ stackhpc_firewalld_rules }}"
```
jackhodgkiss: Will this be removed before merging. Or at least the comment reworded.

Alex-Welsh: For sure, it will be removed. I only added it cause I was hoping that it would apply the rules everywhere automatically when I deployed a multinode from this branch. Sadly it didn't turn out like that anyway.

etc/kayobe/inventory/group_vars/all/firewall
markgoddard: Actually `|` is literal, and `>` is folded. `-` is the chomping mode...

etc/kayobe/kolla/globals.yml
```diff
@@ -53,3 +53,6 @@ prometheus_instance_label: "{% raw %}{{ ansible_facts.hostname }}{% endraw %}"
 # in Yoga. This is required to include a valid value for the flavor_id label on
 # openstack_nova_server_status metrics.
 prometheus_openstack_exporter_compute_api_version: "2.1"
+
+# Open up ports in firewalld for services on the public API network.
+enable_external_api_firewalld: true
```
markgoddard: This should be opt-in as well (and done so in ci-aio and ci-multinode)

etc/kayobe/environments/ci-aio/controllers.yml
```diff
@@ -10,3 +10,9 @@ controller_bootstrap_user: "{{ os_distribution if os_distribution == 'ubuntu' el
 # for the exact configuration.
 controller_lvm_groups:
   - "{{ stackhpc_lvm_group_rootvg }}"
+
+# Controller firewalld configuration. See inventory/group_vars/all/firewall for
+# the exact configuration.
+controller_firewalld_zones: "{{ stackhpc_firewalld_zones }}"
+controller_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+controller_firewalld_enabled: true
```
markgoddard: Default zone too?

etc/kayobe/environments/ci-aio/controllers.yml
```diff
@@ -10,3 +10,9 @@ controller_bootstrap_user: "{{ os_distribution if os_distribution == 'ubuntu' el
 # for the exact configuration.
```
markgoddard, line 10: How about some multinode config?

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
+---
+#########################################
+# Stackhpc default firewall configuration
```
markgoddard:
```suggestion
# StackHPC default firewall configuration
```

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+stackhpc_firewalld_zones: |
+  {% set network_zones = [] %}
+  {% for network in network_interfaces %}
+  {% if network | net_zone %}
+  {% set _ = network_zones.append({'zone': network | net_zone }) %}
+  {% endif %}
+  {% endfor %}
+  {{ network_zones }}
+
+# Rules are filtered to ensure that none are conflicting
```
markgoddard: So this just takes the first rule in the case of a conflict? Would it not make more sense to fail?

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
+---
+#########################################
+# Stackhpc default firewall configuration
+#########################################
+
+stackhpc_firewalld_zones: |
```
markgoddard, line 14: A short comment might be nice for the two that users might reference.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+      (stackhpc_seed_hypervisor_firewalld_rules if 'seed-hypervisor' in group_names else []) +
+      (stackhpc_wazuh_manager_infra_vm_firewalld_rules if 'wazuh-manager' in group_names else []) +
+      (stackhpc_ansible_control_infra_vm_firewalld_rules if 'ansible-control' in group_names else []) }}
+
+###############################################################################
+# Controller firewalld rules
+
+stackhpc_controller_firewalld_rules: >-
+  {{ stackhpc_controller_firewalld_rules_default |
+    selectattr('enabled', 'true') |
```
markgoddard: nit: indentation

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+        network: "{{ provision_wl_net_name }}"
+        state: enabled
+      - service: ntp
+        zone: "{{ provision_wl_net_name | net_zone }}"
+        network: "{{ provision_wl_net_name }}"
+        state: enabled
+      - service: tftp
+        zone: "{{ provision_wl_net_name | net_zone }}"
+        network: "{{ provision_wl_net_name }}"
+        state: enabled
```
markgoddard: These are only for overcloud ironic

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Ah, I must've reverted the change at some point. I did switch it but it appears to have switched back.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+  {% if network | net_zone %}
+  {% set _ = network_zones.append({'zone': network | net_zone }) %}
+  {% endif %}
+  {% endfor %}
+  {{ network_zones }}
+
+# Rules are filtered to ensure that none are conflicting
+stackhpc_firewalld_rules: |
+  {% set stackhpc_firewalld_rules_filtered = [] %}
+  {% for rule in stackhpc_firewalld_rules_unfiltered %}
```
markgoddard: I guess you could add the zone to the rule here based on the network name. Would need to check the zone is defined at this point too though.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+        network: "{{ octavia_net_name }}"
+        state: enabled
+    enabled: "{{ kolla_enable_octavia | bool }}"
+  # Overcloud Ironic
+  - rules:
+      - port: 8089/tcp
+        zone: "{{ provision_wl_net_name | net_zone }}"
+        network: "{{ provision_wl_net_name }}"
+        state: enabled
+    enabled: "{{ kolla_enable_octavia | bool }}"
```
markgoddard:
```suggestion
    enabled: "{{ kolla_enable_ironic | bool }}"
```

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+    select }}
+
+stackhpc_storage_firewalld_rules_default:
+  # Common
+  - rules:
+      - service: ssh
+        zone: "{{ admin_oc_net_name | net_zone }}"
+        network: "{{ admin_oc_net_name }}"
+        state: enabled
+      - port: 9283/tcp
```
markgoddard, line 221: This one deserves a comment. It's the Ceph prometheus exporter.

markgoddard: Maybe also make it conditional on being in the mgrs group and having prometheus enabled.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+      - service: dhcpv6-client
+        zone: "{{ public_net_name | net_zone }}"
+        network: "{{ public_net_name }}"
+        state: disabled
+    enabled: true
+  - rules:
+      - service: ssh
+        zone: "{{ public_net_name | net_zone }}"
+        network: "{{ public_net_name }}"
+        state: disabled
```
markgoddard: I suppose that strictly these rules are for when you use the default `public` zone, and could apply to any network. Maybe the correct condition is that `public` is one of the zones referenced in the config on this host, and for the SSH rule it should also not be the admin_oc_net_name

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+    enabled: "{{ public_net_name | net_zone != provision_oc_net_name | net_zone }}"
```
markgoddard:
```suggestion
    enabled: "{{ public_net_name | net_zone != admin_oc_net_name | net_zone }}"
```

etc/kayobe/inventory/group_vars/all/firewall
markgoddard: I'm wondering if it would help to have a list of common rules?

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,41 @@
...
+sets exist for the following groups:
+
+* Controllers - ``stackhpc_controller_firewalld_rules``
+* Compute - ``stackhpc_compute_firewalld_rules``
+* Storage - ``stackhpc_storage_firewalld_rules``
+* Monitoring - ``stackhpc_monitoring_firewalld_rules``
+* Wazuh Manager Infrastructure VM - ``stackhpc_wazuh_manager_infra_vm_firewalld_rules``
+* Ansible Control host Infrastructure VM - ``stackhpc_ansible_control_infra_vm_firewalld_rules``
+* Seed - ``stackhpc_seed_firewalld_rules``
+* Seed Hypervisor - ``stackhpc_seed_hypervisor_firewalld_rules``
```
markgoddard: I wonder if there's a config dump command we could provide to list the firewall rules that will be applied to each host.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+  - rules:
+      - service: ceph
+        zone: "{{ storage_net_name | net_zone }}"
+        network: "{{ storage_net_name }}"
+        state: enabled
+      - service: ceph-mon
+        zone: "{{ storage_net_name | net_zone }}"
+        network: "{{ storage_net_name }}"
+        state: "{{ 'enabled' if 'mons' in group_names else 'disabled' }}"
+    enabled: "{{ stackhpc_enable_ceph | default(false) | bool }}"
```
markgoddard: I'd just use the inventory here rather than enabling a new flag. There is a `ceph` group.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+        zone: "{{ admin_oc_net_name | net_zone }}"
+        network: "{{ admin_oc_net_name }}"
+        state: enabled
+      - port: 9283/tcp
+        zone: "{{ provision_oc_net_name | net_zone }}"
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+    enabled: true
+  - rules:
+      - service: ssh
```
markgoddard: Unsure why we would need this (I know it's in the RUG config)

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+        zone: "{{ storage_net_name | net_zone }}"
+        network: "{{ storage_net_name }}"
+        state: disabled
+    enabled: "{{ storage_net_name | net_zone != provision_oc_net_name | net_zone }}"
+  # Ceph
+  - rules:
+      - service: ceph
+        zone: "{{ storage_net_name | net_zone }}"
+        network: "{{ storage_net_name }}"
```
markgoddard, line 228: We probably also need this for the storage_mgmt_net_name

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+
+stackhpc_firewalld_rules_unfiltered: |
+  {{  (stackhpc_controller_firewalld_rules if 'controllers' in group_names else []) +
+      (stackhpc_compute_firewalld_rules if 'compute' in group_names else []) +
+      (stackhpc_storage_firewalld_rules if 'storage' in group_names else []) +
+      (stackhpc_monitoring_firewalld_rules if 'monitoring' in group_names else []) +
+      (stackhpc_seed_firewalld_rules if 'seed' in group_names else []) +
+      (stackhpc_seed_hypervisor_firewalld_rules if 'seed-hypervisor' in group_names else []) +
+      (stackhpc_wazuh_manager_infra_vm_firewalld_rules if 'wazuh-manager' in group_names else []) +
+      (stackhpc_ansible_control_infra_vm_firewalld_rules if 'ansible-control' in group_names else []) }}
```
markgoddard: Need to document the ansible-control group? Or just localhost?

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+    enabled: true
+  - rules:
+      - service: ssh
+        state: disabled
+        zone: "{{ public_net_name | net_zone }}"
+        network: "{{ public_net_name }}"
+    enabled: "{{ public_net_name | net_zone != provision_oc_net_name | net_zone }}"
+  # Pulp server
+  - rules:
+      - service: http
```
markgoddard:
```suggestion
      - port: {{ pulp_port }}/tcp
```

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+        zone: "{{ provision_oc_net_name | net_zone }}"
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      # nginx
```
markgoddard: This is bifrost's, not pulp's

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+    enabled: true  #FIXME: Make rules conditional on Bifrost deployment
+  # Redfish exporter
+  - rules:
+      - port: 9610/tcp
+        zone: "{{ provision_oc_net_name | net_zone }}"
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+    enabled: true
```
markgoddard: This one isn't in SKC yet, see https://github.com/stackhpc/stackhpc-kayobe-config/pull/501

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,437 @@
...
+      # Ironic inspector API
+      - port: 5050/tcp
+        zone: "{{ provision_oc_net_name | net_zone }}"
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      # Ironic API
+      - port: 6385/tcp
+        zone: "{{ provision_oc_net_name | net_zone }}"
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
```
markgoddard: These can go with dhcp & tftp

markgoddard (changes requested):

etc/kayobe/environments/ci-aio/controllers.yml
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: I wasn't sure how to make it fail. I'm testing a weird strategy of just doing something impossible in that case, to force it to break.

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: How about a long comment?

doc/source/configuration/firewall.rst
Alex-Welsh: Added

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Removed for now. Can come back to it if necessary

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Moved

etc/kayobe/kolla/globals.yml
Alex-Welsh: Done

Alex-Welsh: Forgot to add it to the docs. I'll include it in the next patch set

etc/kayobe/environments/ci-aio/controllers.yml
Alex-Welsh: Added

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Probably will be once this patch gets merged though?

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done, eventually

etc/kayobe/kolla/globals.yml
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Definitely agree with the common rules idea.
I'm not sure I 100% understand what you mean with your other comment, but I've updated the config with my interpretation

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Now using localhost

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: All ironic rules are now in the same ruleset

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,59 @@
...
+.. code-block:: yaml
+   :caption: ``etc/kayobe/controllers.yml``
+
+   controller_firewalld_enabled: true
+   controller_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+   controller_firewalld_zones: "{{ stackhpc_firewalld_zones }}"
+   # Predefined zones are listed here:
+   # https://firewalld.org/documentation/zone/predefined-zones.html
+   # Unset to leave the default zone unchanged
+   controller_firewalld_default_zone: drop
```
markgoddard: Are you planning to do enough testing to be confident in setting it to drop?

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,59 @@
...
+
+.. code-block:: yaml
+   :caption: ``etc/kayobe/kolla/globals.yml``
+
+   # Open up ports in firewalld for services on the public API network.
+   enable_external_api_firewalld: true
```
markgoddard: This should move to the end to keep the flow of kayobe config.

etc/kayobe/environments/ci-multinode/firewalld.yml
markgoddard, line 1: I know this looks quite neat, but IMO it would be more logical to put these in the relevant files, as a reference for how to enable the firewall.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,413 @@
...
+
+# A templated list of firewalld rules, according to the enabled services,
+# host's group membership, and host's network configuration.
+# This variable can be used to set any *_firewalld_rules kayobe variable e.g.
+# compute_firewalld_rules: "{{ stackhpc_firewalld_rules }}" in compute.yml
+stackhpc_firewalld_rules: |
+  {% set stackhpc_firewalld_rules_verified = [] %}
+  {% for rule in stackhpc_firewalld_rules_unverified | unique %}
+  {% if rule | ansible.utils.remove_keys('state') in stackhpc_firewalld_rules_verified | map('ansible.utils.remove_keys', 'state') %}
+  {% set stackhpc_firewalld_rules_verified = 'Invalid configuration! Two matching firewalld rules exist with different states' + 1 %}
```
markgoddard: Does this error get displayed to users? If so can we include the offending rule within it somehow?

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,413 @@
...
+        state: enabled
+    enabled: true
+  - rules:
+      - service: cockpit
+        network: "{{ public_net_name }}"
+        state: disabled
+      - service: dhcpv6-client
+        network: "{{ public_net_name }}"
+        state: disabled
+    enabled: "{{ 'public' in stackhpc_firewalld_zones }}"
```
markgoddard: nit: strictly this should be firewalld_zones (and elsewhere too)

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,413 @@
...
+  - rules:
+      - service: ssh
+        network: "{{ public_net_name }}"
+        state: disabled
+    enabled: "{{ 'public' in stackhpc_firewalld_zones && admin_oc_net_name | net_zone != 'public' }}"
```
markgoddard:
```suggestion
    enabled: "{{ 'public' in stackhpc_firewalld_zones and admin_oc_net_name | net_zone != 'public' }}"
```

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,413 @@
...
+# A templated list of firewalld rules, according to the enabled services,
+# host's group membership, and host's network configuration.
+# This variable can be used to set any *_firewalld_rules kayobe variable e.g.
+# compute_firewalld_rules: "{{ stackhpc_firewalld_rules }}" in compute.yml
+stackhpc_firewalld_rules: |
+  {% set stackhpc_firewalld_rules_verified = [] %}
+  {% for rule in stackhpc_firewalld_rules_unverified | unique %}
+  {% if rule | ansible.utils.remove_keys('state') in stackhpc_firewalld_rules_verified | map('ansible.utils.remove_keys', 'state') %}
+  {% set stackhpc_firewalld_rules_verified = 'Invalid configuration! Two matching firewalld rules exist with different states' + 1 %}
+  {% elif rule.network in network_interfaces and rule.network | net_zone is not none %}
```
markgoddard:
```suggestion
  {% elif rule.network in network_interfaces and rule.network | net_zone %}
```

etc/kayobe/inventory/group_vars/all/firewall
markgoddard: And do this elsewhere as well

etc/kayobe/inventory/group_vars/all/firewall
markgoddard, line 1: Do we need a corresponding etc/kayobe/firewall.yml with commented out variables?

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,413 @@
...
+  - rules:
+      - service: ssh
+        network: "{{ public_net_name }}"
+        state: disabled
+    enabled: "{{ 'public' in stackhpc_firewalld_zones && admin_oc_net_name | net_zone != 'public' }}"
+
+stackhpc_common_firewalld_rules_extra: []
+
+###############################################################################
+# Common firewalld rules
```
markgoddard: Controller

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,413 @@
...
+
+stackhpc_controller_firewalld_rules: "{{ stackhpc_controller_firewalld_rules_default + stackhpc_controller_firewalld_rules_extra }}"
+
+stackhpc_controller_firewalld_rules_default: |
+  {% set stackhpc_controller_firewalld_rules_formatted = [] %}
+  {% for ruleset in stackhpc_controller_firewalld_rules_template %}
+  {% if ruleset.enabled %}
```
markgoddard: `| bool`

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,413 @@
...
+  # Overcloud Ironic
+  - rules:
+      # Ironic inspector API
+      - port: 5050/tcp
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      # Ironic API
+      - port: 6385/tcp
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
```
markgoddard, line 140: I think these are covered by Kolla Ansible's firewall config. They aren't in the RUG config.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,413 @@
...
+  {% endfor %}
+  {{ stackhpc_storage_firewalld_rules_formatted }}
+
+stackhpc_storage_firewalld_rules_template:
+  # Ceph Prometheus exporter
+  - rules:
+      - port: 9283/tcp
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+    enabled: "{{ kolla_enable_prometheus_ceph_mgr_exporter and 'mgrs' in group_names }}"
```
markgoddard:
```suggestion
    enabled: "{{ kolla_enable_prometheus_ceph_mgr_exporter | bool and 'mgrs' in group_names }}"
```

markgoddard (changes requested):

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,446 @@
...
+        state: disabled
+      - service: dhcpv6-client
+        network: "{{ public_net_name }}"
+        state: disabled
+    enabled: "{{ public_net_name | net_zone in stackhpc_firewalld_zones }}"
+  - rules:
+      - service: ssh
+        network: "{{ public_net_name }}"
+        state: disabled
+    enabled: "{{ public_net_name | net_zone in stackhpc_firewalld_zones and admin_oc_net_name | net_zone != public_net_name | net_zone }}"
```
markgoddard: Strictly this should reference the literal `public` zone, since there's no guarantee that public network is using that zone. Possibly they need to reference the public zone explicitly rather than the public network. In which case you might need to specify this in the logic above:
```
rule.zone | default(rule.network | net_zone)
```

doc/source/configuration/firewall.rst
Alex-Welsh: I was hoping to. Whether I can get it to that point I'm not sure

doc/source/configuration/firewall.rst
Alex-Welsh: Moved

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Can do, but that should probably be one of the last jobs before merging, if not a follow up. The formatting is still a bit experimental for now

etc/kayobe/environments/ci-multinode/firewalld.yml
Alex-Welsh: I'll leave it as is for now because it's convenient for testing but move them all out before merging

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Yes, it's a bit messy but it gets displayed to users, I haven't yet found a way to include the rule

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done. Trying to get out of that bad habit

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Doh

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Yep, fixing the internal network issues resolved that

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -18,7 +18,7 @@ stackhpc_firewalld_zones: |
   {% set _ = network_zones.append({'zone': network | net_zone }) %}
   {% endif %}
   {% endfor %}
-  {{ network_zones }}
+  {{ network_zones | unique }}
```
markgoddard: These filters typically return a Python generator. It's usually better to cast to a list using `| list`

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -78,12 +78,12 @@ stackhpc_common_firewalld_rules_template:
...
         network: "{{ public_net_name }}"
         state: disabled
-    enabled: "{{ public_net_name | net_zone in stackhpc_firewalld_zones }}"
+    enabled: "{{ {'zone': public_net_name | net_zone} in firewalld_zones }}"
   - rules:
       - service: ssh
         network: "{{ public_net_name }}"
         state: disabled
-    enabled: "{{ public_net_name | net_zone in stackhpc_firewalld_zones and admin_oc_net_name | net_zone != public_net_name | net_zone }}"
+    enabled: "{{ {'zone': 'public' | net_zone} in firewalld_zones and admin_oc_net_name | net_zone != public_net_name | net_zone }}"
```
markgoddard: I still don't think this is right. We need to use the literal `public` zone rather than public_net_name

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -190,7 +157,7 @@ stackhpc_controller_firewalld_rules_template:
   # Octavia
   - rules:
       - port: 5555/udp
-        network: "{{ octavia_net_name }}"
+        network: "{{ octavia_net_name | default(public_net_name) }}"
```
markgoddard: I think let's remove the default and make the `enabled` flag dependent upon `octavia_net_name` being defined. If you're only using the OVN backend you don't need this network.

markgoddard (changes requested):

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Just for clarity, will the ```list``` filter also remove duplicates?
 Do I need to use it as well as ```unique``` or instead of?
 Should I apply the list filter to every instance of the "for loop" style variable definitions?

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+functionality of Kayobe.
+
+The firewall configuration is provided in
+``etc/kayobe/inventory/group_vars/all/firewall``.
+
+Enabling StackHPC firewalld rules
+=================================
+
+The standardised firewalld configuration is not used by default and must be
+actively opted into. This can be done as follows:
```
MaxBed4d:
```suggestion
The standardised firewalld configuration is not enabled by default and must be
actively opted into. This can be done by altering or copying ``etc/kayobe/<group>.yml`` to the chosen ``etc/kayobe/environment/<environment>/``, and making the appropriate following changes:
```

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+
+Controller firewalld Configuration
```
MaxBed4d, line 24:
```suggestion
Controller firewalld Configuration - ``etc/kayobe/controllers.yml``
```

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+
+   # A list of firewall rules to apply. Each item is a dict containing
+   # arguments to pass to the firewalld module. Arguments are omitted if not
+   # provided, with the following exceptions:
+   # - offline: true
+   # - permanent: true
+   # - state: enabled
+   controller_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+
+Compute firewalld Configuration
```
MaxBed4d, line 53:
```suggestion
Compute firewalld Configuration - ``etc/kayobe/compute.yml``
```

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+
+   # A list of firewall rules to apply. Each item is a dict containing
+   # arguments to pass to the firewalld module. Arguments are omitted if not
+   # provided, with the following exceptions:
+   # - offline: true
+   # - permanent: true
+   # - state: enabled
+   compute_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+
+Storage firewalld Configuration
```
MaxBed4d, line 82:
```suggestion
Storage firewalld Configuration - ``etc/kayobe/storage.yml``
```

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+
+   # A list of firewall rules to apply. Each item is a dict containing
+   # arguments to pass to the firewalld module. Arguments are omitted if not
+   # provided, with the following exceptions:
+   # - offline: true
+   # - permanent: true
+   # - state: enabled
+   storage_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+
+Monitoring firewalld Configuration
```
MaxBed4d, line 114:
```suggestion
Monitoring firewalld Configuration - ``etc/kayobe/monitoring.yml``
```

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+
+   # A list of firewall rules to apply. Each item is a dict containing
+   # arguments to pass to the firewalld module. Arguments are omitted if not
+   # provided, with the following exceptions:
+   # - offline: true
+   # - permanent: true
+   # - state: enabled
+   monitoring_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+
+Infrastructure VM firewalld Configuration
```
MaxBed4d, line 143:
```suggestion
Infrastructure VM firewalld Configuration - ``etc/kayobe/infra_vms.yml``
```

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+
+   # A list of firewall rules to apply. Each item is a dict containing
+   # arguments to pass to the firewalld module. Arguments are omitted if not
+   # provided, with the following exceptions:
+   # - offline: true
+   # - permanent: true
+   # - state: enabled
+   infra_vm_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+
+Seed firewalld Configuration
```
MaxBed4d, line 178:
```suggestion
Seed firewalld Configuration - ``etc/kayobe/seed.yml``
```

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+
+   # A list of firewall rules to apply. Each item is a dict containing
+   # arguments to pass to the firewalld module. Arguments are omitted if not
+   # provided, with the following exceptions:
+   # - offline: true
+   # - permanent: true
+   # - state: enabled
+   seed_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+
+Seed Hypervisor firewalld Configuration
```
MaxBed4d, line 207:
```suggestion
Seed Hypervisor firewalld Configuration - ``etc/kayobe/seed-hypervisor.yml``
```

doc/source/configuration/firewall.rst
Alex-Welsh: Some deployments don't use kayobe environments at all. I think I'll just remove the path from the file caption, since it doesn't matter whether it's in an env or not

doc/source/configuration/firewall.rst
Alex-Welsh: This is the title of the subsection. I don't think it needs the path in it since there's already a caption on the code block

etc/kayobe/inventory/group_vars/all/firewall
markgoddard: list won't remove duplicates, so you'll also need unique. You should apply it to every list variable defined using filters like unique, select, etc.

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+   # provided, with the following exceptions:
+   # - offline: true
+   # - permanent: true
+   # - state: enabled
+   seed_hypervisor_firewalld_rules: "{{ stackhpc_firewalld_rules }}"
+
+Custom rules
+------------
+
+Custom firewalld rules can be added for any of the following groups using their
```
MaxBed4d:
```suggestion
Custom firewalld rules can be added for any of the following groups within an environment's ``group_vars`` inventory, ``etc/kayobe/environments/<environment>/inventory/group_vars/<group>/firewall`` and using their
```

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,350 @@
...
+.. code-block:: yaml
+   :caption: ``etc/kayobe/kolla/globals.yml``
+
+   enable_external_api_firewalld: true
+   external_api_firewalld_zone: "{{ public_net_name | net_zone }}"
+
+Ensure every network in ``networks.yml`` has a zone defined. The standard
+configuration is to set the internal network zone to ``trusted`` and every
+other zone to the name of the network. See
+``etc/kayobe/environments/ci-multinode/networks.yml`` for a practical example.
```
MaxBed4d:
```suggestion
Prerequisites
----------------

Ensure Kolla Ansible opens up ports in firewalld for services on the public
API network:

.. code-block:: yaml
   :caption: ``etc/kayobe/kolla/globals.yml``

   enable_external_api_firewalld: true
   external_api_firewalld_zone: "{{ public_net_name | net_zone }}"

Ensure every network in ``networks.yml`` has a zone defined. The standard
configuration is to set the internal network zone to ``trusted`` and every
other zone to the name of the network. See
``etc/kayobe/environments/ci-multinode/networks.yml`` for a practical example.

Applying changes
----------------

The ``kayobe configuration dump`` command can be used to view all the rules
that will be applied to a host.

.. code-block:: bash

   kayobe configuration dump --var-name stackhpc_firewalld_rules --limit <host>

If the command above prints a template, rather than a clean list of rules, the
configuration is invalid. The kayobe configuration dump command can be used on
other variables such as ``stackhpc_firewalld_rules_unverified`` or
``stackhpc_*_firewalld_rules`` to debug the configuration. See the `How it
works`_ section for more details.
```

doc/source/configuration/firewall.rst
MaxBed4d:
```suggestion
The standardised firewalld configuration is not enabled by default and must be
actively opted into. This can be done by altering or, in the case of using a ``kayobe`` environment, copying ``etc/kayobe/<group>.yml`` to the chosen ``etc/kayobe/environment/<environment>/``, and making the appropriate following changes:
```

Seems odd that we reference paths and files in terms of kayobe environments later on, but not for this section, when it seems equally valid to do so.

doc/source/configuration/firewall.rst
Alex-Welsh: They don't actually have to be in group vars. Host vars, group vars, and extra vars are all valid. I'll update the doc to make that clearer

doc/source/configuration/firewall.rst
Alex-Welsh: Good point. I've gone one step further and split out the KA config and config validation

doc/source/configuration/firewall.rst
Alex-Welsh: Done

doc/source/configuration/firewall.rst
Alex-Welsh: I've updated the line to mention both environments and the base config

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: I've slapped it on the two big output variables ```stackhpc_firewalld_rules``` and ```stackhpc_firewalld_zones```, which I _think_ will do it. The loops have been behaving fine so far, but if you think I should add it inside them I can do that too

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Okay, I think I got it.

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: It doesn't like not having any default at all for ```octavia_net_name``` so I've made it blank. Otherwise, done

Alex-Welsh:
Okay, I've refactored things yet again, but the format is much more readable now and duplication has been reduced a lot.
The key to it is that now there is only one variable for extra rules - ```stackhpc_firewalld_rules_extra```. Different groups can still have different rules because it can just be defined in group vars. There's 2 main benefits:

1. Significantly reduced complexity
2. Custom rules can now have higher precedence, and default rules can be overridden

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -184,6 +184,15 @@ stackhpc_controller_firewalld_rules_template:
         network: "{{ octavia_net_name | default() }}"
         state: enabled
     enabled: "{{ kolla_enable_octavia | bool and octavia_net_name is not none }}"
+  # Mystery EIDF rules, required for creating VMs, need to check condition
```
jackhodgkiss: I don't think we should be naming customers within SKC.

As for these ports you can use

```
sudo lsof -i -P -n | grep LISTEN
```

To find what using these ports I assume either on the controllers/computes

jackhodgkiss: Port 67 is used by DHCP.

As for 6556 that port is used by [checkmk](https://checkmk.com/) which they do have a playbook for deploying.

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,440 @@
...
+Ensure Kolla Ansible opens up ports in firewalld for services on the public
+API network:
+
+.. code-block:: yaml
+   :caption: ``etc/kayobe/kolla/globals.yml``
+
+   enable_external_api_firewalld: true
+   external_api_firewalld_zone: "{{ public_net_name | net_zone }}"
+
+Ensure every network in ``networks.yml`` has a zone defined. The standard
```
markgoddard, line 264: Seems like this is in the wrong section.

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,440 @@
...
+        network: "{{ admin_oc_net_name }}"
+        zone: "{{ admin_oc_net_name | net_zone }}"
+        state: disabled
+
+Extra rules have higher precedence than the default rules but are not
+validated before being applied. Use with caution. If you need to add a custom
+rule, consider adding it to the default rule list with an appropriate boolean
+condition, and where possible merge your changes back into upstream SKC.
+
+Kolla-Ansible configuration
```
markgoddard: Duplicate of L230

doc/source/configuration/index.rst
```diff
@@ -21,3 +21,4 @@ the various features provided.
    ci-cd
    security-hardening
    cloudkitty
+   firewall
```
markgoddard: Move next to security-hardening?

doc/source/configuration/firewall.rst
```diff
@@ -0,0 +1,440 @@
...
+UDP port 1000 on the ``admin_oc`` network for the Wazuh manager Infrastructure
+VM:
+
+.. code-block:: yaml
+   :caption: ``etc/kayobe/inventory/group_vars/wazuh_manager/firewall``
+
+   stackhpc_firewalld_rules_extra:
+     -  service: ssh
+        network: "{{ provision_oc_net_name }}"
+        zone: "{{ provision_oc_net_name | net_zone }}"
```
markgoddard: Don't need to set network, but zone is required for a custom rule.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,323 @@
...
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      - port: 1514/udp
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      - port: 1515/tcp
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      - port: 443/tcp
+        network: "{{ public_net_name }}"
```
markgoddard: I don't think we can assume that the dashboard is on the public API network. How about this?
```suggestion
        network: "{{ wazuh_dashboard_net_name | default(provision_oc_net_name) }}"
```

markgoddard (commented):
I wrote a watchdog playbook when applying this:

```yaml
---
# This playbook can be applied in advance of rolling out a firewall
# configuration. It sets up a timer that disables the firewalld service after a
# period of time (default 600s). It should be used as follows:
# 1. Enable firewalld-watchdog
#    kayobe playbook run etc/kayobe/ansible/firewalld-watchdog.yml -l <hosts>
# 2. Apply firewall config
#    kayobe <group> host configure -l <hosts> -t network,firewall
# 3. Disable watchdog
#    kayobe playbook run etc/kayobe/ansible/firewalld-watchdog.yml -l <hosts> -e firewalld_watchdog_state=absent
# If the firewall changes result in being locked out of the system, the
# watchdog will disable the firewall after the timeout.
# Remember to disable the watchdog, otherwise the firewall will be disabled!

- name: Create a systemd timer to stop firewalld
  hosts: seed:seed-hypervisor:overcloud:infra-vms
  tags:
    - firewalld-watchdog
  vars:
    # Watchdog state: present or absent.
    firewalld_watchdog_state: present
    # Watchdog timeout in seconds.
    firewalld_watchdog_timeout_s: 600
  become: true
  tasks:
    - when: firewalld_watchdog_state == 'present'
      block:
        - name: Create firewalld-watchdog service unit file
          ansible.builtin.copy:
            dest: /etc/systemd/system/firewalld-watchdog.service
            content: |
              [Unit]
              Description=Firewalld watchdog service

              [Service]
              Type=oneshot
              ExecStart=/usr/bin/systemctl stop firewalld
          register: service_result

        - name: Create firewalld-watchdog timer unit file
          ansible.builtin.copy:
            dest: /etc/systemd/system/firewalld-watchdog.timer
            content: |
              [Unit]
              Description=Firewalld watchdog timer

              [Timer]
              OnActiveSec={{ firewalld_watchdog_timeout_s }}
              Unit=firewalld-watchdog.service

              [Install]
              WantedBy=timers.target
          register: timer_result

    - name: Enable or disable firewalld-watchdog timer
      ansible.builtin.systemd_service:
        name: firewalld-watchdog.timer
        daemon_reload: "{{ service_result is changed or timer_result is changed }}"
        enabled: false
        state: "{{ 'started' if firewalld_watchdog_state == 'present' else 'stopped' }}"

    - name: Remove firewalld-watchdog unit files
      ansible.builtin.file:
        path: "/etc/systemd/system/{{ item }}"
        state: absent
      loop:
        - firewalld-watchdog.service
        - firewalld-watchdog.timer
      when: firewalld_watchdog_state == 'absent'
```

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,323 @@
...
+  {% endfor %}
+  {% endif %}
+  {% endfor %}
+  {{ undef(hint='ERROR: Conflicting firewall rules found') if ({'state':'failure'} in stackhpc_firewalld_rules_formatted) else stackhpc_firewalld_rules_formatted }}
+
+stackhpc_firewalld_rules_template: |
+  {{ stackhpc_common_firewalld_rules_template +
+     (stackhpc_controller_firewalld_rules_template if 'controllers' in group_names else []) +
+     (stackhpc_compute_firewalld_rules_template if 'compute' in group_names else []) +
+     (stackhpc_storage_firewalld_rules_template if 'storage' in group_names else []) +
```
markgoddard: Consider changing to this to avoid issues when Ceph runs on controllers or computes:

```
 (stackhpc_ceph_firewalld_rules_template if 'ceph' in group_names else []) +
```

Alex-Welsh: Done

etc/kayobe/inventory/group_vars/all/firewall
Alex-Welsh: Done

doc/source/configuration/firewall.rst
Alex-Welsh: Added a new heading above

doc/source/configuration/firewall.rst
Alex-Welsh: I've updated the wording here to be more specific about differences between custom and default rules

doc/source/configuration/firewall.rst
Alex-Welsh: Durr

doc/source/configuration/index.rst
Alex-Welsh: Had a shuffle, all the security stuff is now grouped

Alex-Welsh:
Known issues:
Mystery DHCP rule for VM access
Failures in OVS deployments

markgoddard (commented):
When testing this in one deployment we saw an issue where seed/infra VM networking was blocked by the seed hypervisor firewall. We worked around it like this:

```
seed_hypervisor_sysctl_parameters:
  # By default this is 1, which causes layer 2 traffic flowing through Linux
  # bridges to pass through iptables. This blocks traffic from VMs (seed, wazuh) to
  # the Internet.
  net.bridge.bridge-nf-call-iptables: 0
```

This should be documented.

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,323 @@
...
+  - rules:
+      - service: ceph
+        network: "{{ storage_net_name }}"
+        state: enabled
+      - service: ceph
+        network: "{{ storage_mgmt_net_name }}"
+        state: enabled
+      - service: ceph-mon
+        network: "{{ storage_net_name }}"
+        state: "{{ 'enabled' if 'mons' in group_names else 'disabled' }}"
```
jackhodgkiss, line 235:
```suggestion
      - service: ceph-mon
        network: "{{ storage_net_name }}"
        state: "{{ 'enabled' if 'mons' in group_names else 'disabled' }}"
      - port: 8100/tcp
        network: "{{ storage_net_name }}"
        state: "{{ 'enabled' if 'rgws' in group_names else 'disabled' }}"
```

mnasiadka: What is port 8100/tcp? Should we at least have a comment on what it is?

etc/kayobe/inventory/group_vars/all/firewall
```diff
@@ -0,0 +1,323 @@
...
+# This variable can be used to set any *_firewalld_zones kayobe variable e.g.
+# compute_firewalld_zones: "{{ stackhpc_firewalld_zones }}" in compute.yml
+stackhpc_firewalld_zones: |
+  {% set network_zones = [] %}
+  {% for network in network_interfaces %}
+  {% if network | net_zone %}
+  {% set _ = network_zones.append({'zone': network | net_zone }) %}
+  {% endif %}
+  {% endfor %}
+  {{ network_zones | unique | list }}
```
MoteHue, line 21: This final `| list` probably doesn't need to be here.

markgoddard: Filters like unique tend to return a Python generator rather than a proper list, and they don't always act as you would expect.

MoteHue: Interesting, can leave as is then

etc/kayobe/inventory/group_vars/all/firewall
jovial: I think it comes from this bit of config:
```
cephadm_radosgw_services:
  - id: myrgw
    count_per_host: 1
    spec:
      rgw_frontend_port: 8100
  ```
  in the[ docs](https://github.com/stackhpc/stackhpc-kayobe-config/blob/97f8975baefa19f17bc97fd6a887a3bd2acf4df6/doc/source/configuration/cephadm.rst#rados-gateways) and shows up as:
  ```
  stack@controller1:~$ sudo ss -nlpt | grep 8100
LISTEN 0      128                  0.0.0.0:8100       0.0.0.0:*    users:(("radosgw",pid=44178,fd=107))
LISTEN 0      128                     [::]:8100          [::]:*    users:(("radosgw",pid=44178,fd=108))
  ```
  
so might need to be factored out into a variable.  Are we changing this to a non-default port? Was there a conflict?

File: etc/kayobe/inventory/group_vars/all/firewall [hunk 74]
```diff
@@ -0,0 +1,323 @@
...
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      - port: 1514/udp
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      - port: 1515/tcp
+        network: "{{ provision_oc_net_name }}"
+        state: enabled
+      - port: 443/tcp
+        network: "{{ wazuh_dashboard_net_name | default(provision_oc_net_name) }}"
```
jovial, line 261: `wazuh_dashboard_net_name` is undocumented

doc/source/configuration/cephadm.rst
```diff
@@ -344,6 +344,8 @@ should be used in the Kolla Manila configuration e.g.:
 
   manila_cephfs_filesystem_name: manila-cephfs
 
+.. _RGWs-Ceph:
```
jovial:
```suggestion
.. _RGWs-With-Ceph:
```

etc/kayobe/inventory/group_vars/all/firewall
jovial: This has now merged

jovial (commented):
Debug only when firewalld is enabled

We are missing the firewall-cmd executable

etc/kayobe/inventory/group_vars/all/firewall
jovial: It's now merged

jovial (approved):
This worked pretty nicely for me.  I like the way that the rules take into account multiple group membership - really nice job. Lets get this in before it gets stale. It's disabled by default and we can iterate on it once it is.

I've made a jira to fix to the OVS AIO: INFRA-818

PR #1018, opened 2024-04-05
INFRA-629: Update dashboards to use non-deprecated plugin

dougszumski:
No functional change where possible. 

1. Update plugins to fix deprecation notices
2. Remove panels wheres metrics no longer available ( OpenSearch)
3. Check in latest Grafana definition of the dashboard to reduce future diffs

Drink a strong coffee before even considering looking at the json diffs.


Alex-Welsh:
Is this the sort of thing that should go into yoga and get merged up?

dougszumski:
> Is this the sort of thing that should go into yoga and get merged up?

A good question and it's a bit of grey area to me. 

Personally I don't think we should do a major upgrade of Grafana/anything in Yoga, in which case, we don't need these to be backported. Grafana should remain on 10.x in Yoga.

However, Grafana is unpinned in Kolla, so if someone rebuilds the Grafana (Yoga) image in release train, it will get upgraded to 11.x in May 2024? Then we would want to have this PR targeted to Yoga. Ideally I think we should avoid that, unless there is a good reason not to.

Alex-Welsh:
> > Is this the sort of thing that should go into yoga and get merged up?
> 
> A good question and it's a bit of grey area to me.
> 
> Personally I don't think we should do a major upgrade of Grafana/anything in Yoga, in which case, we don't need these to be backported. Grafana should remain on 10.x in Yoga.
> 
> However, Grafana is unpinned in Kolla, so if someone rebuilds the Grafana (Yoga) image in release train, it will get upgraded to 11.x in May 2024? Then we would want to have this PR targeted to Yoga. Ideally I think we should avoid that, unless there is a good reason not to.

I think we'll have to assume that we'll have to rebuild the yoga images at some point due to a security backport. What we do to prepare for that I'm not exactly sure

markgoddard:
I'd probably rather that we pin grafana version to 10.x in yoga than replace all the dashboards.

technowhizz:
One thing to add is we should also aim in this re-do to have all the time frames the same (ie the time window in grafana when looking at a graph.) I think the past 24 hours for all is reasonable? @dougszumski @assumptionsandg

dougszumski (commented):
I've loaded these in to an aio running Grafana 10.4.2 and what we have so far looks fine.

dougszumski:
I see the AngularJS warnings fixed in the above Ceph commits

dougszumski:
Let's merge this and follow up with the remaining dashboards

PR #1028, opened 2024-04-12
Add generic blackbox monitoring configuration

Alex-Welsh:


etc/kayobe/kolla/globals.yml
```diff
@@ -69,5 +69,131 @@ prometheus_ceph_mgr_exporter_endpoints:
...
+  - endpoints:
+      - "barbican:os_endpoint:{{ kolla_public_endpoint }}:{{ barbican_api_public_port }}"
+      - "{{ ('barbican_internal:os_endpoint:' + kolla_internal_endpoint + ':' + barbican_api_port) if kolla_public_endpoint != kolla_internal_endpoint }}"
+    enabled: "{{ enable_barbican | bool }}"
+  - endpoints:
+      - "blazar:os_endpoint:{{ kolla_public_endpoint }}:{{ blazar_api_public_port }}"
+      - "{{ ('blazar_internal:os_endpoint:' + kolla_internal_endpoint + ':' + blazar_api_port) if kolla_public_endpoint != kolla_internal_endpoint }}"
+    enabled: "{{ enable_blazar | bool }}"
+  - endpoints:
+      - "ceph_rgw:os_endpoint:{{ kolla_public_endpoint }}:{{ ceph_rgw_public_port }}"
```
priteau: Have you verified this works? This would hit the S3 API which I am not sure would work with the `os_endpoint` module, since it doesn't contain `versions`.

Alex-Welsh: That one is untested. I think you are right though so will try `http_2xx`

Alex-Welsh: Done

etc/kayobe/kolla/globals.yml
```diff
@@ -69,5 +69,131 @@ prometheus_ceph_mgr_exporter_endpoints:
 {% endfor %}
 {% endif %}
 
-#############################################################################
+
+########################################
```
markgoddard: Definitely too big for Yoga. Merge conflict in Antelope will be annoying but possibly worth it. Second merge conflict in Caracal when if gets moved into k-a will also be annoying :)

etc/kayobe/kolla/globals.yml
```diff
@@ -69,5 +69,131 @@ prometheus_ceph_mgr_exporter_endpoints:
...
+# Prometheus blackbox monitoring options
+########################################
+{% raw %}
+kolla_internal_endpoint: "{{ internal_protocol }}://{{ kolla_internal_fqdn | put_address_in_context('url') }}"
+kolla_public_endpoint: "{{ public_protocol }}://{{ kolla_external_fqdn | put_address_in_context('url') }}"
```
markgoddard: This won't work with a single frontend port (i.e. hostname-based URLs).

etc/kayobe/kolla/globals.yml
markgoddard: One way around this could be to use a separate group_vars file in etc/kayobe/kolla/inventory/group_vars/prometheus-blackbox-exporter to define the rules.

Alex-Welsh: Moved to a new file and rebased on antelope

etc/kayobe/kolla/globals.yml
Alex-Welsh: I've gone with the same format as the upstream version, and included definitions for any missing endpoints

Alex-Welsh:
Marked as draft as recent changes are untested

etc/kayobe/kolla.yml
```diff
@@ -443,7 +443,8 @@ kolla_build_args:
 # List of names of additional host variables to pass through from kayobe hosts
 # to kolla-ansible hosts, if set. See also
 # kolla_overcloud_inventory_pass_through_host_vars_map.
-#kolla_overcloud_inventory_pass_through_host_vars_extra:
+kolla_overcloud_inventory_pass_through_host_vars_extra:
```
Alex-Welsh: I can't say I like this approach but I don't know of a better way to get kayobe to template some of the endpoints

markgoddard: Could put it in kolla/globals.yml?

Alex-Welsh: Yup. Annoyingly I tried exactly that a while ago but it failed for an unrelated reason. Fixed now

dougszumski (commented):
I'm not opposed to merging this as-is, but it would be good to investigate the file based service discovery approach I mentioned upstream. We should be able to do something similar downstream, but using Kayobe to write out the service discovery file(s) instead of the KA roles.

bbezak:
is it ready to merge? it would be good to define blackbox exporter defaults

Alex-Welsh:
> is it ready to merge? it would be good to define blackbox exporter defaults

From the monitoring R&D discussion, we're going to proceed with this style downstream, and investigate file based discovery upstream

dougszumski (approved):
Good stuff @Alex-Welsh

PR #1029, opened 2024-04-15
Fix tox whitespace warning

Alex-Welsh:
See [here](https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/8649973848/job/23717449450?pr=1016#step:5:22) for an example of the warning in CI

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -369,7 +369,7 @@ kolla_build_customizations: "{{ kolla_build_customizations_common | combine(koll
 
 # Dict mapping Kolla Dockerfile ARG names to their values.
 kolla_build_args:
-  node_exporter_version: "1.5.0" # kolla has 1.4.0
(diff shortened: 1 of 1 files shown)


etc/kayobe/kolla.yml
```diff
@@ -369,7 +369,7 @@ kolla_build_customizations: "{{ kolla_build_customizations_common | combine(koll
 
 # Dict mapping Kolla Dockerfile ARG names to their values.
 kolla_build_args:
-  node_exporter_version: "1.5.0" # kolla has 1.4.0
+  node_exporter_version: "1.5.0"  # kolla has 1.4.0
```
markgoddard, line 372: How could this terrible mistake have slipped through?

markgoddard (approved):

etc/kayobe/kolla.yml
Alex-Welsh: Really catastrophic I know. Disaster averted :relieved:

PR #1031, opened 2024-04-15
Add retries to overcloud host image pulp tasks

markgoddard:
Retries have been added to the stackhpc.pulp collection to improve
reliability. Adding the same here.

### DIFF
--- a/etc/kayobe/ansible/pulp-host-image-upload.yml
+++ b/etc/kayobe/ansible/pulp-host-image-upload.yml
@@ -25,6 +25,10 @@
         password: "{{ remote_pulp_password }}"
         file: "{{ found_files.files[0].path }}"
         state: present
+      register: upload_result
+      until: upload_result is success
+      retries: 3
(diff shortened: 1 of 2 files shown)


etc/kayobe/ansible/pulp-host-image-upload.yml
```diff
@@ -25,6 +25,10 @@
         password: "{{ remote_pulp_password }}"
         file: "{{ found_files.files[0].path }}"
         state: present
+      register: upload_result
+      until: upload_result is success
+      retries: 3
+      delay: 5
```
Alex-Welsh: The other plays are fine because they're metadata but this one might benefit from a longer delay since it's trying to upload a relatively large artifact. Maybe a minute or two?

markgoddard: Done

Alex-Welsh (approved):

PR #1033, opened 2024-04-16
Raise alert on degraded network bonds

priteau:
This will raise a alert when at least one of the bond members is down. Adapted from awesome-prometheus-alerts [1].

[1] https://samber.github.io/awesome-prometheus-alerts/rules.html#rule-host-and-hardware-1-34


JohnGarbutt (approved):
This is a very good idea. It looks to me like this should noop where there is no bond.

jovial (approved):

markgoddard (approved):

cityofships:
good stuff

dougszumski (commented):
Nice addition

PR #1035, opened 2024-04-17
CI: Register overcloud host image as shared and add stackhpc-dev

markgoddard:
In order for developers to be able to use overcloud images for testing,
they must be available in the stackhpc-dev project on Leafcloud. To do
this we can make the image shared, and add stackhpc-dev as a member.
Unfortunately someone in the stackhpc-dev project must also accept the
image membership before it becomes active. It's not possible to do this
with the existing app creds which are scoped to the stackhpc-ci project.
Instead we provide instructions on how to do it as a debug message.

### DIFF
--- a/.github/workflows/overcloud-host-image-build.yml
+++ b/.github/workflows/overcloud-host-image-build.yml
@@ -472,7 +472,7 @@ jobs:
         run: |
           source venvs/kayobe/bin/activate &&
           source src/kayobe-config/kayobe-env --environment ci-builder &&
-          kayobe playbook run \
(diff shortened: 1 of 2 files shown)


Alex-Welsh:
@MaxBed4d could you add this task to your backlog? It should be quite an easy change since openstack.cloud collection has been updated now. Just re-target this at Caracal, fix the FIXME comments, and make the project ID variable (for SMS/Leafcloud)

Alex-Welsh:
Superseded by https://github.com/stackhpc/stackhpc-kayobe-config/pull/1631

PR #1038, opened 2024-04-18
fix: conditional error when no seed exists

jackhodgkiss:
In situations where there is no `seed` the `seed-hypervisor` will fail a  host configure against `no_proxy` settings due to the conditional  attempting to access attributes that do not exist.

### DIFF
--- a/etc/kayobe/inventory/group_vars/seed-hypervisor/proxy.yml
+++ b/etc/kayobe/inventory/group_vars/seed-hypervisor/proxy.yml
@@ -1,0 +1,12 @@
+---
+###############################################################################
+# Configuration of HTTP(S) proxies.
+
+# The default value whilst only applied if `http_proxy` and `https_proxy` are set
+# will not work as the conditional fails to evaluate if the seed node does not exist.
+# This makes the `seed-hypervisor` reliant on the seed node.
+no_proxy: []
+
+###############################################################################
+# Dummy variable to allow Ansible to accept this file.
+workaround_ansible_issue_8743: yes


etc/kayobe/inventory/group_vars/seed-hypervisor/proxy.yml
```diff
@@ -0,0 +1,12 @@
+---
+###############################################################################
+# Configuration of HTTP(S) proxies.
+
+# The default value whilst only applied if `http_proxy` and `https_proxy` are set
+# will not work as the conditional fails to evaluate if the seed node does not exist.
+# This makes the `seed-hypervisor` reliant on the seed node.
+no_proxy: []
```
markgoddard, line 8: I'm not sure how this would help. The walled garden docs only mention setting no_proxy for hosts in the overcloud group.

markgoddard (changes requested):

etc/kayobe/inventory/group_vars/seed-hypervisor/proxy.yml
jackhodgkiss: The problem is `no_proxy` still needs to evaluated for the conditional which means attempting to expand the default value for which it cannot due to the lack of the seed-vm at the time of creating a seed-hypervisor.

https://github.com/openstack/kayobe/blob/ece51deb3bff2b639ceb6378204a2b9d32b2de36/ansible/proxy.yml#L40

markgoddard: Ah, ok. Kayobe default for no_proxy is:

```
no_proxy:
  - "127.0.0.1"
  - "localhost"
  - "{{ ('http://' ~ docker_registry) | urlsplit('hostname') if docker_registry else '' }}"
```

So this comes from `docker_registry` -> `stackhpc_docker_registry` -> `pulp_url` -> `groups.seed.0`.

So I think we should either make pulp_url conditional on `groups.seed | length > 0`, or document the change you've made in this PR in the walled garden guide. We don't have any other walled garden config in by default, so I don't think it makes sense to add this.

jackhodgkiss: I will look at making `pulp_url` conditional. Thanks.

Alex-Welsh:
@jackhodgkiss any updates on this?

etc/kayobe/inventory/group_vars/seed-hypervisor/proxy.yml
Alex-Welsh: @jackhodgkiss any updates on this?

PR #1040, opened 2024-04-19
Ceph dashboard osd usage

technowhizz:
Not entirely happy with the way Grafana shows the histogram so if anyone has any suggestions feel free to let me know.

There is also a bug with the threshold colouring. It doesn't show up until you edit the panel and change something appearance related like opacity.

image

### DIFF
--- a/etc/kayobe/kolla/config/grafana/dashboards/ceph/ceph_osds.json
+++ b/etc/kayobe/kolla/config/grafana/dashboards/ceph/ceph_osds.json
@@ -885,6 +885,150 @@
       }
     },
     {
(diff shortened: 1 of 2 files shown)


technowhizz:
Updated the graphs to look like this: 
image

Unfortunately it seems that the colouring is broken in grafana. Additionally it appears that in the histograms you can no longer label the y-axis

dougszumski (approved):
Thanks Dawud, let's hope the histogram gets fixed

Alex-Welsh:
@technowhizz are you happy to merge this?

PR #1041, opened 2024-04-22
[prometheus] Rename network interfaces to network names

tomclark0:
Prometheus Node Exporter and cAdvisor both provide network statistics using the interface/device names. This play causes Prometheus to relabel these fields to human readable names based on the networks as defined in kayobe  e.g. bond1.1838 may become storage_network.
The default labels are preserved with the prefix 'original_'.

Prometheus.yml.j2 based on upstream master


etc/kayobe/ansible/prometheus-network-names.yml
```diff
@@ -0,0 +1,40 @@
+- name: Prometheus friendly network names
+  hosts: overcloud
```
dougszumski, line 2: We may need to expand this based on https://stackhpc.atlassian.net/browse/DEV-404 (I'm trying to figure out what happened to that), but we can cover it under a separate PR. It's currently correct.

dougszumski (commented):
Excellent, this works well.

Please could you add a release note to https://github.com/stackhpc/stackhpc-kayobe-config/tree/stackhpc/2023.1/releasenotes/notes

You can use `reno new add-friendly-interface-names` or something similar from the root of Kayobe config.

Please also add a small doc section to `https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2023.1/doc/source/configuration/monitoring.rst`?

etc/kayobe/ansible/prometheus-network-names.yml
```diff
@@ -0,0 +1,40 @@
...
+            {%- set friendly_name = i -%}
+            {%- set unique_id = device_name ~ '_' ~ friendly_name -%}
+            {%- set _ = if_list.append({
+              'unique_id': unique_id,
+              'device_name': device_name,
+              'friendly_name': friendly_name
+            }) -%}
+          {%- endfor -%}
+          {{ if_list }}
+      delegate_to: "{{ inventory_hostname }}"
```
markgoddard: I think this line is a noop.

etc/kayobe/ansible/prometheus-network-names.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    - name: Deduplicate the aggregated list based on unique IDs
+      set_fact:
+        prometheus_network_maps_blob: "{{ prometheus_network_maps_aggregated | unique(attribute='unique_id') }}"
+      run_once: true
+      delegate_to: localhost
+
+    - name: Apply template with deduplicated network maps
+      ansible.builtin.template:
+        src: prometheus.yml.j2
```
markgoddard, line 35: :grimacing: triple templating

etc/kayobe/ansible/prometheus-network-names.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    - name: Aggregate network maps from all hosts
+      set_fact:
+        prometheus_network_maps_aggregated: "{{ groups['overcloud'] | map('extract', hostvars, 'host_network_maps') | flatten }}"
+      run_once: true
+      delegate_to: localhost
+
+    - name: Deduplicate the aggregated list based on unique IDs
+      set_fact:
+        prometheus_network_maps_blob: "{{ prometheus_network_maps_aggregated | unique(attribute='unique_id') }}"
```
markgoddard, line 29: The issue with this approach is that the map is global, but different hosts may have different uses for interfaces of the same name. It'll probably work fine in most cases, but could lead to confusing results if there's a conflict. The need to deduplicate highlights the issue.

etc/kayobe/ansible/prometheus.yml.j2
```diff
@@ -0,0 +1,287 @@
...
+        labels:
+          instance: "{{ hostvars[host].prometheus_instance_label }}"
+{% endif %}
+{% endfor %}
+{% endraw %}
+    metric_relabel_configs:
+     - replacement: $1
+       source_labels: [device]
+       target_label: 'original_device'
+{% for net_map in prometheus_network_maps_blob %}
```
markgoddard, line 52: Kolla ansible does have access to at least some of this data, although the networks don't have quite the same names as those in kayobe config.

```
api_interface
storage_interface
tunnel_interface
```
etc.

Of course the value of these may vary between hosts.

markgoddard (changes requested):

etc/kayobe/ansible/prometheus.yml.j2
```diff
@@ -0,0 +1,287 @@
+{{ '{%' }} raw {{ '%}' }}
```
mnasiadka, line 1: Sorry to crash the party - but that means we're going to maintain our version of prometheus.yml.j2 - and every time some new feature comes out in upstream land - we need to adapt it?

etc/kayobe/ansible/prometheus-network-names.yml
tomclark0: Agreed. The only solution I saw to this issue would be to have a seperate scrape_config job for each distinct host network config.
This would also resolve the need to deduplicate.
I had thought this could be implemented with support for discrete files per scrape_config, which would also reduce the maintance burden for the templates.
This is now supported in prometheus upstream https://github.com/prometheus/prometheus/pull/12019 but would be outside the scope of this MR.

markgoddard: I think I recall reading about the impact of relabelling on performance. I expect there is some optimisation that could be applied to limit the number of relabel configs but a simple approach would be O(n^2) with the number of hosts.

etc/kayobe/ansible/prometheus.yml.j2
dougszumski: In the near term we can switch to separate scrape configs to reduce the burden (supported from `2.43.0`). We just need to add support for it in KA (the binary was only updated the other day).

https://github.com/prometheus/prometheus/issues/8543

We only need to maintain a small fraction of the file then.

etc/kayobe/ansible/prometheus.yml.j2
dougszumski: This was considered - the issue is that only some of the info is there as you point out.

etc/kayobe/ansible/prometheus-network-names.yml
dougszumski: The other approach considered was to have a pull through wrapper around node-exporter, that could augment specific metrics with additional labels based on some static, host specific config. All other metrics would just pull through unaffected.

That would solve the issue with host specific network interface naming, and re-labelling complexity, but it's a larger task.  It could follow on from INFRA-388.

There was also the approach of just naming network interfaces with a human readable name, rather than the VLAN, but it was considered too invasive (although supported on Ubuntu as I understand).

I would prefer to put effort into the node exporter wrapper than try and handle all the corner cases with this approach. It should be fine in many cases and is opt-in. We do need to make clear the limitations.

etc/kayobe/ansible/prometheus-network-names.yml
tomclark0: Thanks for the catch.

doc/source/configuration/monitoring.rst
```diff
@@ -169,3 +169,38 @@ If you notice ``HaproxyServerDown`` or ``HaproxyBackendDown`` prometheus
...
+following play can be run. This takes network names as defined in kayobe and 
+relabels the devices/interfaces in Prometheus to make use of these names.
+
+
+.. code-block:: console
+
+    kayobe playbook run etc/kayobe/ansible/prometheus-network-names.yml
+    kayobe overcloud service reconfigure --kt prometheus
+
+This first generates a template based on the promtheus.yml.j2
```
markgoddard:
```suggestion
This first generates a template based on the prometheus.yml.j2 
```

doc/source/configuration/monitoring.rst
```diff
@@ -169,3 +169,38 @@ If you notice ``HaproxyServerDown`` or ``HaproxyBackendDown`` prometheus
 alerts after deployment it's likely the os_exporter secrets have not been
 set correctly, double check you have entered the correct authentication
 information appropiate to your cloud and re-deploy.
+
+Friendly Network Names
+=======================
+For operators that prefer to see descriptive or friendly interface names the
```
markgoddard: I guess it doesn't need to be perfect if it's an opt-in thing.

doc/source/configuration/monitoring.rst
```diff
@@ -169,3 +169,38 @@ If you notice ``HaproxyServerDown`` or ``HaproxyBackendDown`` prometheus
...
+    kayobe playbook run etc/kayobe/ansible/prometheus-network-names.yml
+    kayobe overcloud service reconfigure --kt prometheus
+
+This first generates a template based on the promtheus.yml.j2 
+``etc/kayobe/ansible/`` and which is further templated for use with
+kolla-ansible.
+
+
+This helps Prometheus provide insights that can be more easily understood by 
+those without an intimate understanding of a given site.Prometheus Node Exporter
```
markgoddard:
```suggestion
those without an intimate understanding of a given site. Prometheus Node Exporter
```

doc/source/configuration/monitoring.rst
```diff
@@ -169,3 +169,38 @@ If you notice ``HaproxyServerDown`` or ``HaproxyBackendDown`` prometheus
...
+ and cAdvisor both provide network statistics using the
+interface/device names. This play causes Prometheus to relabel these fields to
+ human readable names based on the networks as defined in kayobe 
+ e.g. bond1.1838 may become storage_network.
+
+The default labels are preserved with the prefix ``original_``.
+For node_exporter, ``device`` is then used for network names, while
```
markgoddard: Presumably this is to avoid changing the dashboards?

etc/kayobe/ansible/prometheus.yml.j2
markgoddard: Perhaps if kolla-ansible provided variables that allow you to define relabel configs for each scrape config that would allow us to avoid the custom template?

markgoddard (changes requested):

doc/source/configuration/monitoring.rst
tomclark0: Exactly, no changes required on any dashboard. Alerts can include the interface name since for host troubleshooting that's likely to be a lot more useful.

etc/kayobe/ansible/prometheus.yml.j2
tomclark0: I think we would need a high level of customisation to cover all use cases. 
To cover reuse of network devices for different purposes across different nodes, a new scrape_config for the different nodes would be needed. 

I'm not sure if seperate scrape_config files, possibly being templated similarly to haproxy services, would offer a meaningful alternative.

doc/source/configuration/monitoring.rst
```diff
@@ -169,3 +169,38 @@ If you notice ``HaproxyServerDown`` or ``HaproxyBackendDown`` prometheus
 alerts after deployment it's likely the os_exporter secrets have not been
 set correctly, double check you have entered the correct authentication
 information appropiate to your cloud and re-deploy.
+
+Friendly Network Names
+=======================
+For operators that prefer to see descriptive or friendly interface names the 
+following play can be run. This takes network names as defined in kayobe and 
+relabels the devices/interfaces in Prometheus to make use of these names.
```
dougszumski, line 177: Please could we add a warning note here for anyone thinking about enabling it? Something like:

- The change creates new time series by changing labels. This will mean that for a period equal to the retention time, graphs may appear disjointed.
- Any custom Grafana or Prometheus configuration added by the customer may be affected by the relabelling
- Billing systems monitoring network traffic may need to be updated  
- The feature currently requires that network interface names are consistent across each network
- Before turning on the feature, it should be discussed with the customer. All points above must be considered.

doc/source/configuration/monitoring.rst
```diff
@@ -169,3 +169,38 @@ If you notice ``HaproxyServerDown`` or ``HaproxyBackendDown`` prometheus
...
+
+The default labels are preserved with the prefix ``original_``.
+For node_exporter, ``device`` is then used for network names, while 
+``original_device`` is used for the interface itself.
+For cAdvisor, ``interface`` is used for network names, and ``original_interface``
+is used to preserve the interface name.
+
+**Known-Limitations/Untested**
+The current implementation does not cover known edge cases:
+* Reusing the same network devices under a different network name.
```
dougszumski: Include interface naming requirement

etc/kayobe/ansible/prometheus.yml.j2
dougszumski: A good idea @markgoddard - it would also be useful for CloudKitty. I have created https://stackhpc.atlassian.net/browse/INFRA-668

The conclusion in the R&D meeting is that we should land this feature and iterate on it.

doc/source/configuration/monitoring.rst
dougszumski: There will also be future iterations as time allows.

doc/source/configuration/monitoring.rst
```diff
@@ -169,3 +169,67 @@ If you notice ``HaproxyServerDown`` or ``HaproxyBackendDown`` prometheus
...
+  ``original_interface`` is used to preserve the interface name.
+
+:Known-Limitations/Considerations/Requirements:
+
+Before enabling this feature, the implications must be discussed with the 
+customer. The following are key considerations for that conversation:
+
+* Only network names defined within kayobe are within scope.
+* Tenant network interfaces, including SR-IOV are not considered or modified.
+* In the case of bonded interfaces, only the bond itself is relablled.
```
dougszumski: nit: s/relablled/relabelled

only the bond *and interfaces tagged on it*?

dougszumski (commented):
Thanks Tom. I'm happy to merge this and iterate on it, if others are happy?

dougszumski:
nit: pep8 is complaining about whitespace

dougszumski (approved):

tomclark0:
@markgoddard - Are you happy to proceed with this as an opt-in?

markgoddard (approved):

PR #1045, opened 2024-04-23
Fix Barbican templating

Alex-Welsh:
The existing template removes the newline after {% endraw%}, so the two config options get stacked onto the same line, breaking Barbican

### DIFF
--- a/etc/kayobe/environments/ci-multinode/kolla/config/barbican.conf
+++ b/etc/kayobe/environments/ci-multinode/kolla/config/barbican.conf
@@ -7,7 +7,9 @@ enabled_secretstore_plugins=vault_plugin
 [vault_plugin]
 vault_url = https://{{ kolla_internal_vip_address }}:8200
 use_ssl = True
-ssl_ca_crt_file = {% raw %}{{ openstack_cacert }}{% endraw %}
+{% raw %}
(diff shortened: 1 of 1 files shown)


markgoddard (commented):
Should it go to yoga first?

Alex-Welsh:
> Should it go to yoga first?

@MoteHue Suggested that it could be to do with the config merging in Antelope. If you think it would affect yoga as well, then I'll rebase

MoteHue:
> > Should it go to yoga first?
> 
> @MoteHue Suggested that it could be to do with the config merging in Antelope. If you think it would affect yoga as well, then I'll rebase

This isn't due to the config merging, it's just how {% raw %} tags work. The similar "bug" I've seen is that defining an option within an environment and a base config file will cause the lines to be duplicated.

Personally, I'd change this to the following because it's not immediately obvious that we'd need to keep that newline hanging around.

```
{% raw %}
ssl_ca_crt_file = {{ openstack_cacert }}
{% endraw %}
```

Might be able to do something with symbols in the tags too? 
`{% endraw +%}` perhaps?

Alex-Welsh:
I've Yoga'd it and moved around the tags

markgoddard (approved):

PR #1046, opened 2024-04-23
Use Openstack dashboard as Grafana home dashboard

darmach:
Replaced default home dashboard in Grafana with Openstack Dashboard. Disabled annotations of default home dashboard.


technowhizz:
Will this mean that we wont be able to see the current firing alerts in Grafana?

dougszumski:
Thanks @darmach.  @technowhizz is right, it would be a shame to lose the alerts. I will add them in another commit to this PR. We also need to pull in the changes from [Jake](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1018/commits/ad41be1062f7df742ccee712a0c6105ab818d496#diff-35a13ca5d0e44e1f09e0e9fd6795317652893c8fdd0890089a672fd30c5b4b0a). I can handle that . We should also avoid have two copies.

dougszumski:
> Thanks @darmach. @technowhizz is right, it would be a shame to lose the alerts. I will add them in another commit to this PR. We also need to pull in the changes from [Jake](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1018/commits/ad41be1062f7df742ccee712a0c6105ab818d496#diff-35a13ca5d0e44e1f09e0e9fd6795317652893c8fdd0890089a672fd30c5b4b0a). I can handle that . We should also avoid have two copies.

That took far longer  than expected. I tried including a list of dashboards in the overview page as before, but the 'search' folder option in the widget references dashboards by folder UID. The UIDs change when the dashboard is loaded and you end up with an error / non-working table:

Screenshot from 2024-04-26 17-04-53

dougszumski:
Please could someone test the update works? I'm using Grafana v10.4.2.

I think there's lots more we could do with this home dashboard, but this is a step in a better direction at least.

Alex-Welsh:
@dougszumski What do we need to do to move this PR forwards? It seems relatively simple but has been sitting for a while and has conflicts

dougszumski:
> @dougszumski What do we need to do to move this PR forwards? It seems relatively simple but has been sitting for a while and has conflicts

Someone needs to review it!

@darmach  please could you take a look?

darmach:
@dougszumski I tested that, looks fine!

Alex-Welsh (approved):

PR #1047, opened 2024-04-23
Update vault haproxy template

technowhizz:
This is to stop alerts that are firing at client sites. This will also alert us if a vault is sealed and/or needs bootstrapping. 

In the future we should look at using the built in exporters as [suggested by](https://stackhpc.slack.com/archives/C01HX4N1G3X/p1710325604898319?thread_ts=1710183886.551489&cid=C01HX4N1G3X) @dougszumski

### DIFF
--- a/doc/source/configuration/vault.rst
+++ b/doc/source/configuration/vault.rst
@@ -108,9 +108,10 @@ Setup HAProxy config for Vault
          option httpchk GET /v1/sys/health
          # https://www.vaultproject.io/api-docs/system/health
          # 200: initialized, unsealed, and active
+         # 429: backup
          # 501: not initialised (required for bootstrapping)
          # 503: sealed (required for bootstrapping)
-         http-check expect rstatus (200|501|503)
+         http-check expect rstatus (200|429)
 
       {% for host in groups['control'] %}
       {% set host_name = hostvars[host].ansible_facts.hostname %}


doc/source/configuration/vault.rst
```diff
@@ -108,9 +108,10 @@ Setup HAProxy config for Vault
          option httpchk GET /v1/sys/health
          # https://www.vaultproject.io/api-docs/system/health
          # 200: initialized, unsealed, and active
+         # 429: backup
          # 501: not initialised (required for bootstrapping)
          # 503: sealed (required for bootstrapping)
```
markgoddard, line 113: I'd like to see a successful multinode deployment with this because I remember explicitly needing 501/503 (as mentioned in these comments), and not wanting 429 to avoid hitting backup vaults.

markgoddard, line 113: These comments should be removed if we remove 501/503.

markgoddard (changes requested):

doc/source/configuration/vault.rst
technowhizz: I think @mnasiadka [suggested](https://stackhpc.slack.com/archives/C044YBYMJBZ/p1710240011742489?thread_ts=1710239842.598759&cid=C044YBYMJBZ) that the bootstrapping is done via the ip of the node as opposed to using the vip. But I will try and do this in a multinode and update this PR

markgoddard (commented):
@technowhizz I have rolled this change into #1090 and marked you as a co-author.

technowhizz:
change is in #1090

PR #1075, opened 2024-05-16
Add alerts for low available swap space

seunghun1ee:
As exhausting memory on hypervisor causes OOM kills, alerts for low available swap space are added.
Storage nodes usually utilise all of the swap spaces, so this alert can be spammed with them.
Therefore, silencing rules for storage nodes are going to be necessary.
We can implement per group alerting by overriding prometheus.yml.j2 from Kolla-ansible to avoid silencing approach but that needs separate discussion.

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/system.rules
+++ b/etc/kayobe/kolla/config/prometheus/system.rules
@@ -24,6 +24,24 @@ groups:
       summary: "Prometheus exporter at {{ $labels.instance }} reports low memory"
       description: "Available memory is {{ $value }} GiB."
 
+  - alert: LowSwapSpace
+    expr: (node_memory_SwapFree_bytes / node_memory_SwapTotal_bytes) < {% endraw %}{{ alertmanager_node_free_swap_warning_threshold_ratio }}{% raw %}
+    for: 1m
+    labels:
+      severity: warning
+    annotations:
+      summary: "Swap space at {{ $labels.instance }} reports low memory"
+      description: "Available swap space is {{ $value | humanizePercentage }}. Running out of swap space causes OOM Kills."
+
+  - alert: LowSwapSpace
+    expr: (node_memory_SwapFree_bytes / node_memory_SwapTotal_bytes) < {% endraw %}{{ alertmanager_node_free_swap_critical_threshold_ratio }}{% raw %}
+    for: 1m
+    labels:
+      severity: critical
+    annotations:
+      summary: "Swap space at {{ $labels.instance }} reports low memory"
+      description: "Available swap space is {{ $value | humanizePercentage }}. Running out of swap space causes OOM Kills."
+
   - alert: HostOomKillDetected
     expr: increase(node_vmstat_oom_kill[5m]) > 0
     for: 5m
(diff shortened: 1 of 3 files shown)


etc/kayobe/kolla/config/prometheus/system.rules
```diff
@@ -24,6 +24,24 @@ groups:
...
+      severity: warning
+    annotations:
+      summary: "Swap space at {{ $labels.instance }} reports low memory"
+      description: "Available swap space is {{ $value | humanizePercentage }}. Running out of swap space causes OOM Kills."
+
+  - alert: LowSwapSpace
+    expr: (node_memory_SwapFree_bytes / node_memory_SwapTotal_bytes) < {% endraw %}{{ alertmanager_node_free_swap_critical_threshold_ratio }}{% raw %}
+    for: 1m
+    labels:
+      severity: critical
```
dougszumski, line 40: This looks good - distinguishing by severity.

dougszumski (commented):
Thanks @seunghun1ee, good job. 

Please could you add a release note, and mention that we currently have a one-size fits all policy as an interim solution. We can advise that operators may need to tune the threshold for the alerts.

dougszumski:
Renders fine: 
image

etc/kayobe/kolla/config/prometheus/system.rules
seunghun1ee: Thank you

seunghun1ee:
> Thanks @seunghun1ee, good job.
> 
> Please could you add a release note, and mention that we currently have a one-size fits all policy as an interim solution. We can advise that operators may need to tune the threshold for the alerts.

Thank you @dougszumski 
I added release note.

dougszumski (approved):
Many thanks @seunghun1ee

etc/kayobe/kolla/config/prometheus/system.rules
```diff
@@ -24,6 +24,24 @@ groups:
       summary: "Prometheus exporter at {{ $labels.instance }} reports low memory"
       description: "Available memory is {{ $value }} GiB."
 
+  - alert: LowSwapSpace
+    expr: (node_memory_SwapFree_bytes / node_memory_SwapTotal_bytes) < {% endraw %}{{ alertmanager_node_free_swap_warning_threshold_ratio }}{% raw %}
```
markgoddard, line 28: What happens when there is no swap, i.e. node_memory_SwapTotal_bytes is 0?

seunghun1ee: That's good point. It seems like either adding conditional templating or updating ``expr`` for these are needed. @markgoddard  How should I do it? Do you want me to revert this merge and re-do the PR?

markgoddard: It depends on how long it will take for you to test it out and provide a fix. If it can be done quickly then no need to revert.

seunghun1ee: In that case I think we should revert this change. I'm working on customer's system this week.

dougszumski: The divide by zero evaluates to NaN, and the alert appears as 'OK'. Context: https://www.robustperception.io/get-thee-to-a-nannary/

dougszumski: Example from a no swap test env: 
image

dougszumski: So I think this alert is fine as it is, but a good question to ask none-the-less.

PR #896, opened 2024-01-26
Internal TLS certificate for os_capacity

assumptionsandg:
DNM: Wait for #697 to merge.

Set internal haproxy certificate in clouds.yaml

PR #1079, opened 2024-05-20
OS Capacity: Support providing a CA certificate

markgoddard:
For clouds that use an internal CA, it is necessary to provide a CA certificate to OS capacity.

### DIFF
--- a/doc/source/configuration/monitoring.rst
+++ b/doc/source/configuration/monitoring.rst
@@ -137,6 +137,8 @@ depending on your configuration, you may need set the
 ``kolla_enable_prometheus_ceph_mgr_exporter`` variable to ``true`` in order to
 enable the ceph mgr exporter.
 
+.. _os-capacity:
+
 OpenStack Capacity
 ==================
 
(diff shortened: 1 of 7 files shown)


markgoddard:
Adapted from https://github.com/stackhpc/stackhpc-kayobe-config/pull/896/

assumptionsandg (commented):
Has this been tested?

When testing #896 on SMS lab the container would always crash on SSL validation even when the cacert was provided and mounted in the container.

That could’ve just been me messing it up though.

markgoddard:
> Has this been tested?
> 
> When testing #896 on SMS lab the container would always crash on SSL validation even when the cacert was provided and mounted in the container.
> 
> That could’ve just been me messing it up though.

I've tested it in a multinode, and it fixed the container's restart loop.

assumptionsandg (approved):
> I've tested it in a multinode, and it fixed the container's restart loop.

Ah nice one, I didn't think it was just the restart loop causing problems.

LGTM, thanks for getting this one over the line :)

PR #1080, opened 2024-05-21
Fail kolla image build when critical CVEs are detected

seunghun1ee:
Added condition in ``scan-images.sh`` to add images to additional list ``critical-images.txt`` that can be used to check if there was any critical CVEs found during image scanning.
The container image build job will fail when critical CVEs are found from images but the push for other images will be performed.
Changed push dirty image option to false. (When it's set to true, images with critical CVEs will also get pushed and the job will not fail)

### DIFF
--- a/.github/workflows/stackhpc-container-image-build.yml
+++ b/.github/workflows/stackhpc-container-image-build.yml
@@ -42,8 +42,7 @@ on:
         description: Push scanned images that have vulnerabilities?
         type: boolean
         required: false
-        # NOTE(Alex-Welsh): This default should be flipped once we resolve existing failures
-        default: true
+        default: false
 
 env:
   ANSIBLE_FORCE_COLOR: True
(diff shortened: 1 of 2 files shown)


assumptionsandg (approved):
Do we still want to be merging features into Yoga? Otherwise LGTM

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -288,8 +294,15 @@ jobs:
         run: if [ $(wc -l < image-build-logs/push-failed-images.txt) -gt 0 ]; then cat image-build-logs/push-failed-images.txt && exit 1; fi
         if: ${{ !cancelled() }}
 
-      - name: Fail when images failed scanning
-        run: if [ $(wc -l < image-build-logs/dirty-images.txt) -gt 0 ]; then cat image-build-logs/dirty-images.txt && exit 1; fi
+      # NOTE(seunghun1ee): Currently we want to mark the job fail only when critical CVEs are detected.
```
Alex-Welsh, line 297: When we want to fail on high CVEs, we don't need to split up dirty.txt and critical.txt, they can be in the same file

markgoddard (approved):

PR #1082, opened 2024-05-22
Add RADOS gateway usage exporter support

seunghun1ee:


etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
```diff
@@ -0,0 +1,68 @@
+---
+- name: Deploy RADOS gateway usage exporter
+  hosts: monitoring
+  gather_facts: false
+  vars:
+    venv: "{{ virtualenv_path }}/openstack"
```
Alex-Welsh, line 7: Why does the venv assume the name `openstack`?

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/80-radosgw-exporter.yml
```diff
@@ -0,0 +1,21 @@
+# yamllint disable-file
+---
```
MoteHue, line 2: Does this work if ``stackhpc_enable_radosgw_usage_exporter = false``?
We normally see issues with empty yaml files, wondering if we need this:
https://github.com/stackhpc/stackhpc-kayobe-config/blob/afd30ccff36e55eb97fa536e6b319503fbc33f7e/etc/kayobe/ssh.yml#L19

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
seunghun1ee: Currently it needs to query ec2 credential with openstackclient. I'll try generalise this as it gets developed more. This is still work in progress.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/80-radosgw-exporter.yml
seunghun1ee: This was made based on https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/yoga/etc/kayobe/kolla/config/prometheus/prometheus.yml.d/70-oscapacity.yml

So, I assumed it's fine. Will have a look on that too.

seunghun1ee: @MoteHue From our CI tests, no error was occurred from ``stackhpc_enable_radosgw_usage_exporter = false``
I'd say it's safe.

Alex-Welsh:
@seunghun1ee what's the status of this PR? It's already a bit stale, can we get it moving again?

seunghun1ee:
@Alex-Welsh The playbook and Prometheus template were tested. This was set to WIP because the exporter itself is missing one criteria: Reducing the privileges of the credential that the exporter use.

I can close this and create new PR to Antelope branch if we are okay to move on with one missing criteria.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/80-radosgw-exporter.yml
Alex-Welsh: It works because it's not Ansible, it's just a prometheus config file using YAML

Alex-Welsh:
> @Alex-Welsh The playbook and Prometheus template were tested. This was set to WIP because the exporter itself is missing one criteria: Reducing the privileges of the credential that the exporter use.
> 
> I can close this and create new PR to Antelope branch if we are okay to move on with one missing criteria.

I see what you mean. Using admin creds isn't ideal but we do it in a lot of places. I don't think it should be a blocker in this instance.

Please do rebase this and change the target branch. Since it's a new feature, I think we should just put it in Caracal

Alex-Welsh:
Also, please add a release note

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
```diff
@@ -0,0 +1,152 @@
+---
+- name: Deploy RADOS gateway usage exporter
+  hosts: monitoring
+  gather_facts: false
+  tags: radosgw_usage_exporter
+  vars:
+    venv: "{{ virtualenv_path }}/openstack"
+  tasks:
```
Alex-Welsh, line 8: Could you add the `when: stackhpc_enable_radosgw_usage_exporter` to the play rather than each task individually? If not, then can you add the tasks to a block and add it there?

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
Alex-Welsh, line 1: It would be helpful if you could run ansible-lint against this file, because I'm adding it as CI in https://github.com/stackhpc/stackhpc-kayobe-config/pull/1368 so will conflict otherwise. I already have an ansible-lint environment set up so I can run it for you if that's easier

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
```diff
@@ -0,0 +1,152 @@
...
+          RADOSGW_SERVER: "{{ radosgw_server }}"
+          ADMIN_ENTRY: admin
+          ACCESS_KEY: "{{ ec2.Access }}"
+          SECRET_KEY: "{{ ec2.Secret }}"
+          VIRTUAL_PORT: "{{ stackhpc_radosgw_usage_exporter_port | string }}"
+        entrypoint: "{{ [ 'python', '-u', './radosgw_usage_exporter.py', '--insecure' ] if not stackhpc_radosgw_usage_exporter_verify else omit }}"
+      vars:
+        ec2: "{{ credential.stdout | from_json | first }}"
+        host: "{{ endpoint.stdout | from_json | first }}"
+        radosgw_server: "{{ host.URL | regex_replace('(https?://)([^:/]+):([0-9]+)/.*', '\\1\\2:\\3') }}" # Drop Swift api address
```
Alex-Welsh: Generally I'm not a fan of gnarly regex like this but I understand in some places it's necessary. Can you think of any alternatives here?

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
```diff
@@ -0,0 +1,152 @@
...
+
+    - name: Update CA certificate of RADOS gateway usage exporter container
+      community.docker.docker_container_exec:
+        container: radosgw_usage_exporter
+        command: update-ca-certificates
+        user: root
+      become: true
+      when:
+        - stackhpc_enable_radosgw_usage_exporter
+        - stackhpc_radosgw_usage_exporter_cacert | length > 0
```
Alex-Welsh: Again, add the `stackhpc_radosgw_usage_exporter_cacert | length > 0` condition to a block if you can

seunghun1ee:
@Alex-Welsh I found the way to use ceph_rgw user's ec2 credential instead of admin user.

etc/kayobe/stackhpc-monitoring.yml
```diff
@@ -53,3 +53,18 @@ redfish_exporter_default_password: "{{ ipmi_password }}"
...
+
+# Port to expose RADOS gateway usage exporter. Default is 9242
+stackhpc_radosgw_usage_exporter_port: 9242
+
+# Path to a certificate for internal TLS in the RADOS gateway usage exporter.
+stackhpc_radosgw_usage_exporter_cacert: ""
+
+# Whether TLS certificate verification is enabled for the RADOS gateway usage
+# exporter for querying Ceph RADOS gateway APIs. Default is true
+stackhpc_radosgw_usage_exporter_verify: true
```
Alex-Welsh: Seems strange that the default is true but there's no default cert or cert path.
Would it make sense for it to be dependent on something like `kolla_enable_tls_internal`?

Alex-Welsh (changes requested):
One general change, could you add it as a post-hook to service deploy since the default for `stackhpc_enable_radosgw_usage_exporter` is false?

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
seunghun1ee: I found we can use urlsplit ansible filter!

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
seunghun1ee: Done

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
```diff
@@ -0,0 +1,151 @@
...
+              ADMIN_ENTRY: admin
+              ACCESS_KEY: "{{ ec2.Access }}"
+              SECRET_KEY: "{{ ec2.Secret }}"
+              VIRTUAL_PORT: "{{ stackhpc_radosgw_usage_exporter_port | string }}"
+            entrypoint: "{{ ['python', '-u', './radosgw_usage_exporter.py', '--insecure'] if not stackhpc_radosgw_usage_exporter_verify else omit }}"
+          vars:
+            ec2: "{{ credential.stdout | from_json | first }}"
+          become: true
+
+        - name: Ensure the exporter to use certificate
```
Alex-Welsh: nit: the grammar could be improved here but it's not very important

releasenotes/notes/add-radosgw-usage-exporter-support-93d55c544418b05a.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Adds RADOS Gateway usage exporter support.
+
+    To deploy the exporter, set the variable ``stackhpc_enable_radosgw_usage_exporter``
+    to true. Then run playbook ``deploy-radosgw-usage-exporter.yml``.
```
Alex-Welsh, line 7: Nit: also need to set cert path if internal tls is enabled

Alex-Welsh (commented):
Couple of nits but almost there

etc/kayobe/ansible/deploy-radosgw-usage-exporter.yml
Alex-Welsh: Maybe this:
```suggestion
        - name: Ensure the exporter uses a custom certificate
```

Alex-Welsh (approved):

PR #1085, opened 2024-06-03
Revert "Add alerts for low available swap space"

markgoddard:
Reverts stackhpc/stackhpc-kayobe-config#1075

There is an unresolved question about the behaviour of the alert rule when the denominator is zero.


seunghun1ee (approved):
Thanks for finding that issue!
I'll open PR again once I address this issue.

dougszumski:
The original alert works as intended (comments on the original PR), so no need for this revert.

PR #1086, opened 2024-06-03
yoga: Use Rocky Linux 9 as a base for kayobe-automation

markgoddard:
The default CentOS Stream 8 base image is no longer available due to EOL.


jackhodgkiss (approved):
LGTM. Rocky 9 has been used the basis of kayobe automation successfully without issue. I will open a PR within to change the default from CentOS Stream 8.

markgoddard:
RL9 change looks good, but CS8 jobs are broken by removal of upstream repos.

PR #1087, opened 2024-06-04
Add custom playbooks to manage Dell firmware

priteau:


markgoddard:
IMHO it's time to stop pushing features into yoga. Also wondering if we need a stackhpc.dell collection for this kind of thing?

priteau:
I rebased to target Antelope.

doc/source/operations/dell-firmware-update.rst
```diff
@@ -0,0 +1,161 @@
...
+
+   # share the new version
+   [root@seed /]# drm --deployment-type=share --location=/dell_firmware -r=idrac_repo:1.01
+   [root@seed /]# ls -ltra /dell_firmware | tail -1
+   -rw-r--r--   1 root root 7103842 Jan  9 14:24 idrac_repo_Catalog.xml
+
+Then update the ``dell_drm_repo`` variable in ``drac-firmware-update.yml`` if
+required.
+
+Manually adding and update file
```
MaxBed4d: Is this section an alternative to the ``Updating the Repo`` section, or is this part required after it? 
If it is an alternative to the previous section, it may be worth mentioning it in the title as an "(alternative method)", if it is not an alternative it may be worth specifying in the title what files they are specifically updating to help differentiate between them.

MaxBed4d (commented):
Looks good to me, and the way it has been documented makes it relatively easy to grasp. I would like to test this out to make sure it does work as intended, unless it already has. 

My only real suggestion would be to add a short set of numbered steps somewhere in the docs to outline which steps should be done in which order to prevent some of the ''note:'' warnings e.g. ''The playbook will likely fail with an error if the iDRAC firmware is being updated''.

etc/kayobe/ansible/maintenance/drac-firmware-update.yml
```diff
@@ -0,0 +1,64 @@
...
+  connection: local
+  serial: "{{ lookup('env', 'ANSIBLE_SERIAL') | default(1, true) }}"
+
+  collections:
+   - dellemc.openmanage
+
+  vars:
+    # Run with ``-e dell_drm_apply_update=true`` to apply the firmware updates.
+    dell_drm_apply_update: False
+    dell_drm_address: "https://{{ oob_oc_net_name | net_ip(inventory_hostname=groups['seed'][0]) }}"
```
Alex-Welsh, line 13: The seed is assumed here when previously the criteria was "Choose a host where it won't conflict with another service". I think you should either add a dell_drm group or make it clear that the seed is the default host.

doc/source/operations/dell-firmware-update.rst
```diff
@@ -0,0 +1,161 @@
...
+DRM needs to listen on port 443 and needs access to the out-of-band management
+network. Choose a host where it won't conflict with another service.
+
+To run DRM in a container, first start a container, that has a Docker volume to
+host all the firmware files:
+
+.. code-block:: bash
+
+   docker volume create dell_firmware
+   docker run --detach -v dell_firmware:/dell_firmware --name dell-drm --network host --restart always rockylinux:9.3 sleep infinity
```
Alex-Welsh: There's a lot of manual steps here just to set up the containers properly. Would it be possible to just move everything to a script?

Alex-Welsh (commented):
Also, needs a release note

Alex-Welsh:
@priteau what's the status of this PR? I see you've updated it somewhat recently. Is it ready for review?

PR #1091, opened 2024-06-06
Update Magnum CAPI Helm driver to version 1.0.0

sd109:


mnasiadka:
Should we have a reno here?

markgoddard:
and some new images :)

sd109:
I actually remembered to write a release note for once... and then forgot to add it to the commit 🤦 Will get the images built and tags updated.

markgoddard (approved):

PR #1093, opened 2024-06-07
Build neutron-mlnx-agent kolla container

tomclark0:
Enabling building of neutron-mlnx-agent kolla container for Infiniband VF support.
This container is also pulled for neutron-eswitchd, another requirement for Infiniband VF support.

### DIFF
--- a/etc/kayobe/pulp.yml
+++ b/etc/kayobe/pulp.yml
@@ -477,6 +477,7 @@ stackhpc_pulp_images_kolla:
   - neutron-dhcp-agent
   - neutron-l3-agent
   - neutron-metadata-agent
+  - neutron-mlnx-agent
   - neutron-openvswitch-agent
   - neutron-server
   - neutron-sriov-agent


etc/kayobe/pulp.yml
```diff
@@ -477,6 +477,7 @@ stackhpc_pulp_images_kolla:
   - neutron-dhcp-agent
   - neutron-l3-agent
   - neutron-metadata-agent
+  - neutron-mlnx-agent
```
JohnGarbutt, line 480: I was expecting the eswitchd image in here?

Do we need to install networking-mlnx into the nova-compute image as well, so os-vif can use the ebrctl?
https://opendev.org/x/networking-mlnx/src/commit/0a69b9715505f21fe7cca8c408c42b618d2e72d7/setup.cfg#L55

JohnGarbutt: ah, sorry, there isn't one, even better!

JohnGarbutt:
I think we should add some infiniband docs into here:
https://github.com/stackhpc/stackhpc-kayobe-config/tree/stackhpc/2023.1/doc/source/configuration

etc/kayobe/pulp.yml
markgoddard, line 480: This is a list of images that get synced into pulp, but doesn't control which images are built in our CI pipeline. For that you would need to set `kolla_enable_neutron_mlnx: true` in `etc/kayobe/environments/ci-builder/stackhpc-ci.yml`.

For context, this adds the regex to the kolla-build command https://opendev.org/openstack/kayobe/src/branch/master/ansible/inventory/group_vars/all/kolla#L206

markgoddard (changes requested):

tomclark0:
Closing in favor of https://github.com/stackhpc/stackhpc-kayobe-config/pull/1211

PR #1097, opened 2024-06-12
Use secure CirrOS image in tempest by default

Alex-Welsh:


Alex-Welsh:
Looking at docker-rally, we are actually using the passwordless image by default. Does it help to add it explicitly here for visibility, or should we just remove the redundant override?

.automation.conf/tempest/tempest-ci-multinode.overrides.conf
```diff
@@ -1,6 +1,9 @@
 [DEFAULT]
 openstack_client_http_timeout = 300
 
+[openstack]
+img_url=https://github.com/stackhpc/cirros/releases/download/20231206/cirros-d231206-x86_64-disk.img
```
jovial, line 5: I think this is a rally config option rather than a tempest one

Alex-Welsh:
Setting is already configured by default

PR #1099, opened 2024-06-14
Add condition to ensure swap is not zero for swap alert

seunghun1ee:
I thought I introduced divide by zero situation even though Prometheus can handle zero denominator with NaN. Let's add condition explicitly checking if swap space is not zero for humans.


dougszumski (approved):
Thanks @seunghun1ee

seunghun1ee:
@dougszumski Thanks for approval. The checks for Centos8 is currently broken. Could you force merge this branch?

PR #1102, opened 2024-06-18
Yoga Infiniband

tomclark0:
Bringing up networking-mlnx in yoga to support Infiniband based networking via neutron with nova support.

Yoga is used as this is the release available in the environment for testing and validation.
There are currently no builds of networking-mlnx since the zed release cycle.

This patch enables building of the **neutron-mlnx-agent** container.
This container also provides the neutron-eswitchd container.
The networking-mlnx package is required in neutron_server and added as a plugin.
The networking-mlnx package is required in nova-compute and added as a plugin.

The networking-mlnx plugin has additional dependencies on system-level packages python3-libvirt python3-ethtool.
These are added to neutron_server and nova_compute. 

networking-mlnx is required in nova-compute as it provides the ebrctl binary used for managing Infiniband VIFs - e.g. https://github.com/openstack/nova/blob/unmaintained/yoga/nova/privsep/libvirt.py#L115-L118


tomclark0:
Working here in favour of https://github.com/stackhpc/stackhpc-kayobe-config/pull/1093 as the current environment available for Infiniband is Yoga based.

etc/kayobe/kolla.yml
```diff
@@ -424,6 +444,18 @@ kolla_build_blocks:
     {% endif %}
     {% endraw %}
     {% endif %}
+  nova_compute_footer: |
+    {% if kolla_base_distro in ['centos', 'rocky'] %}
+    # Install networking-mlnx plugin and dependencies
+    {% raw %}
+    {% if base_package_type == 'rpm' %}
+    ADD plugins-archive /
```
jovial: Package installs look reasonable. I think:

```
  neutron_base_packages_append
    - python3-libvirt 
    - python3-ethtool
  nova_base_packages_append:
    - python3-libvirt
    - python3-ethtool
```

and
```
  neutron_pip_packages_append:
    - /plugins/*
  nova_pip_packages_append:
    - /plugins/*
```

Would be more normal.  Was there a particular reason not to do it this way? Also need something similar for ubuntu?

EDIT: Realised we don't need _pip_packages_append:

jovial: Thinking something more like this: https://github.com/stackhpc/stackhpc-kayobe-config/commit/a22bf2deaa1d7cebfbabc3655fd5973bc88f1edb

etc/kayobe/kolla.yml
```diff
@@ -448,6 +456,12 @@ kolla_build_customizations_common:
     - /additions/*
   ironic_inspector_pip_packages_append:
     - /additions/*
+  neutron_server_packages_append:
+    - python3-libvirt
+    - python3-ethtool
+  nova_compute_packages_append:
```
jovial, line 462: Annoyingly, this gets overridden by: [nova_compute_packages_append:](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1102/files#diff-12d955d16a7bd8d6dc7bae8b56c2e485f4445434f0fb67546ea09c9e8b9b504dR483)

Due to https://github.com/stackhpc/stackhpc-kayobe-config/pull/1102/files#diff-12d955d16a7bd8d6dc7bae8b56c2e485f4445434f0fb67546ea09c9e8b9b504dR516

I think we need to append them to:

```
  nova_compute_packages_remove:
    - openvswitch
```

in kolla_build_customizations_el as well.

jovial: The other thing we need is a release note

Alex-Welsh:
@tomclark0 What's the status of this PR? we'll be dropping support for Yoga fairly soon so it'd be good to get this finished off

Alex-Welsh:
@tomclark0 should we close this PR?

PR #1073, opened 2024-05-15
Fix rabbitmq-reset.yml trying to restart wrong services

darmach:
Fixes egrep passed to systemctl attempting to restart other services containing nova in the name:

```
  stderr: |-
    Failed to restart dev-disk-byx2did-dmx2dnamex2drootvgx2dlv_nova.device: Job type restart is not applicable for unit dev-disk-byx2did-dmx2dnamex2drootvgx2dlv_nova.device.
    Failed to restart dev-mapper-rootvgx2dlv_nova.device: Job type restart is not applicable for unit dev-mapper-rootvgx2dlv_nova.device.
    Failed to restart dev-rootvg-lv_nova.device: Job type restart is not applicable for unit dev-rootvg-lv_nova.device.
```

PR #1103, opened 2024-06-19
Improve rabbitmq reset playbook matching

Alex-Welsh:
Two issues were picked up during the Sussex deployment. The original command matched against the correct lines but also these:
```
ceph-42b27b44-271a-11ef-8338-58a2e1dff6f2@mds.manila-cephfs.artemis-cp-01.rygqwc.service
●
●
●
```
The first issue is that it was matching against ceph containers which happen to have ```manila``` in the name
The second was that it was matching against failed units, which show up in the list like this:
```
kolla-haproxy-container.service      loaded    active   running   docker kolla-haproxy-container.service
● kolla-heat_api-container.service      loaded    failed   failed    docker kolla-heat_api-container.service
```
So it's matching against the dot
Stopped containers don't need to be stopped and restarted so we should only match against running units

### DIFF
--- a/etc/kayobe/ansible/rabbitmq-reset.yml
+++ b/etc/kayobe/ansible/rabbitmq-reset.yml
@@ -68,4 +68,4 @@
     # The following services use RabbitMQ.
     - name: Restart OpenStack services
       shell: >-
-        systemctl -a | egrep '(barbican|blazar|cinder|cloudkitty|designate|heat|ironic|keystone|magnum|manila|neutron|nova|octavia)' | awk '{ print $1 }' | xargs systemctl restart
+        systemctl list-units --type=service | tr ' ' '\n' | egrep '^kolla-(barbican|blazar|cinder|cloudkitty|designate|heat|ironic|keystone|magnum|manila|neutron|nova|octavia).*-container\.service' | uniq | xargs systemctl restart


MoteHue:
> Stopped containers don't need to be stopped and restarted so we should only match against running units

Is this always true? I reckon a container could get into a failed state due to rabbit issues, it might be better to match against the description as this is the last field and always matches the kolla container name.

Alex-Welsh:
> > Stopped containers don't need to be stopped and restarted so we should only match against running units
> 
> Is this always true? I reckon a container could get into a failed state due to rabbit issues, it might be better to match against the description as this is the last field and always matches the kolla container name.

Feels a bit wrong to match against the description rather than the ID but fair point

MoteHue (dismissed):

etc/kayobe/ansible/rabbitmq-reset.yml
```diff
@@ -68,4 +68,4 @@
     # The following services use RabbitMQ.
     - name: Restart OpenStack services
       shell: >-
-        systemctl -a | egrep '(barbican|blazar|cinder|cloudkitty|designate|heat|ironic|keystone|magnum|manila|neutron|nova|octavia)' | awk '{ print $1 }' | xargs systemctl restart
+        systemctl list-units --type=service | egrep '(barbican|blazar|cinder|cloudkitty|designate|heat|ironic|keystone|magnum|manila|neutron|nova|octavia)' | egrep -v ceph | awk '{ print $NF }' | xargs systemctl restart
```
priteau: What about using a grep for `^kolla-` instead of filtering out `ceph`?

priteau: Are you sure about `$NF` in your awk here?

priteau (changes requested):

etc/kayobe/ansible/rabbitmq-reset.yml
priteau: I see this is to match the description. A bit dangerous if this changes in kolla-ansible.

priteau: Something like `egrep 'kolla-(barbican|…)'`

Alex-Welsh: Done

Alex-Welsh: New approach, splitting every word out to a new line and selecting unique entries with better regex matching (must start with `kolla-`, end with `-container.service`

MoteHue: Will this now pick up each container twice, from the name and description? Could there be any issues from calling the restart twice in quick succession?

MoteHue: `systemctl list-units` supports passing a pattern in, can we do that instead of the grep? https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#list-units%20PATTERN%E2%80%A6

Alex-Welsh: I added `| uniq` to stop it duplicating lines.

I don't think the systemctl pattern really helps. Grep is more versatile here

MoteHue (approved):

darmach:
@Alex-Welsh  It's always good to check if a PR attempting to fix your issue is not already opened :)

https://github.com/stackhpc/stackhpc-kayobe-config/pull/1073

darmach:
Oh, I see that you've seen that one (as you've upvoted Michal's suggestion) :)
Why open the new pull request then @Alex-Welsh ?

Alex-Welsh:
> Oh, I see that you've seen that one (as you've upvoted Michal's suggestion) :) Why open the new pull request then @Alex-Welsh ?

I can barely remember what I had for breakfast this morning, let alone anything that happened in May :laughing:. I'm happy to close this one and continue with yours if you update it

darmach:
@Alex-Welsh actually, it's my bad! I fixed matching aswell and never pushed to that branch I pasted earler! That's why I found this redundant. Great job and thanks!

MoteHue:
https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2023.1/etc/kayobe/ansible/stop-openstack-services.yml Just realised we might want to change this too, as it will also be matching against the dot on failed services. Less impactful as stopping a failed service probably won't do much, but still seems worth doing.

Alex-Welsh:
Closing in favor of https://github.com/stackhpc/stackhpc-kayobe-config/pull/1073

PR #1106, opened 2024-06-21
Disables password expiration and inactivity policies

jovial:
This was causing the kayobe and kolla service accounts to be locked out of the system.

### DIFF
--- a/etc/kayobe/inventory/group_vars/overcloud/cis
+++ b/etc/kayobe/inventory/group_vars/overcloud/cis
@@ -75,6 +75,10 @@ rhel9cis_max_log_file_size: 1024
 # `rhel9cis_bootloader_password_hash`
 rhel9cis_set_boot_pass: false
 
+# NOTICE: rule disabled otherwise rule will prevent access to accounts
+# as it will expire passwords older than one year.
+rhel9cis_rule_5_6_1_1: false
+
 ##############################################################################
 # Ubuntu Jammy CIS Hardening Configuration
 
@@ -159,4 +163,20 @@ ubtu22cis_max_log_file_size: 1024
 # ubtu22cis_bootloader_password_hash
 ubtu22cis_rule_1_4_1: false
 ubtu22cis_rule_1_4_3: false
+
+# Disable: Ensure minimum days between password changes is configured
+ubtu22cis_rule_5_5_1_1: false
+
+# Disable: Ensure password expiration is 365 days or less
+ubtu22cis_rule_5_5_1_2: false
+
+# Disable: Ensure password expiration warning days is 7 or more
+ubtu22cis_rule_5_5_1_3: false
+
+# Disable: Ensure inactive password lock is 30 days or less
+ubtu22cis_rule_5_5_1_4: false
+
+# Disable: Ensure all users last password change date is in the past
+ubtu22cis_rule_5_5_1_5: false
+
 ##############################################################################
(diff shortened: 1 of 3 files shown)


jovial:
I've not tested this yet. It seems common to use 99999 but we may be able to use -1 instead. See: https://linux.die.net/man/1/chage

MoteHue:
> I've not tested this yet. It seems common to use 99999 but we may be able to use -1 instead. See: https://linux.die.net/man/1/chage

FWIW I found running the chage command with -1 didn't make any change, I figured that was why it's common to use 99999

etc/kayobe/inventory/group_vars/overcloud/cis
```diff
@@ -75,6 +75,12 @@ rhel9cis_max_log_file_size: 1024
 # `rhel9cis_bootloader_password_hash`
 rhel9cis_set_boot_pass: false
 
+# Disable account locking after inactivity period
+rhel9cis_inactivelock:
```
markgoddard: @jackhodgkiss did it this way for a client deployment. Is there any difference?
```
---
# NOTICE: rule disabled otherwise rule will prevent access to accounts
# as it will expire passwords older than one year.
rhel9cis_rule_5_6_1_1: false
```

jovial: I think that should work for a new deployment, but I wanted to restore the defaults on deployments where this is already applied. Which would you prefer?

markgoddard: Makes sense to try to fix existing deployments if we can.

jovial:
At least on Ubuntu, you'd would of had to run the hardening before host configure, as by default, it doesn't change existing users without [ubtu22cis_disruption_high](https://github.com/ansible-lockdown/UBUNTU22-CIS/blob/1c8c40feac5b2d8ea9c13e94ad34cb41ed054c9b/tasks/section_5/cis_5.5.x.yml#L90):
```
ubuntu@will-controller-01:~$ sudo chage -l stack
Last password change                                    : Jul 08, 2024
Password expires                                        : never
Password inactive                                       : never
Account expires                                         : never
Minimum number of days between password change          : 0
Maximum number of days between password change          : 99999
Number of days of warning before password expires       : 7
ubuntu@will-controller-01:~$ sudo chage -l kolla
Last password change                                    : Jul 08, 2024
Password expires                                        : never
Password inactive                                       : never
Account expires                                         : never
Minimum number of days between password change          : 0
Maximum number of days between password change          : 99999
Number of days of warning before password expires       : 7
```
I've also found the logic in the shell command checking which entries to update will never increase the expiry:
"Ensure password expiration is 365 days or less". So may need some tasks to do this.

markgoddard (dismissed):

etc/kayobe/inventory/group_vars/overcloud/cis
```diff
@@ -159,4 +163,20 @@ ubtu22cis_max_log_file_size: 1024
...
 ubtu22cis_rule_1_4_1: false
 ubtu22cis_rule_1_4_3: false
+
+# Disable: Ensure minimum days between password changes is configured
+ubtu22cis_rule_5_5_1_1: false
+
+# Disable: Ensure password expiration is 365 days or less
+ubtu22cis_rule_5_5_1_2: false
+
+# Disable: Ensure password expiration warning days is 7 or more
```
technowhizz: Do we need to disable the password expiration warning is 7 days or more? I think it should be enabled if anything?

jovial: Disabling the rule just means that we won't make any changes to the current policy, but you are right, this one seems relatively benign. This would have the effect of changing the default to 7 days for new users (which is also the distro default, so wouldn't have an effect on new systems).

jovial: done - thanks

markgoddard (approved):

PR #1112, opened 2024-06-25
Fix build and rotation of amphora images on Rocky 9

jovial:
EPEL is required for the debootstrap package.

### DIFF
--- a/etc/kayobe/ansible/octavia-amphora-image-build.yml
+++ b/etc/kayobe/ansible/octavia-amphora-image-build.yml
@@ -8,7 +8,7 @@
       package:
         name: epel-release
       become: true
-      when: 
+      when:
         - ansible_facts.os_family == "RedHat"
         - dnf_custom_repos | falsy
 
--- a/etc/kayobe/ansible/octavia-amphora-image-build.yml
+++ b/etc/kayobe/ansible/octavia-amphora-image-build.yml
@@ -8,7 +8,9 @@
       package:
         name: epel-release
       become: true
-      when: ansible_facts.os_family == "RedHat"
+      when: 
+        - ansible_facts.os_family == "RedHat"
+        - dnf_custom_repos | falsy
 
     - name: Ensure packages are installed
       become: true
--- a/etc/kayobe/ansible/octavia-amphora-image-build.yml
+++ b/etc/kayobe/ansible/octavia-amphora-image-build.yml
@@ -4,6 +4,12 @@
   vars:
     amphora_dib_upper_constraints_file: "{{ pip_upper_constraints_file }}"
   tasks:
+    - name: Install EPEL
+      package:
+        name: epel-release
+      become: true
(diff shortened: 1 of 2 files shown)


jovial:
This seems to be working for me now

Alex-Welsh (dismissed):

etc/kayobe/ansible/octavia-amphora-image-build.yml
```diff
@@ -4,6 +4,12 @@
   vars:
     amphora_dib_upper_constraints_file: "{{ pip_upper_constraints_file }}"
   tasks:
+    - name: Install EPEL
+      package:
+        name: epel-release
```
markgoddard, line 9: We generally try to avoid enabling EPEL. But if we need it, we might want to consider temporarily enabling it. Do we need to consider using the release train EPEL repo here?

etc/kayobe/ansible/octavia-amphora-image-register.yml
```diff
@@ -39,7 +40,14 @@
         checksum_algorithm: md5
       changed_when: false
       register: image_checksum
-      when: image_info.openstack_image
+      when: image_info.image
+
+    - name: Assert that Amphora image exists
+      assert:
+        that: image_checksum.stat.exists
```
markgoddard: The image_checksum variable is registered by a conditional task.

etc/kayobe/ansible/octavia-amphora-image-build.yml
jovial: So this is for debootstrap on Rocky. At least in the environment I was testing, the seed wasn't configured to use release train snapshots, but fair enough, the builder group is customisable. We could also consider building Rocky based images on a Rocky Host OS?

etc/kayobe/ansible/octavia-amphora-image-register.yml
jovial: Good spot - thanks. I've removed the conditional and moved it to the top of the file so that we fail early.

etc/kayobe/ansible/octavia-amphora-image-build.yml
jovial: https://github.com/stackhpc/stackhpc-kayobe-config/blob/699769ce3756c6dc80c1183e1c87f730ff05bb80/etc/kayobe/inventory/group_vars/overcloud/stackhpc-repos#L5

markgoddard: That's true - in general it doesn't use release train repos.

markgoddard (dismissed):

markgoddard:
pep8 failed

jovial:
> pep8 failed

Darn - shouldn't have edited in the github UI :laughing: - hopefully fixed now

priteau (approved):

PR #1116, opened 2024-06-26
Refer to internal docs for building amphora images

jovial:
Easy to miss that these exist without a direct link. The original [PR](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1100) targeted zed so I've updated that branch, but we should be careful not to forget to port this forward to the current docs.

### DIFF
--- a/doc/source/operations/octavia.rst
+++ b/doc/source/operations/octavia.rst
@@ -2,6 +2,8 @@
 Octavia
 =======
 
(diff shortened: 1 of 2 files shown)


MoteHue (approved):
I'd also missed these docs exist, LGTM.

markgoddard (approved):

PR #1104, opened 2024-06-21
Enable CIS benchmark hardening in AIO and Multinode

jovial:
Currently we do not have coverage of this feature in CI.

PR #1119, opened 2024-06-27
Stop changing permissions on files

jovial:
These are causing changes to docker overlay filesystems with possible unintended consequences. It is also really slow to loop through so many files in ansible.

### DIFF
--- a/etc/kayobe/inventory/group_vars/overcloud/cis
+++ b/etc/kayobe/inventory/group_vars/overcloud/cis
@@ -115,9 +115,22 @@ ubtu22cis_sshd:
   deny_users: ""
   deny_groups: ""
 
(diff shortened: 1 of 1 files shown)


markgoddard (approved):

markgoddard (commented):
Do we not need to do the same for RHEL?

markgoddard (commented):
Should these group_vars be moved to the cis-hardening group?

jovial:
> Should these group_vars be moved to the cis-hardening group?

Will do that in https://github.com/stackhpc/stackhpc-kayobe-config/pull/1104

jovial:
> Do we not need to do the same for RHEL?

I believe I checked and couldn't see it doing the same kind of permission changes

PR #1121, opened 2024-06-28
Do not disable IPV6 by default

jovial:
We are currently prevented in doing so by:

https://bugs.launchpad.net/kolla-ansible/+bug/2071443

You only hit this issue after a reboot.

### DIFF
--- a/etc/kayobe/inventory/group_vars/overcloud/cis
+++ b/etc/kayobe/inventory/group_vars/overcloud/cis
@@ -133,4 +133,10 @@ ubtu22cis_max_log_file_size: 1024
 # ubtu22cis_bootloader_password_hash
 ubtu22cis_rule_1_4_1: false
 ubtu22cis_rule_1_4_3: false
+
+# The way this is disabled currently breaks kolla's IPV6 check, see:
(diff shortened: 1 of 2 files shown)


jovial:
Alternatively we could use ``ubtu22cis_ipv6_disable == 'sysctl'`` or fix the kolla bug and backport.

etc/kayobe/inventory/group_vars/overcloud/cis
```diff
@@ -133,4 +133,10 @@ ubtu22cis_max_log_file_size: 1024
 # ubtu22cis_bootloader_password_hash
 ubtu22cis_rule_1_4_1: false
 ubtu22cis_rule_1_4_3: false
+
+# The way this is disabled currently breaks kolla's IPV6 check, see:
+# https://bugs.launchpad.net/kolla-ansible/+bug/2071443
+# Also matches RHEL hardening behavior.
+ubtu22cis_ipv6_required: true
```
jovial, line 140: Also matches what we do on Rocky

markgoddard (approved):

PR #1131, opened 2024-07-03
CI: Increase aio volume size to 50GB in upgrade jobs

markgoddard:
New images seem to be running out of space.

(cherry picked from commit 006931eface514f15f9aa7e6f1c27de45f5068a5)


Alex-Welsh (approved):
lgtm as long as CI passes

PR #1137, opened 2024-07-08
Yoga CVE-2024-6387

MoteHue:
Only need to bump Jammy, Focal LTS is not vulnerable.
No new containers needed, we don't build Jammy containers in Yoga.


Alex-Welsh:
Got some conflicts there, and CI isn't running :thinking:

PR #1140, opened 2024-07-08
Fix fluentd getting 413 errors from OpenSearch

priteau:
This is a temporary fix until a solution is pushed upstream.

### DIFF
--- a/etc/kayobe/kolla/config/fluentd/output/00-local.conf
+++ b/etc/kayobe/kolla/config/fluentd/output/00-local.conf
@@ -1,0 +1,85 @@
+{% raw %}
+{% for item in syslog_facilities | selectattr('enabled') %}
+<match syslog.{{ item.facility }}.**>
+  @type copy
+  <store>
+    @type file
+    path /var/log/kolla/{{ item.logdir }}/{{ item.logfile }}
+    append true
+    # Disable timestamp in filename for logs
+    <buffer []>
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla/config/fluentd/output/00-local.conf
```diff
@@ -0,0 +1,85 @@
...
+{% if fluentd_opensearch_user != '' and fluentd_opensearch_password != ''%}
+       user {{ fluentd_opensearch_user }}
+       password {{ fluentd_opensearch_password }}
+{% endif %}
+       logstash_format true
+       logstash_prefix {{ opensearch_log_index_prefix }}
+       reconnect_on_error true
+       request_timeout {{ fluentd_opensearch_request_timeout }}
+       suppress_type_name true
+       bulk_message_request_threshold 20M
```
markgoddard, line 74: It looks like this and L79 are the new ones.

markgoddard (approved):

etc/kayobe/kolla/config/fluentd/output/00-local.conf
priteau: Would you like a comment to make them stand out?

markgoddard: Might have been nice, but probably not worth a respin.

PR #1144, opened 2024-07-09
Use package repos in AIO CI

MoteHue:
Probably need to wait for the VM to become active after reboot

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -315,6 +315,26 @@ jobs:
         env:
           KAYOBE_AUTOMATION_SSH_PRIVATE_KEY: ${{ steps.ssh_key.outputs.ssh_key }}
 
+      - name: Update host packages
+        run: |
+          docker run -t --rm \
+            -v $(pwd):/stack/kayobe-automation-env/src/kayobe-config \
+            -e KAYOBE_ENVIRONMENT -e KAYOBE_VAULT_PASSWORD -e KAYOBE_AUTOMATION_SSH_PRIVATE_KEY \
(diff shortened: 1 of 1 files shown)


.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -284,6 +284,26 @@ jobs:
...
+        env:
+          KAYOBE_AUTOMATION_SSH_PRIVATE_KEY: ${{ steps.ssh_key.outputs.ssh_key }}
+
+      - name: Reboot VM
+        run: |
+          docker run -t --rm \
+            -v $(pwd):/stack/kayobe-automation-env/src/kayobe-config \
+            -e KAYOBE_ENVIRONMENT -e KAYOBE_VAULT_PASSWORD -e KAYOBE_AUTOMATION_SSH_PRIVATE_KEY \
+            $KAYOBE_IMAGE \
+            /stack/kayobe-automation-env/src/kayobe-config/.automation/pipeline/playbook-run.sh etc/kayobe/ansible/reboot.yml
```
markgoddard, line 334: This playbook should wait for an SSH connection.

MoteHue:
Looks like Rocky is also having issues with recovering from reboots :(

Alex-Welsh:
Well... some of it works

Upgrades both fail which is kinda expected because it would have to work on Zed as well. I'd be happy to just skip the tasks on upgrade jobs to ignore that (as long as we properly test it in Caracal).

We should definitely fix it for Rocky though. Just need to find a free afternoon to dig into it.

Anyone is welcome to pick this up. Just boot up a Rocky AIO, set a root password, reboot it, and use the horizon console to work out why it's inaccessible.

PR #1150, opened 2024-07-10
Adds a workflow to create an AIO

jovial:
This creates an AIO with the usual workflow but doesn't delete it. The idea is you can login with a secondary key and debug why the failure is occurring. Just remember to manually delete the VM afterwards.


.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -1,328 +1,36 @@
 ---
```
markgoddard, line 1: What's going on here?

terraform/aio/templates/userdata.cfg.tpl
```diff
@@ -10,6 +10,9 @@ runcmd:
 # This means only the instance needs to be cleaned up if the destroy fails.
 ssh_authorized_keys:
   - ${ssh_public_key}
+%{ if debug_ssh_key != "" }
+  - ${debug_ssh_key}
+%{ endif }
```
markgoddard, line 15: I like this approach. We should do something similar for multinodes.

.github/workflows/stackhpc-container-image-build.yml
jovial: Oh, it so I could test it before merging. I needed the action to already exist in https://github.com/stackhpc/stackhpc-kayobe-config/actions
to be able to trigger it before merging . The idea was to move to a new file once it worked. Not sure if there is an easier way?

markgoddard: perhaps add a workflow_dispatch trigger to the PR workflow?

Alex-Welsh:
This looks handy, @jovial is there any chance of it being resurrected?

Alex-Welsh:
No updates in a few years, closing

PR #1157, opened 2024-07-12
Split Ubuntu Docker CE package into Focal/Jammy

Alex-Welsh:
Hitting the same issues as we saw in release train a while ago, docker CE fails to sync because there are conflicting packages which look the same but have different checksums. The solution is to split them out into foca/jammy

See https://github.com/stackhpc/stackhpc-release-train/pull/279


etc/kayobe/apt.yml
```diff
@@ -61,8 +61,13 @@ stackhpc_apt_repositories:
     components: upload
     architecture: amd64
     trusted: yes
-  - url: "{{ stackhpc_repo_docker_ce_ubuntu_url }}"
-    suites: "{{ ansible_facts.distribution_release }}"
+  - url: "{{ stackhpc_repo_docker_ce_ubuntu_focal_url }}"
```
markgoddard: Should this use the same URL as the first 2 repos (L51)? Similar for suites. We don't want both repos enabled.

markgoddard (changes requested):

etc/kayobe/apt.yml
Alex-Welsh: Done

markgoddard (dismissed):

markgoddard:
CI failed:
```
RUNNING HANDLER [apt : Update apt cache] ***************************************
Friday 12 July 2024  11:48:18 +0000 (0:00:00.614)       0:00:11.507 *********** 
fatal: [controller0]: FAILED! => changed=false 
  msg: 'Failed to update apt cache: W:Updating from such a repository can''t be done securely, and is therefore disabled by default., W:See apt-secure(8) manpage for repository creation and user configuration details., E:The repository ''https://ark.stackhpc.com/pulp/content/docker-ce/ubuntu-focal/20231020T014922 focal Release'' does not have a Release file.'
```

etc/kayobe/pulp-repo-versions.yml
```diff
@@ -16,7 +16,8 @@ stackhpc_pulp_repo_centos_stream_9_nfv_openvswitch_version: 20230929T005202
 stackhpc_pulp_repo_centos_stream_9_openstack_yoga_version: 20231005T010906
 stackhpc_pulp_repo_centos_stream_9_opstools_version: 20230615T071742
 stackhpc_pulp_repo_centos_stream_9_storage_ceph_pacific_version: 20230709T010022
-stackhpc_pulp_repo_docker_ce_ubuntu_version: 20231020T014922
+stackhpc_pulp_repo_docker_ce_ubuntu_focal_version: 20231020T014922
+stackhpc_pulp_repo_docker_ce_ubuntu_jammy_version: 20231020T014922
```
markgoddard: These are probably wrong

markgoddard: https://ark.stackhpc.com/pulp/content/docker-ce/ubuntu-focal/
https://ark.stackhpc.com/pulp/content/docker-ce/ubuntu-jammy/

markgoddard: Latest are
- [20240711T233948/](https://ark.stackhpc.com/pulp/content/docker-ce/ubuntu-focal/20240711T233948/)
- [20240711T233948/](https://ark.stackhpc.com/pulp/content/docker-ce/ubuntu-jammy/20240711T233948/)

Alex-Welsh: Latest don't work because they are missing dependencies from other repos which are behind [1], so we'd need to do a full refresh. Any idea what we should do?

[1] https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/9906796640/job/27369403050#step:23:1627

Alex-Welsh: I've switched to the oldest available packages to see if the conflict is gone

markgoddard: If focal is still too new we could revert to the old repo. In your case for jammy you're going to bump the main jammy repos so a newer docker should work.

Alex-Welsh: looks better now

markgoddard (dismissed):

markgoddard:
Do you happen to know the current and new versions of docker for focal and jammy? Might be worth a reno.

Alex-Welsh:
> Do you happen to know the current and new versions of docker for focal and jammy? Might be worth a reno.

Did some ark curling and found the new version is `5:25.0.0-1`, old version is `5:24.0.6-1`, adding a reno now

markgoddard (approved):

PR #1170, opened 2024-07-18
Change the way the Wazuh admin password is generated to always be valid

MaxBed4d:


etc/kayobe/ansible/templates/wazuh-secrets.yml.j2
```diff
@@ -7,7 +7,7 @@ secrets_wazuh:
   # Strengthen default wazuh api user pass
   wazuh_api_users:
     - username: "wazuh"
-      password: "{{ secrets_wazuh.wazuh_api_users[0].password | default(lookup('password', '/dev/null length=30' ), true) }}"
+      password: "{{ secrets_wazuh.wazuh_api_users[0].password | wazuh_password }}"
```
MoteHue:
```suggestion
      password: "{{ secrets_wazuh.wazuh_api_users[0].password | default(wazuh_password) }}"
```

etc/kayobe/ansible/wazuh-secrets.yml
```diff
@@ -14,6 +14,54 @@
         path: "{{ wazuh_secrets_path | dirname }}"
         state: directory
 
+    - name: Generate a random password which meets the Wazuh password requirements
```
dougszumski: Normally, you would put code like this in a filter plugin. It's good that you dug out the requirements.

etc/kayobe/ansible/templates/wazuh-secrets.yml.j2
dougszumski: I think it would be simpler to use:

https://docs.ansible.com/ansible/latest/collections/community/general/random_string_lookup.html

You can configure it to meet the complexity requirements.

Eg. `lookup('community.general.random_string', min_lower=1, min_upper=1, min_special=1, min_numeric=1)`

Or if installing that is a pain, you could use `lookup` to generate a few sub passwords, using limited characters sets, concatenate them together and then randomise the resultant string.

dougszumski (commented):
Thanks @MaxBed4d, this is functional but I think we can do it all in Ansible to make it easier to maintain.

etc/kayobe/ansible/wazuh-secrets.yml
MaxBed4d: I think I have a neater way of doing this.

etc/kayobe/ansible/templates/wazuh-secrets.yml.j2
MaxBed4d: I was thinking that having a small python script which will keep regenerating if requirements aren't met means that there is no chance of the password check failing and may even prevent the password being shown.

dougszumski: It's better to rely on existing code if possible. I think random_string should be configurable to meet the requirements.

MaxBed4d: Updated the method to use the lookup!

Alex-Welsh (approved):

MoteHue (approved):

dougszumski (approved):
Thanks @MaxBed4d

PR #1171, opened 2024-07-18
Encrypt wazuh-secrets.yml once templated

MaxBed4d:


Alex-Welsh:
@MaxBed4d you don't need to make PRs for your own branches :slightly_smiling_face:

MaxBed4d:
> @MaxBed4d you don't need to make PRs for your own branches 🙂

The changes here were to achieve a different goal from the other branch, however this branch relies and builds on the changes from the other branch. 

The two branches felt like two separate PRs, but this could not be merged into ``main`` before the other, so to prevent that, I thought that merging this into the other branch was a better idea.

Alex-Welsh:
> > @MaxBed4d you don't need to make PRs for your own branches 🙂
> 
> The changes here were to achieve a different goal from the other branch, however this branch relies and builds on the changes from the other branch.
> 
> The two branches felt like two separate PRs, but this could not be merged into `main` before the other, so to prevent that, I thought that merging this into the other branch was a better idea.

In the future, one PR for "Wazuh deployment script improvements" with two commits would be fine. I also don't really see how this one relies on the other, they wouldn't conflict.

dougszumski (approved):
Thanks @MaxBed4d - this works fine.

etc/kayobe/ansible/wazuh-secrets.yml
```diff
@@ -15,13 +15,15 @@
...
         src: wazuh-secrets.yml.j2
         dest: "{{ wazuh_secrets_path }}"
-      notify: Please encrypt keys
 
-  handlers:
-    - name: Please encrypt keys
-      debug:
-        msg: >-
-          Please encrypt the keys using Ansible Vault.
+    - name: In-place encrypt wazuh-secrets
```
markgoddard, line 23: The downside of this no longer being a handler is that it will reencrypt the content each time you run the playbook, even if there are no plain text changes.

MaxBed4d: Would something like [this](https://stackoverflow.com/questions/64965281/detect-file-difference-change-with-ansible) work?

markgoddard: Yes, or going back to a handler.

dougszumski: Oops, good spot.

dougszumski: @MaxBed4d please could you also update the doc for the modified process: https://github.com/stackhpc/stackhpc-kayobe-config/blob/Wazuh-Docs-TLS/doc/source/configuration/wazuh.rst#the-short-version

MaxBed4d: @dougszumski Of course! On it!

PR #1173, opened 2024-07-19
Make packet drop alert configurable

Alex-Welsh:


MoteHue (dismissed):
Always helpful with SKC merges when we don't need to carry custom rules around.

dougszumski (commented):
Thanks @Alex-Welsh , this is useful. Please could you extend to packet errors? That has also caused a lot of unwarranted alert spam at some sites.

dougszumski (approved):
Thanks!

PR #1175, opened 2024-07-22
Stop changing permissions on files on Rocky 9

markgoddard:
A similar change was made for Ubuntu systems in #1119, but it did not
apply to Rocky 9 systems. This changes brings the two into line.


jovial (approved):
Very nice - sorry I missed this.

PR #1180, opened 2024-07-23
yoga: Stop changing permissions on files in CIS

markgoddard:
- **Stop changing permissions on files on Rocky 9**
- **Stop changing permissions on files (#1119)**


markgoddard:
How about it @jovial ?

jovial (approved):
Makes sense. I always forget people still use yoga these days :laughing:

PR #1181, opened 2024-07-23
yoga: Reduce multinode LVM volume sizes

markgoddard:
- **Changing LVM layout for controllers. We've been running out of space when testing upgrades in Multinode.**
- **ci-multinode: Further reduce LVM volumes**


Alex-Welsh (approved):
Personally I'd prefer to just increase the disk size but this works for now

PR #1186, opened 2024-07-23
2023.1 Fix CVE-2024-40767

seunghun1ee:


priteau:
You should update `kolla_sources` and the public nova repository so these changes are not lost at the next rebuild.

priteau (approved):
Merging to allow pulling images.

PR #1187, opened 2024-07-23
Bump hashicorp/setup-terraform to v3

priteau:


priteau:
Fails with `this._parser?.end()`. Node version issues?

PR #1200, opened 2024-07-25
Build nova from our fork

priteau:
This is necessary to address OSSA-2024-002 [1] until patches are merged upstream.

[1] https://security.openstack.org/ossa/OSSA-2024-002.html


markgoddard (approved):

markgoddard:
Do we need some zed images still?

priteau:
We still need to build images. But we don't need this change anymore because the security fix was merged upstream 🙃

PR #1205, opened 2024-07-30
Fix issue with prometheus and hyphens in blackbox exporter

technowhizz:
Fixes an issue where prometheus wasn't happy with hyphens in the names of targets


bbezak (approved):

cityofships:
Before:
image

After:
image

cityofships (approved):

cityofships:
One more issue is that `rabbitmq_enable_tls` is not respected / passed correctly

PR #1206, opened 2024-08-01
Bump up Cephadm Ansible collection

cityofships:


etc/kayobe/ansible/requirements.yml
```diff
@@ -1,7 +1,7 @@
 ---
 collections:
   - name: stackhpc.cephadm
-    version: 1.15.1
+    version: 1.17.0
```
markgoddard, line 4: Any notable changes to add to a reno?

jackhodgkiss: I would think these two would be worth a mention which make cephadm key stateless and makes reef default.

https://github.com/stackhpc/ansible-collection-cephadm/pull/145
https://github.com/stackhpc/ansible-collection-cephadm/pull/152

markgoddard: We do override the version, so ref change won't affect us.

markgoddard:
I've proposed bumping to 1.18.0 in https://github.com/stackhpc/stackhpc-kayobe-config/pull/1219 (draft), and added a reno.

MoteHue:
This is good to close, now https://github.com/stackhpc/stackhpc-kayobe-config/pull/1219 is merged?

PR #1209, opened 2024-08-05
yoga: Fix reboot hang on systems with molly-guard, add timeout variable

markgoddard:
- **Prevent hanging before reboot on systems running molly-guard**
- **Add reboot timeout to reboot playbook**


Alex-Welsh (changes requested):
2nd reboot on L110 of the upgrade playbook

Alex-Welsh (approved):

PR #1216, opened 2024-08-08
Add Ansible's YAML callback plugin configuration

darmach:
Added Ansible's YAML callback plugin to Ansible configuration to make errors (finally) more digestible.

### DIFF
--- a/etc/kayobe/ansible.cfg
+++ b/etc/kayobe/ansible.cfg
@@ -10,6 +10,10 @@ inject_facts_as_vars = False
 callbacks_enabled = ansible.posix.profile_tasks
 # Silence warning about invalid characters found in group names
 force_valid_group_names = ignore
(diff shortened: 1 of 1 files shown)


etc/kayobe/ansible.cfg
```diff
@@ -10,6 +10,10 @@ inject_facts_as_vars = False
 callbacks_enabled = ansible.posix.profile_tasks
 # Silence warning about invalid characters found in group names
 force_valid_group_names = ignore
+# Use the YAML callback plugin.
```
markgoddard, line 13: erm... see L3-6

markgoddard (changes requested):

etc/kayobe/ansible.cfg
darmach: OH MY GOD

PR #1222, opened 2024-08-12
Sync antelope caracal

Alex-Welsh:


etc/kayobe/kolla.yml
```diff
@@ -147,6 +147,14 @@ kolla_sources:
...
     location: https://github.com/stackhpc/networking-generic-switch.git
     reference: stackhpc/{{ openstack_release }}
+  octavia-api-plugin-ovn-octavia-provider:
+    type: git
+    location: https://github.com/stackhpc/ovn-octavia-provider.git
+    reference: stackhpc/{{ openstack_release }}
+  octavia-driver-agent-plugin-ovn-octavia-provider:
+    type: git
+    location: https://github.com/stackhpc/ovn-octavia-provider.git
+    reference: stackhpc/{{ openstack_release }}
```
markgoddard, line 157: Do we still need these forks, and if so do Caracal branches exist?

markgoddard: It looks like the upstream bug fix has not yet merged, so we should backport it to Caracal.

PR #1223, opened 2024-08-12
Disable CIS shell timeout

Alex-Welsh:


markgoddard (commented):
What's the reasoning? Just that it's annoying?

Alex-Welsh:
> What's the reasoning? Just that it's annoying?

Yep, it's just a real pain losing your session constantly

etc/kayobe/inventory/group_vars/cis-hardening/cis
```diff
@@ -31,6 +31,9 @@ rhel9cis_rule_1_9: false
 # Disable requirement for password when using sudo
 rhel9cis_rule_5_3_4: false
 
+# Disable shell inactivity timeout
+rhel9cis_rule_5_6_3: false
```
markgoddard, line 35: I guess the issue here is we slowly reduce our CIS score over time by disabling these things. Anyone else have thoughts?

markgoddard:
> > What's the reasoning? Just that it's annoying?
> 
> Yep, it's just a real pain losing your session constantly

I guess the issue here is we slowly reduce our CIS score over time by disabling these things. Anyone else have thoughts?

bbezak:
> > > What's the reasoning? Just that it's annoying?
> > 
> > 
> > Yep, it's just a real pain losing your session constantly
> 
> I guess the issue here is we slowly reduce our CIS score over time by disabling these things. Anyone else have thoughts?

SSH connection timeout is important CIS feature. IMHO If we're enabling CIS, SSH connection timeout should stay

mnasiadka:
> > > > What's the reasoning? Just that it's annoying?
> > > 
> > > 
> > > Yep, it's just a real pain losing your session constantly
> > 
> > 
> > I guess the issue here is we slowly reduce our CIS score over time by disabling these things. Anyone else have thoughts?
> 
> SSH connection timeout is important CIS feature. IMHO If we're enabling CIS, SSH connection timeout should stay

If we're disabling that, we're also lowering the coverage of CIS standard on customer sites - I don't think we can do it just because it's annoying.

PR #1228, opened 2024-08-16
Use default crypto policy in AIOs for ed25519

Alex-Welsh:
AIOs are usually their own ansible control host, so CIS hardening can cut off user access when using ed25519 keys. This change sets the crypto policy back to DEFAULT.

### DIFF
--- a/etc/kayobe/environments/ci-aio/inventory/group_vars/cis-hardening/cis
+++ b/etc/kayobe/environments/ci-aio/inventory/group_vars/cis-hardening/cis
@@ -1,0 +1,3 @@
+---
+# Don't use FIPS since it blocks ed25519 keys
+rhel9cis_crypto_policy: DEFAULT


etc/kayobe/environments/ci-aio/inventory/group_vars/cis-hardening/cis
```diff
@@ -0,0 +1,3 @@
+---
+# Don't use FIPS since it blocks ed25519 keys
+rhel9cis_crypto_policy: DEFAULT
```
markgoddard, line 3: Did the [check](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1096/files) in cis.yml not work for you?

Alex-Welsh: That's only for kayobe/KA keys right? This is so I can use an ed25519 when I get into the AIO myself

markgoddard: ok. There's a bit of tension here between aio usability and testing production config (and potentially hiding issues if we don't)

Alex-Welsh: Fair point. Perhaps the compromise is a precheck in the automated deployment script that checks authorized_keys and breaks/warns the user?

markgoddard: That could work, for those using the script.

markgoddard: Perhaps that could also be added to the cis.yml playbook, to prevent the same issue in production?

markgoddard: With a flag to disable the check if you know what you're doing.

Alex-Welsh: Given that there's a check on the playbook already, I'm going to close this PR for now and potentially come back to it at a later date. At the moment the value isn't worth the effort

PR #1234, opened 2024-08-20
Bump kayobe automation

Alex-Welsh:
This is required for Caracal AIO CI upgrade checks.

A prerequisite for the upgrade is a migration to quorum queues in RabbitMQ, which is executed using a script. The script cannot be run without recent changes to kayobe-automation.

An example of the failure can be found [here](https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/10471423373/job/28998915831?pr=1203#step:29:55)


markgoddard (changes requested):
@jackhodgkiss just pushed a fix to kayobe-automation that we should include.

markgoddard (approved):

PR #1236, opened 2024-08-20
CI: Make .github/workflows a symlink

markgoddard:
This allows client configurations to easily remove the symlink, avoiding
skipped workflows showing up in client PRs.


markgoddard:
Doesn't seem to work :(

PR #1237, opened 2024-08-20
Work around Rocky OVN Caracal bug for upgrade job

Alex-Welsh:


terraform/aio/vm.tf
```diff
@@ -89,10 +89,12 @@ resource "null_resource" "kayobe-aio" {
...
       script_path = "/home/${var.ssh_username}/start.sh"
     }
-
+    # In Caracal, OVN fails on Rocky 9 hosts with hostnames that contain a
+    # subdomain (a dot in the name). Required here for the Caracal upgrade Job
     inline = [
       "#!/bin/sh",
-      "echo 'connected!'"
+      "echo 'connected!'",
+      "sudo hostnamectl set-hostname $(sed s/.novalocal// /etc/hostname)"
```
markgoddard: Is there going to be an issue if we don't reboot the host to apply the new hostname?

markgoddard: e.g. deploy an aio without rebooting, then later reboot

Alex-Welsh: Not sure, but the CI here and on [this PR](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1233/) will tell us if anything breaks

markgoddard: It won't test my example though, will it?

Alex-Welsh: True, though I'd be inclined to accept that as a risk. It's only a temporary workaround, given that I'm assuming we'll have to fix it upstream soon so we don't hit the problem in production

MoteHue: > deploy an aio without rebooting, then later reboot

This would be fairly common with AIOs spawned for dev work outside of CI.

Alex-Welsh: New plan, adapting the fix-hostname playbook to our needs. Testing [here](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1233/)

etc/kayobe/inventory/groups
```diff
@@ -132,6 +132,9 @@ rgws
 [cis-hardening:children]
 overcloud
 
+[reboot_hosts:children]
+storage
+
```
markgoddard: This is only used for multinodes. Move to ci-multinode env?

etc/kayobe/ansible/fix-hostname.yml
```diff
@@ -1,6 +1,6 @@
 ---
-- name: Fix hostname on storage nodes for cephadm
-  hosts: storage
+- name: Fix hostname
+  hosts: reboot_hosts
```
markgoddard: This group doesn't really match the name or content of the playbook

markgoddard (changes requested):

etc/kayobe/inventory/groups
Alex-Welsh: Done

etc/kayobe/ansible/fix-hostname.yml
Alex-Welsh: Somehow managed to write that three times without realising it was wrong. Fixed now

Alex-Welsh:
Works!
https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/10488024781/job/29049828546?pr=1233

markgoddard (approved):

PR #1244, opened 2024-08-22
Explicitly set rabbit flags for MN-CI upgrades

Alex-Welsh:
See https://github.com/stackhpc/stackhpc-kayobe-config/blob/7daaead629796270f183f801c4ec5929eff7796d/etc/kayobe/environments/ci-aio/kolla/globals.yml#L18
for the same format in ci-aio


MoteHue (dismissed):
"Set durable queue flags" not quite a correct description, but not worth blocking over

Alex-Welsh:
> "Set durable queue flags" not quite a correct description, but not worth blocking over

Oh but it is

MoteHue (approved):

PR #1248, opened 2024-08-23
Pinning kolla, kolla-ansible and kayobe in caracal

darmach:
Pinned Kolla, Kolla-Ansible and Kayobe to tags - Caracal:
 - Kolla pinned to stackhpc/18.1.0.2
 - Kolla-Ansible pinned to stackhpc/18/1/0/2
 - Kayobe pinned to stackhpc/16.1.0.1
 
 To be merged after @Alex-Welsh merges https://review.opendev.org/c/openstack/kolla-ansible/+/926618

### DIFF
--- a/etc/kayobe/stackhpc.yml
+++ b/etc/kayobe/stackhpc.yml
@@ -141,15 +141,11 @@ stackhpc_repo_elrepo_9_version: "{{ stackhpc_repo_distribution }}"
 
 # Kolla source repository.
 stackhpc_kolla_source_url: "https://github.com/stackhpc/kolla"
-stackhpc_kolla_source_version: "stackhpc/{{ openstack_release }}"
+stackhpc_kolla_source_version: "stackhpc/18.1.0.2"
 
 # Kolla Ansible source repository.
 stackhpc_kolla_ansible_source_url: "https://github.com/stackhpc/kolla-ansible"
-# FIXME: Waiting for
-# https://review.opendev.org/c/openstack/kolla-ansible/+/926198 to merge and
-# sync to stackhpc/2024.1
-# stackhpc_kolla_ansible_source_version: "stackhpc/{{ openstack_release }}"
-stackhpc_kolla_ansible_source_version: "fix-prometheus"
+stackhpc_kolla_ansible_source_version: "stackhpc/18.1.0.2"
 
 ###############################################################################
 # Container image registry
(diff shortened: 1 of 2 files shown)


bbezak:
shouldn't we automate the pinning process? I don't think manual versions updating is a way to go

etc/kayobe/stackhpc.yml
```diff
@@ -141,15 +141,11 @@ stackhpc_repo_elrepo_9_version: "{{ stackhpc_repo_distribution }}"
...
 stackhpc_kolla_source_url: "https://github.com/stackhpc/kolla"
-stackhpc_kolla_source_version: "stackhpc/{{ openstack_release }}"
+stackhpc_kolla_source_version: "stackhpc/18.1.0.2"
 
 # Kolla Ansible source repository.
 stackhpc_kolla_ansible_source_url: "https://github.com/stackhpc/kolla-ansible"
-# FIXME: Waiting for
-# https://review.opendev.org/c/openstack/kolla-ansible/+/926198 to merge and
-# sync to stackhpc/2024.1
-# stackhpc_kolla_ansible_source_version: "stackhpc/{{ openstack_release }}"
```
markgoddard, line 151: https://github.com/stackhpc/kolla-ansible/pull/596, available in stackhpc/18.1.0.4...

etc/kayobe/stackhpc.yml
markgoddard, line 1: It would be better to move these versions to a separate file to allow for easier automation. `stackhpc-source-versions.yml`?

markgoddard (changes requested):

markgoddard:
> shouldn't we automate the pinning process? I don't think manual versions updating is a way to go

Agree, this is just the first step.

Alex-Welsh:
@darmach I think all the critical fixes are merged upstream now. We can merge this if you rebase & update the pins

mnasiadka:
Maybe we're a bit naive in here, who is going to pick up the burden of maintaining the pins before we get to an automated solution?

Alex-Welsh:
> Maybe we're a bit naive in here, who is going to pick up the burden of maintaining the pins before we get to an automated solution?

Whoever needs the most recent changes probably

mnasiadka:
> > Maybe we're a bit naive in here, who is going to pick up the burden of maintaining the pins before we get to an automated solution?
> 
> Whoever needs the most recent changes probably

That's the naive part - who is going to watch for reasonable patches that come from upstream that would be beneficial to pull in? (especially security related patches).
One option is that it would be the same person approving PRs in Kayobe and Kolla-Ansible downstream forks - but would prefer an automated one (and @darmach has been so extremely fast with that one, that I don't know if we can put the burden of automating it on him) ;-)

Alex-Welsh:
We should probably close this for now, since we decided to not pin anything until we can automate it

Alex-Welsh:
Superseded by https://github.com/stackhpc/stackhpc-kayobe-config/pull/1532

PR #1252, opened 2024-08-30
Add missing -r option (pip)

seunghun1ee:
-r option is required to parse the path as requirements file


Alex-Welsh (approved):
Not sure how we missed that

markgoddard (commented):
oops, my bad

PR #1254, opened 2024-09-02
Further fixes to Prometheus blackbox exporter endpoints

MoteHue:
Enabled items under prometheus_blackbox_exporter_endpoints_kayobe were not picked up because the attribute gets set to `enabled: 'True'`, rather the than `enabled: true` that the default values get. Likely a similar issue to this:
https://github.com/ansible/ansible/issues/11905#issuecomment-130496173

### DIFF
--- a/etc/kayobe/kolla/inventory/group_vars/prometheus-blackbox-exporter
+++ b/etc/kayobe/kolla/inventory/group_vars/prometheus-blackbox-exporter
@@ -149,7 +149,7 @@ heat_cfn_internal_base_endpoint: "{{ heat_cfn_internal_fqdn | kolla_url(internal
 heat_cfn_public_base_endpoint: "{{ heat_cfn_external_fqdn | kolla_url(public_protocol, heat_api_cfn_public_port) }}"
 heat_internal_base_endpoint: "{{ heat_internal_fqdn | kolla_url(internal_protocol, heat_api_port) }}"
 heat_public_base_endpoint: "{{ heat_external_fqdn | kolla_url(public_protocol, heat_api_public_port) }}"
-horizon_public_endpoint: "{{ horizon_external_fqdn | kolla_url(public_protocol, horizon_listen_port) }}"
+horizon_public_endpoint: "{{ horizon_external_fqdn | kolla_url(public_protocol, horizon_tls_port if kolla_enable_tls_external | bool else horizon_port) }}"
 ironic_inspector_internal_endpoint: "{{ ironic_inspector_internal_fqdn | kolla_url(internal_protocol, ironic_inspector_port) }}"
 ironic_inspector_public_endpoint: "{{ ironic_inspector_external_fqdn | kolla_url(public_protocol, ironic_inspector_public_port) }}"
 magnum_internal_base_endpoint: "{{ magnum_internal_fqdn | kolla_url(internal_protocol, magnum_api_port) }}"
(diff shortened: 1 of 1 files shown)


Alex-Welsh (commented):
It should be using the [true jinja test](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-tests.true) rather than matching against the string, so I'd assume it would accept any truthy value. Evidently not

Alex-Welsh (dismissed):

etc/kayobe/kolla/inventory/group_vars/prometheus-blackbox-exporter
```diff
@@ -9,7 +9,7 @@
 prometheus_blackbox_exporter_endpoints: >-
   {{ (prometheus_blackbox_exporter_endpoints_kayobe | default([]) +
   prometheus_blackbox_exporter_endpoints_default) |
-  selectattr('enabled', 'true') |
+  selectattr('enabled', 'match', 'true|True') |
```
markgoddard: Would this ignore any that use a proper boolean true, or "yes"?

MoteHue: Rewritten this so it now captures all truthy values correctly. It follows the same format/convention @Alex-Welsh has in his firewalld config PR

etc/kayobe/pulp.yml
```diff
@@ -532,6 +532,8 @@ stackhpc_pulp_images_kolla:
   - prometheus-openstack-exporter
   - prometheus-v2-server
   - rabbitmq
+  - rabbitmq-3-12
+  - rabbitmq-3-13
```
Alex-Welsh: Stray config?

MoteHue: Yep, it's gone

Alex-Welsh (commented):
LGTM but I helped write it so someone else should approve

Alex-Welsh (commented):
Slight scope creep but I'll allow it. LGTM but someone else should approve

MoteHue:
> Slight scope creep but I'll allow it.

I've kept the changes in separate commits, so at least it will be distinct in the blame.

etc/kayobe/kolla/inventory/group_vars/prometheus-blackbox-exporter
```diff
@@ -6,13 +6,16 @@
...
+prometheus_blackbox_exporter_endpoints: |
+  {% set endpoints = [] %}
+  {% for dict_item in (prometheus_blackbox_exporter_endpoints_kayobe | default([]) + prometheus_blackbox_exporter_endpoints_default) %}
+  {% if dict_item.enabled | bool %}
+  {% for endpoint in dict_item.endpoints %}
+  {% set _ = endpoints.append(endpoint) %}
+  {% endfor %}
+  {% endif %}
+  {% endfor %}
+  {{ (endpoints + prometheus_blackbox_exporter_endpoints_custom) | unique | list }}
```
markgoddard: There is a `select` in the original filter pipeline. Is it possible that an endpoint could be set to an empty string?

Alex-Welsh: I think if `kolla_same_external_internal_vip` is true, then there will be a lot of empty string endpoints, so select would probably be quite important

MoteHue: Done

markgoddard (approved):

PR #1255, opened 2024-09-02
zed: ci-multinode: Add Kolla Ansible TLS config to globals.yml

markgoddard:
Previously we provided an additional file, globals-tls-config.yml, which
would be appended to globals.yml at an appropriate point during
multinode deployment. This could result in a merge conflict, which
cannot be resolved by an automated process such as a GitHub Actions
workflow.

This change adds the TLS config to globals.yml conditionally, based on
whether internal TLS is enabled.

We are not removing globals-tls-config.yml for now, since it is used in
the terraform-kayobe-multinode repo as a proxy for internal TLS support.
The file has already been removed in the Antelope branch.


markgoddard:
CI now passing.

Alex-Welsh (approved):

PR #1257, opened 2024-09-02
Add hooks for fixing hostname bug in Multinode for Caracal upgrade job

Alex-Welsh:
Upgrades in Multinodes in Caracal will currently fail in CI upgrade jobs because they are deployed with FQDNs in Antelope. This change hooks in the fix-hostname and reboot playbooks to make sure they are fixed before any openstack services get deployed

### DIFF
--- a/etc/kayobe/ansible/fix-hostname.yml
+++ b/etc/kayobe/ansible/fix-hostname.yml
@@ -21,3 +21,9 @@
         cmd: hostnamectl set-hostname "{{ inventory_hostname }}"
       when: current_hostname.stdout != inventory_hostname
       become: true
+
+- name: Reboot hosts
+  import_playbook: "{{ playbook_dir | realpath }}/reboot.yml"
+  vars:
+    reboot_hosts: fix-hostname
+  when: current_hostname.stdout != inventory_hostname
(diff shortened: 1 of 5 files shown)


seunghun1ee (dismissed):

etc/kayobe/environments/ci-multinode/hooks/overcloud-host-configure/pre.d/20-reboot.yml
```diff
@@ -0,0 +1 @@
+../../../../../ansible/reboot.yml
```
markgoddard: This is a bad idea. It would hook a reboot into every host configure command, and would be pretty surprising and potentially damaging. Is there another way this can be fixed?

markgoddard (changes requested):

etc/kayobe/environments/ci-multinode/hooks/overcloud-host-configure/pre.d/20-reboot.yml
markgoddard: Perhaps apply the reboot from the fix-hostname.yml playbook if a change in hostname is detected?

Alex-Welsh: Done

markgoddard (approved):

PR #1270, opened 2024-09-09
CIS - don't shut down when out of space

Alex-Welsh:
Currently, CIS hardening will shut down hosts when they run out of space for logs. They then can't be started again without significant time and effort.

This change will instead make them log warnings instead of shutting down.

### DIFF
--- a/etc/kayobe/inventory/group_vars/cis-hardening/cis
+++ b/etc/kayobe/inventory/group_vars/cis-hardening/cis
@@ -55,7 +55,8 @@ rhel9cis_no_world_write_adjust: false
 rhel9cis_auditd:
   space_left_action: syslog
   action_mail_acct: root
-  admin_space_left_action: halt
+  admin_space_left_action: syslog
+  max_log_file: 10
   max_log_file_action: rotate
 
 # Max size of audit logs (MB)
@@ -156,7 +157,7 @@ ubtu22cis_suid_adjust: false
 ubtu22cis_auditd:
   action_mail_acct: root
   space_left_action: syslog
-  admin_space_left_action: halt
+  admin_space_left_action: syslog
   max_log_file_action: rotate
 
 # Max size of audit logs (MB)


etc/kayobe/inventory/group_vars/cis-hardening/cis
```diff
@@ -55,7 +55,8 @@ rhel9cis_no_world_write_adjust: false
 rhel9cis_auditd:
   space_left_action: syslog
   action_mail_acct: root
-  admin_space_left_action: halt
+  admin_space_left_action: syslog
+  max_log_file: 10
```
technowhizz, line 59: Why are we adding max log file here?

technowhizz:
I thought we had changed this. Good catch @Alex-Welsh

mnasiadka:
As long as we don't have a solution for shipping audit logs to a separate location - that's probably the sensible option.

etc/kayobe/inventory/group_vars/cis-hardening/cis
priteau: > This keyword specifies the maximum file size in megabytes. When this limit is reached, it will trigger a configurable action. The value given must be numeric.

10 MB is quite small.

Alex-Welsh: It's just specifying the default
https://github.com/ansible-lockdown/RHEL9-CIS/blob/devel/defaults/main.yml#L798

I can remove it if you want?

Alex-Welsh:
Any more opinions on this @mnasiadka @priteau?

etc/kayobe/inventory/group_vars/cis-hardening/cis
markgoddard: Let's keep the default for now, consider a separate PR for increasing the size?

markgoddard (approved):

PR #1279, opened 2024-09-11
growroot: Fix regex for md (sw-raid) cases

mnasiadka:


grzegorzkoper (approved):
LGTM +1
Tested in SMSLAB

markgoddard:
How does it relate to #818?

mnasiadka:
> How does it relate to #818?

It doesn't, I haven't seen it - and given the endless thread in there and no resolution - maybe we can merge something that is not ideal but still helpful?

markgoddard:
> > How does it relate to #818?
> 
> It doesn't, I haven't seen it - and given the endless thread in there and no resolution - maybe we can merge something that is not ideal but still helpful?

The issue is that there's scope for regression, and other related issues.

mnasiadka:
> > > How does it relate to #818?
> > 
> > 
> > It doesn't, I haven't seen it - and given the endless thread in there and no resolution - maybe we can merge something that is not ideal but still helpful?
> 
> The issue is that there's scope for regression, and other related issues.

Let's go with #818, I tested it on all three possible disk names.

PR #1280, opened 2024-09-11
Reboot with bootstrap user in ci-aio env

Alex-Welsh:


etc/kayobe/environments/ci-aio/reboot.yml
```diff
@@ -0,0 +1,3 @@
+---
+# Ensure that the reboot playbook is always executed using the boostrap user
+reboot_with_bootstrap_user: true
```
markgoddard: Can it not be applied at the import_playbook level?

Alex-Welsh: I like the idea, I'll try it out

markgoddard (approved):
LGTM if it works

Alex-Welsh:
https://github.com/stackhpc/stackhpc-kayobe-config/pull/1281
works :)

PR #1293, opened 2024-09-17
Update secret rotation doc

seunghun1ee:
During Grafana password update, I found that there can be a case that the ID of ``grafana_local_admin`` is not 1 (Default admin ID).

Added a note to address the case.

### DIFF
--- a/doc/source/operations/secret-rotation.rst
+++ b/doc/source/operations/secret-rotation.rst
@@ -74,7 +74,7 @@ Full method
    the state of the cloud before any changes are made
 
 2. Edit your Kolla-Ansible checkout to include changes not yet included
-   upstream. 
+   upstream.
 
 .. _kolla-change:
 
@@ -98,7 +98,7 @@ Full method
       .. code:: bash
 
          git fetch https://review.opendev.org/openstack/kolla-ansible refs/changes/78/903178/2 && git cherry-pick FETCH_HEAD
-   
+
    3. Re-install Kolla-Ansible from source in your Kolla-Ansible Python
       environment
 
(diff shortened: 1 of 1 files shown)


doc/source/operations/secret-rotation.rst
```diff
@@ -129,7 +129,7 @@ Full method
       ^redis_master_password
       ^memcache_secret_key
       _ssh_key
-         
+
```
Alex-Welsh, line 132: Are you sure this whitespace isn't required?

doc/source/operations/secret-rotation.rst
```diff
@@ -234,6 +234,28 @@ Full method
 
           grafana-cli admin reset-admin-password --password-from-stdin
 
+    .. note::
```
Alex-Welsh: Could you please add this to the regular instructions rather than as a note? I think we should just get everyone to check the user ID, rather than letting some people get the error and have to go back and do it again

doc/source/operations/secret-rotation.rst
seunghun1ee: This was unintended, my VScode did it automatically I think.
However, I compared with current doc and copying the block of this code gives same result between current version and this change.
If we want a precise code block, it might be better to provide a link to raw content of github gist.
Which can be downloaded by using cURL or wget

doc/source/operations/secret-rotation.rst
seunghun1ee: Sounds good.

Alex-Welsh (approved):

PR #1294, opened 2024-09-18
(2024.1) Reintroduce networking-mlnx

jovial:
Building here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/10926534480


MoteHue (approved):
LGTM, I double-checked the images are promoted too

PR #1307, opened 2024-09-27
Changes to default tempest concurrency

Alex-Welsh:
Increases the default tempest concurrency from 2 to 16 and adds an explanatory comment.
Increased the Multinode concurrency from 1 to 4


seunghun1ee (approved):
Let's go

jovial (approved):
Should speed up runs - looks good

PR #1308, opened 2024-09-27
Add upgrade-prerequisites script for CI

Alex-Welsh:
Adds a script to automate any pre-upgrade actions. Designed to be used in CI

### DIFF
--- a/tools/upgrade-prerequisites.sh
+++ b/tools/upgrade-prerequisites.sh
@@ -19,11 +19,11 @@ function prechecks() {
 
 function rabbit_upgrade() {
     # Ensure RabbitMQ is upgraded to 3.13
-    if kayobe overcloud host command run -l controllers -b --command "docker exec rabbitmq rabbitmqctl --version | grep -F 3.11" --show-output; then
+    if kayobe overcloud host command run -l controllers -b --command "docker exec rabbitmq rabbitmqctl --version | grep -F 3.11." --show-output; then
         kayobe kolla ansible run "rabbitmq-upgrade 3.12"
     fi
     sleep 200
-    if kayobe overcloud host command run -l controllers -b --command "docker exec rabbitmq rabbitmqctl --version | grep -F 3.12" --show-output; then
+    if kayobe overcloud host command run -l controllers -b --command "docker exec rabbitmq rabbitmqctl --version | grep -F 3.12." --show-output; then
         kayobe kolla ansible run "rabbitmq-upgrade 3.13"
     fi
 }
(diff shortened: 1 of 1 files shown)


Alex-Welsh:
See also https://github.com/stackhpc/terraform-kayobe-multinode/pull/73

seunghun1ee:
I like this approach. More generalised.

seunghun1ee (dismissed):

tools/upgrade-prerequisites.sh
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function rabbit_migration() {
+    if ! kayobe overcloud host command run -l controllers -b --command "docker exec $RABBITMQ_CONTAINER_NAME rabbitmqctl list_queues type | grep quorum"; then
+        # Set quorum flag, execute RabbitMQ queue migration script, unset quorum flag (to avoid git conflicts)
+        sed -i -e 's/om_enable_rabbitmq_high_availability: true/om_enable_rabbitmq_high_availability: false/' \
+               -e 's/om_enable_rabbitmq_quorum_queues: false/om_enable_rabbitmq_quorum_queues: true/' \
+            $KAYOBE_CONFIG_PATH/environments/$KAYOBE_ENVIRONMENT/kolla/globals.yml
+
+        $KAYOBE_CONFIG_PATH/../../tools/rabbitmq-quorum-migration.sh
```
MoteHue:
```suggestion
        $KAYOBE_CONFIG_ROOT/tools/rabbitmq-quorum-migration.sh
```

tools/upgrade-prerequisites.sh
```diff
@@ -0,0 +1,48 @@
...
+    if [[ ! $KAYOBE_CONFIG_PATH ]]; then
+        echo "Environment variable \$KAYOBE_CONFIG_PATH is not defined"
+        echo "Ensure your environment is set up to run kayobe commands"
+        exit 2
+    fi
+}
+
+function rabbit_upgrade() {
+    # Ensure RabbitMQ is upgraded to 3.13
+    if kayobe overcloud host command run -l controllers -b --command "docker exec rabbitmq rabbitmqctl --version | grep -F 3.11" --show-output; then
```
MoteHue: I think this should be a safe check, given RabbitMQ are moving on to versions 4.x now. But it is possible this could pick up when the two minor points of a version are 3.11. e.g: 3.13.11. 

Probably not an issue, but we could change the grep anyway?

```suggestion
    if kayobe overcloud host command run -l controllers -b --command "docker exec rabbitmq rabbitmqctl --version | grep -F 3.11." --show-output; then
```

tools/upgrade-prerequisites.sh
```diff
@@ -0,0 +1,48 @@
...
+        kayobe kolla ansible run "rabbitmq-upgrade 3.12"
+    fi
+    sleep 200
+    if kayobe overcloud host command run -l controllers -b --command "docker exec rabbitmq rabbitmqctl --version | grep -F 3.12" --show-output; then
```
MoteHue: ditto
```suggestion
    if kayobe overcloud host command run -l controllers -b --command "docker exec rabbitmq rabbitmqctl --version | grep -F 3.12." --show-output; then
```

MoteHue (changes requested):

MoteHue (approved):

PR #1310, opened 2024-10-01
Validate inputs to github workflows

Alex-Welsh:
Adds some simple validation to manually-run GitHub workflows to avoid common issues


MoteHue (approved):
Made all the actions fail from this branch, works a charm.

PR #1322, opened 2024-10-10
add multiarch (aarch64 + amd64) container image builds for R9

bbezak:
add multiarch (aarch64 + amd64) container image builds for R9

Five persistent ARM64 runners now share a single bare-metal host in SMS Lab.
The workflow is updated to work in that non-ephemeral environment.

- add rocky-9/aarch64 to the build matrix (for multiarch manifests)
- route ARM jobs to ["self-hosted","sms","arm64"]
- wipe $GITHUB_WORKSPACE at job start and keep Kayobe state in
$GITHUB_WORKSPACE/opt/kayobe
- on ARM runners, prune only images built in the current run
- template repo URLs and snapshot versions on {{ kolla_base_arch }}


.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -205,14 +217,14 @@ jobs:
           kayobe seed container image build $args
         env:
           KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
-        if: inputs.seed
+        if: inputs.seed && matrix.arch == 'amd64'
```
Alex-Welsh: Could we also build aarch Seed images?

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -71,18 +71,19 @@ jobs:
       # Dynamically define job matrix.
       # We need a separate matrix entry for each distribution, when the relevant input is true.
       # https://stackoverflow.com/questions/65384420/how-do-i-make-a-github-action-matrix-element-conditional
+      # For now include only RL9 in aarch64
```
Alex-Welsh, line 98: Also worth adding a `#NOTE` explaining that we have to build arm and x86 together for the multiarch docker manifest

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -297,6 +309,120 @@ jobs:
...
+        env:
+          KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
+
+      - name: Create and Push Docker Manifests
+        run: |
+          set -ex
+          mkdir -p logs
+          images=$(cat all-pushed-images.txt | sort | uniq)
+          # Filter out Ubuntu and Rocky Bifrost images
+          manifest_images=$(echo "$images" | grep 'rocky' | grep -E '.*-(amd64|aarch64)$' | sed -E 's/-(amd64|aarch64)$//' | sort | uniq)
```
Alex-Welsh: Is `grep 'rocky'` not redundant here? Since only the Rocky images have the arch suffix

etc/kayobe/environments/ci-builder/stackhpc-ci.yml
Alex-Welsh, line 1: It would be cleaner to do this in pulp.yml using the same sort of format we have for the regular Rocky repos

.github/workflows/stackhpc-container-image-build.yml
bbezak: > Could we also build aarch Seed images?

bifrost is unbuildable upstream for now

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

etc/kayobe/environments/ci-builder/stackhpc-ci.yml
bbezak: > It would be cleaner to do this in pulp.yml using the same sort of format we have for the regular Rocky repos

I'll rework it similarly as those:

`# pulp-repo-versions.yml for the selected minor release.
stackhpc_pulp_repo_rocky_9_baseos_version: "{{ lookup('vars', 'stackhpc_pulp_repo_rocky_9_' ~ stackhpc_pulp_repo_rocky_9_minor_version ~ '_baseos' ~ ('_aarch64' if kolla_base_arch == 'aarch64' else '') ~ '_version') }}"
stackhpc_pulp_repo_rocky_9_appstream_version: "{{ lookup('vars', 'stackhpc_pulp_repo_rocky_9_' ~ stackhpc_pulp_repo_rocky_9_minor_version ~ '_appstream' ~ ('_aarch64' if kolla_base_arch == 'aarch64' else '') ~ '_version') }}"
stackhpc_pulp_repo_rocky_9_extras_version: "{{ lookup('vars', 'stackhpc_pulp_repo_rocky_9_' ~ stackhpc_pulp_repo_rocky_9_minor_version ~ '_extras' ~ ('_aarch64' if kolla_base_arch == 'aarch64' else '') ~ '_version') }}"
stackhpc_pulp_repo_rocky_9_crb_version: "{{ lookup('vars', 'stackhpc_pulp_repo_rocky_9_' ~ stackhpc_pulp_repo_rocky_9_minor_version ~ '_crb' ~ ('_aarch64' if kolla_base_arch == 'aarch64' else '') ~ '_version') }}"
stackhpc_pulp_repo_rocky_9_highavailability_version: "{{ lookup('vars', 'stackhpc_pulp_repo_rocky_9_' ~ stackhpc_pulp_repo_rocky_9_minor_version ~ '_highavailability' ~ ('_aarch64' if kolla_base_arch == 'aarch64' else '') ~ '_version') }}"
`

bbezak: > It would be cleaner to do this in pulp.yml using the same sort of format we have for the regular Rocky repos

I introduced `stackhpc_pulp_repo_multiarch*` variables. let's see if that'll work well. Not sure if you had sth like that in mind

bbezak: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/11573842692/job/32216809694

https://github.com/stackhpc/stackhpc-kayobe-config/pull/1322/files#diff-a81dee0d818c231fc153d50d2a54c4d1301ef917dbc72f61bc91cf6d5cb291c8R221-R241

bbezak:
this is still running cross arch arm builds - which are a bit slower then amd64 - https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/14084922544. Ideally we would use runners on new arm boxes in sms for that - however that is not yet deployed

Alex-Welsh:
The content changes look good, I think we can rebase, rebuild images, and merge if we want.

It depends on the customer requirements but we might also be better off merging this into Epoxy and just having it as a new feature from there onwards

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -134,6 +139,10 @@ jobs:
       - generate-tag
       - runner-selection
     steps:
+      - name: Purge workspace (Arm runner only)
```
Alex-Welsh: Might be a good idea to also run this on the x86 runners as well. Even if it doesn't really do anything, it's nice if there's less unique things between the workflows

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -211,10 +221,19 @@ jobs:
         continue-on-error: true
         run: |
           args="${{ inputs.regexes }}"
+          if [[ "${{ matrix.distro.arch }}" == 'aarch64' ]]; then
+            args="$args -e kolla_base_arch=${{ matrix.distro.arch }}"
```
Alex-Welsh, line 216: `kolla_base_arch` defaults to `"{{ ansible_facts.architecture }}"`. Can't we use that?

https://github.com/stackhpc/kayobe/blob/stackhpc/2024.1/ansible/inventory/group_vars/all/kolla#L57

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -211,10 +221,19 @@ jobs:
...
           args="$args -e kolla_base_distro=${{ matrix.distro.name }}"
           args="$args -e kolla_base_distro_version=${{ matrix.distro.release }}"
-          args="$args -e kolla_tag=${{ steps.write-kolla-tag.outputs.kolla-tag }}"
+          if [[ "${{ matrix.distro.name }}" == 'rocky' ]]; then
+            args="$args -e kolla_tag=${{ steps.write-kolla-tag.outputs.kolla-tag }}-${{ matrix.distro.arch }}"
+          else
+            args="$args -e kolla_tag=${{ steps.write-kolla-tag.outputs.kolla-tag }}"
+          fi
           args="$args -e stackhpc_repo_mirror_auth_proxy_enabled=true"
+          args="$args -e kolla_build_log_path=/var/log/kolla-build-${{ steps.write-kolla-tag.outputs.kolla-tag }}.log"
```
Alex-Welsh: Why not write these straight to `image-build-logs/kolla-build-overcloud.log`? They get moved there later

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -239,14 +264,14 @@ jobs:
           kayobe seed container image build $args
         env:
           KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
-        if: inputs.seed
+        if: inputs.seed && matrix.distro.arch == 'amd64'
```
Alex-Welsh, line 252: I think I asked this before but forgot the answer, why don't we build bifrost for arm as well?

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -331,6 +356,124 @@ jobs:
...
+        run: |
+          sudo apt update
+          sudo apt install -y git unzip python3-wheel python3-pip python3-venv curl jq wget openssh-server openssh-client
+      - name: Install gh
+        run: |
+          sudo mkdir -p -m 755 /etc/apt/keyrings && wget -qO- https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo tee /etc/apt/keyrings/githubcli-archive-keyring.gpg > /dev/null
+          sudo chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg
+          echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+          sudo apt update
+          sudo apt install gh -y
```
Alex-Welsh: Unused?

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -331,6 +356,124 @@ jobs:
...
+    runs-on: ${{ needs.runner-selection.outputs.runner_name_container_image_build }}
+    permissions: {}
+    needs:
+      - container-image-build
+      - runner-selection
+    steps:
+      - name: Install package dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y git unzip python3-wheel python3-pip python3-venv curl jq wget openssh-server openssh-client
```
Alex-Welsh: nit: These deps can be trimmed down a bit, but I'm not worried about getting it perfect because I know how long it takes to iterate on these.

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -331,6 +356,124 @@ jobs:
...
+          localhost ansible_connection=local ansible_python_interpreter=/usr/bin/python3
+          EOF
+      # See etc/kayobe/ansible/roles/pulp_auth_proxy/README.md for details.
+      # NOTE: We override pulp_auth_proxy_conf_path to a path shared by the
+      # runner and dind containers.
+      - name: Deploy an authenticating package repository mirror proxy
+        run: |
+          source venvs/kayobe/bin/activate &&
+          source src/kayobe-config/kayobe-env --environment ci-builder &&
+          kayobe playbook run $KAYOBE_CONFIG_PATH/ansible/pulp-auth-proxy.yml -e pulp_auth_proxy_conf_path=/home/runner/_work/pulp_proxy
```
Alex-Welsh: unused?

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -331,6 +356,124 @@ jobs:
...
+        uses: actions/download-artifact@v4
+
+      - name: Combine pushed images lists
+        run: |
+          find . -name 'push-attempt-images.txt' -exec cat {} + > all-pushed-images.txt
+      - name: Log in to Docker registry
+        run: |
+          source venvs/kayobe/bin/activate &&
+          source src/kayobe-config/kayobe-env --environment ci-builder &&
+          kayobe playbook run ${KAYOBE_CONFIG_PATH}/ansible/docker-registry-login.yml
```
Alex-Welsh: I'm wondering, if this is the only time kayobe is used in this action, we could reduce the complexity and time a lot by replacing it with a simple docker login. Then we don't need to set up kayobe at all

We would need to inject the creds, but that can be done with GH secrets

.github/workflows/stackhpc-container-image-build.yml
```diff
@@ -331,6 +356,124 @@ jobs:
...
+              fi
+            done
+            if [ -n "$arch_images" ]; then
+              echo "Creating manifest for $base_image with images:$arch_images" | tee -a logs/manifest-creation.log
+              docker manifest create "$base_image" $arch_images | tee -a logs/manifest-creation.log
+              docker manifest push "$base_image" | tee -a logs/manifest-creation.log
+            else
+              echo "No images found for $base_image, skipping." | tee -a logs/manifest-creation.log
+            fi
+          done
```
Alex-Welsh: nit: this is quite long for a GH action task, ideally it would be better to make it a separate script in the `tools/` directory

.github/workflows/stackhpc-container-image-build.yml
bbezak: this workflow also supports cross-arch builds - which we could fallback to as well easily - hence this var based on matrix - not real hardware

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

.github/workflows/stackhpc-container-image-build.yml
bbezak: bifrost is currently not buildable in kolla upstream - it needs some work there

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

.github/workflows/stackhpc-container-image-build.yml
bbezak: done

Alex-Welsh (approved):

PR #1327, opened 2024-10-11
INFRA-388 Converting smartmon into python and adding mock tests

technowhizz:


etc/kayobe/ansible/scripts/smartmon.py
```diff
@@ -0,0 +1,156 @@
...
+    result = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+    if parse_json:
+        return json.loads(result.stdout)
+    else:
+        return result.stdout.strip()
+
+def parse_smartctl_attributes(disk, disk_type, serial, json_data):
+    labels = f'disk="{disk}",type="{disk_type}",serial_number="{serial}"'
+    metrics = []
+    smartmon_attrs = set([
```
dougszumski: style nit: It would be helpful to move these to a 'constant' eg. `SMARTMON_ATTRS = set(..` and put each attribute on a new line.

etc/kayobe/ansible/scripts/smartmon.py
```diff
@@ -0,0 +1,156 @@
...
+            output.append(f"# TYPE smartmon_{metric_name} gauge")
+            last_metric = metric_name
+        output.append(f"smartmon_{metric}")
+    return '\n'.join(output)
+
+def main():
+    try:
+        version_output = run_command([SMARTCTL_PATH, '-j'], parse_json=True)
+        smartctl_version_list = version_output.get('smartctl', {}).get('version', [])
+        if smartctl_version_list:
```
dougszumski: Have you looked at Python libraries for doing this?

Eg. This one looks reasonable: https://pypi.org/project/pySMART/

It can handle loading all the device metrics in via `smartctl` as you are doing here. 

It seems to provide some abstraction over the health state as well.

The work you've put into the tests should be directly usable

etc/kayobe/ansible/scripts/test_smartmon.py
```diff
@@ -0,0 +1,265 @@
...
+from smartmon import (
+    parse_smartctl_info,
+    parse_smartctl_attributes,
+    main,
+)
+
+class TestSmartMon(unittest.TestCase):
+    @patch('smartmon.run_command')
+    def test_parse_smartctl_info(self, mock_run_command):
+        devices_info = [
```
dougszumski: It would be useful to factor this example data out into a dedicated file. Then we can have a file for each HW profile, and loop over them.

dougszumski (commented):
This is great start, many thanks. It would be good to investigate external libraries for the smart data collection / parsing.

etc/kayobe/ansible/scripts/smartmon.py
```diff
@@ -0,0 +1,213 @@
...
+def run_command(command, parse_json=False):
+    """
+    Helper to run a subprocess command and optionally parse JSON output.
+    """
+    result = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+    if parse_json:
+        return json.loads(result.stdout)
+    return result.stdout.strip()
+
+def parse_device_info(device):
```
dougszumski, line 91: It's worth considering a type hint to describe that the expected `type` of `device` is a pySMART device, or at least describing it in the doc string.

etc/kayobe/ansible/scripts/smartmon.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Inspect all public attributes on device.if_attributes
+    for attr_name in dir(device.if_attributes):
+        if attr_name.startswith("_"):
+            continue  # skip private / special methods
+        val = getattr(device.if_attributes, attr_name, None)
+        if callable(val):
+            continue  # skip methods
+
+        # Convert CamelCase or PascalCase -> snake_case, e.g. dataUnitsRead -> data_units_read
```
dougszumski: This deserves its own method. Does it also need a reference? 
https://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case

etc/kayobe/ansible/scripts/test_smartmon.py
```diff
@@ -0,0 +1,262 @@
...
+            device.if_attributes = None
+
+        return device
+
+    def test_parse_device_info(self):
+        """
+        Test parse_device_info() for every JSON fixture in ./drives/.
+        We do subTest() so each fixture is tested individually.
+        """
+        for fixture_path in self.fixture_files:
```
dougszumski, line 140: If we have lots of fixture files and the test fails, we will get `generic test_parse_device_info [failed]`, and then have to go and figure out which device(s) weren't parsed correctly. This is ok at the moment, but if we have lots of test files it could be come unwieldy. 

Generally it's best to keep tests very simple and targeted, so that if one fails, it gives you a strong pointer about where the issue is. 

A way around this would be to factor out the test logic into a `_test_parse_device_info(self, fixture_file)` method and then have a test per device, with each test just calling  `_test_parse_device_info`.

etc/kayobe/ansible/scripts/drives/nvme.json
```diff
@@ -0,0 +1,24 @@
+{
+  "device_info": {
+    "name": "/dev/nvme0",
```
dougszumski: Perhaps rename the subfolder `drives` to `tests`? And then later on (not in this change) we can move the script to a dedicated repo.

etc/kayobe/ansible/scripts/test_smartmon.py
```diff
@@ -0,0 +1,262 @@
...
+class TestSmartMon(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # Collect all *.json files from ./drives/
+        data_folder = os.path.join(os.path.dirname(__file__), "drives")
+        cls.fixture_files = glob.glob(os.path.join(data_folder, "*.json"))
+
+    def create_mock_device_from_json(self, device_info, if_attributes=None):
+        """
+        Given a 'device_info' dict and optional 'if_attributes', build
```
dougszumski, line 37: Can you create an actual pySmart device object here from the file output?
Eg. https://github.com/truenas/py-SMART/blob/master/tests/test_device.py#L43

etc/kayobe/ansible/scripts/test_smartmon.py
```diff
@@ -0,0 +1,262 @@
...
+                        smart_healthy_found,
+                        f"Expected device_smart_healthy={expected_val} for {dev_name}, not found."
+                    )
+
+    def test_parse_if_attributes(self):
+        """
+        Test parse_if_attributes() for every JSON fixture in ./drives/.
+        We do subTest() so each fixture is tested individually.
+        """
+        for fixture_path in self.fixture_files:
```
dougszumski, line 200: Similar here - let's swap the loop for dedicated tests

etc/kayobe/ansible/scripts/test_smartmon.py
```diff
@@ -0,0 +1,262 @@
...
+
+                dev_name = device_info["name"]
+                dev_iface = device_info["interface"]
+                dev_serial = device_info["serial"].lower()
+
+                # For each numeric attribute in JSON, if it's in SMARTMON_ATTRS,
+                # we expect a line in the script's output.
+                for attr_key, attr_val in if_attrs.items():
+                    # Convert from e.g. "criticalWarning" -> "critical_warning"
+                    snake_key = re.sub(r'(?<!^)(?=[A-Z])', '_', attr_key).lower()
```
dougszumski: I'd consider this too complex for a unit test. It would be better to hard code the expected outcomes for a given input.

dougszumski (commented):
Thanks @technowhizz , good effort.

etc/kayobe/ansible/scripts/test_smartmon.py
technowhizz: So I gave this a shot but it turns out that it also requires a mock for smartctl and this made it a lot more complicated to just use a pySmart device. I think it makes more sense to mock it here

etc/kayobe/ansible/scripts/test_smartmon.py
technowhizz: Should we do this as a part 2? I think it would be a lot of work to hard code it

etc/kayobe/ansible/scripts/smartmon.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+    Convert a list of lines like "some_metric{...} value"
+    into a Prometheus text output with # HELP / # TYPE lines.
+    """
+    output = []
+    last_metric = ""
+    for metric in sorted(metrics):
+        metric_name = metric.split("{")[0]
+        if metric_name != last_metric:
+            output.append(f"# HELP smartmon_{metric_name} SMART metric {metric_name}")
```
dougszumski: Please can you switch to the Prometheus library here? It is a slight change that it will write directly to a file for scraping. The advantage is that it handles all the metric formatting. A rough example:

```
from prometheus_client import CollectorRegistry, Gauge, write_to_textfile

metric_registry = CollectorRegistry()
metric_output_path = os.environ['SMART_METRIC_OUTPUT_PATH'])

for metric in metrics:
   g = Gauge(metric_name, 'some help stirng', registry=registry)
   g.set(metric_value)

write_to_textfile(metric_output_path, registry)
```

https://prometheus.github.io/client_python/exporting/textfile/

etc/kayobe/ansible/scripts/test_smartmon.py
technowhizz: Will do as a part 2

dougszumski (approved):
Thanks @technowhizz . Let's merge and then iterate it on it.

PR #1328, opened 2024-10-11
Increase size of connection tracking table

priteau:
We have seen occurences of Cinder or Glance activity filling up the conntrack table. This has been seen on controllers and storage nodes (Ceph). We have also seen large amounts of traffic to an Octavia amphora causing the conntrack table of a compute host to fill up.

Quadruple the maximum size of the table on these hosts (the default size is 262,144 for hosts with more than 4 GiB of memory [1]).

Monitoring hosts are left unchanged since they are believed to be unaffected.

[1] https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt


priteau:
The conntrack table can only be increased if the `nf_conntrack` module is loaded, which is the case only when the firewall is activated.

MoteHue:
> The conntrack table can only be increased if the `nf_conntrack` module is loaded, which is the case only when the firewall is activated.

In the situation where we've enabled the firewall then chosen to disable it, is it safe to assume nf_conntrack will have also been disabled/unloaded?

priteau:
> > The conntrack table can only be increased if the `nf_conntrack` module is loaded, which is the case only when the firewall is activated.
> 
> In the situation where we've enabled the firewall then chosen to disable it, is it safe to assume nf_conntrack will have also been disabled/unloaded?

I am not sure, this needs to be tested.

Alex-Welsh:
Is this still an issue @priteau?

PR #1333, opened 2024-10-16
trivy: Add more repositories to handle rate limits - 2024.1

bbezak:
More repositories are configured to avoid issues when ghcr.io hits rate limits. If the primary repository fails due to rate limiting, the system will fall back to the public ECR repository [1]

[1] https://github.com/aquasecurity/trivy/discussions/7668#discussioncomment-10892093

### DIFF
--- a/tools/scan-images.sh
+++ b/tools/scan-images.sh
@@ -54,6 +54,10 @@ for image in $images; do
           --severity HIGH,CRITICAL \
           --output image-scan-output/${filename}.json \
           --ignore-unfixed \
+          --db-repository ghcr.io/aquasecurity/trivy-db:2 \
+          --db-repository public.ecr.aws/aquasecurity/trivy-db \
+          --java-db-repository ghcr.io/aquasecurity/trivy-java-db:1 \
+          --java-db-repository public.ecr.aws/aquasecurity/trivy-java-db \
           $image); then
     # Clean up the output file for any images with no vulnerabilities
     rm -f image-scan-output/${filename}.json


tools/scan-images.sh
```diff
@@ -54,6 +54,10 @@ for image in $images; do
           --severity HIGH,CRITICAL \
           --output image-scan-output/${filename}.json \
           --ignore-unfixed \
+          --db-repository ghcr.io/aquasecurity/trivy-db:2 \
+          --db-repository public.ecr.aws/aquasecurity/trivy-db \
+          --java-db-repository ghcr.io/aquasecurity/trivy-java-db \
```
Alex-Welsh: should this be `ghcr.io/aquasecurity/trivy-java-db:1`?

https://github.com/aquasecurity/trivy/commit/3562529ddfb26d301311ed450c192e17011353df#diff-39850860d29b4340dd9348be6ff8891c761b0bf742fb978480d5f6763c9142abR108

bbezak: good catch, thx

Alex-Welsh:
LGTM but we should merge this into Antelope first then up to Caracal

bbezak:
> LGTM but we should merge this into Antelope first then up to Caracal

willdo. I'm not sure why I don't like "frontporting" approach :)

Alex-Welsh:
> > LGTM but we should merge this into Antelope first then up to Caracal
> 
> willdo. I'm not sure why I don't like "frontporting" approach :)

I don't either. I think we should switch to backporting from our latest stable branch

PR #1335, opened 2024-10-16
Fix upgrade-prerequisites

Alex-Welsh:
The KAYOBE_CONFIG_ROOT variable is out of scope so is undefined:
```/home/ubuntu/src/kayobe-config/tools/upgrade-prerequisites.sh: line 38: /tools/rabbitmq-quorum-migration.sh: No such file or directory```

Even though the script is being called using that very same variable: https://github.com/stackhpc/terraform-kayobe-multinode/blob/main/ansible/files/multinode.sh#L374

### DIFF
--- a/tools/rabbitmq-quorum-migration.sh
+++ b/tools/rabbitmq-quorum-migration.sh
@@ -13,7 +13,7 @@ fi
 
 if [[ ! "$1" = "--skip-checks" ]]; then
     # Fail if clocks are not synced
-    if ! kayobe overcloud host command run -l controllers -b --command "timedatectl status | grep 'synchronized: yes'"; then
+    if ! ( kayobe overcloud host command run -l controllers -b --command "timedatectl status | grep 'synchronized: yes'" ); then
         echo "Failed precheck: Time not synced on controllers"
         echo "Use 'timedatectl status' to check sync state"
         echo "Either wait for sync or use 'chronyc makestep'"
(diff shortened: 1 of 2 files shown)


Alex-Welsh:
Converted to draft - untested

seunghun1ee:
I had to make this change on customer system upgrade. So, this is tested. However we didn't considered the case when the system doesn't use Kayobe environment. I'll add a commit for this.

Alex-Welsh:
> I had to make this change on customer system upgrade. So, this is tested. However we didn't considered the case when the system doesn't use Kayobe environment. I'll add a commit for this.

We discussed this in person but for future reference, this script is designed to be run in CI only. There are too many edge cases on customer deployments to cover

Alex-Welsh:
Tested and working here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/11662590141/job/32469246299

There's a weird order of operations here:
This PR needs to merge
Then this one in the github workflows: https://github.com/stackhpc/stackhpc-openstack-gh-workflows/pull/8
Then I need to make a new release of that repo
Then I need to make a new PR to SKC `stackhpc/2024.1` which will bump the version it uses to that release

I've been testing this change with some experimental branches:
https://github.com/stackhpc/stackhpc-kayobe-config/compare/stackhpc/2024.1...caracal-multinode
https://github.com/stackhpc/stackhpc-openstack-gh-workflows/compare/main...alex-test-branch

MoteHue (approved):

PR #1343, opened 2024-10-24
INFRA-804: document GRUB boot issue on RL9

priteau:
INFRA-804: document GRUB boot issue on RL9 and add hook to automate removal of ``--root-dev-only``.


MoteHue (commented):
Should we also mention this in the host package updates sections?
It's a nasty one to miss

priteau:
> Should we also mention this in the host package updates sections? It's a nasty one to miss

I updated Dawud's patch to include a note in the host package update section.

MoteHue (approved):

PR #1345, opened 2024-10-25
fix: test if `admin-openrc.sh` exists before deploying os-capacity

jackhodgkiss:
During service deploys using Kayobe Automation this playbook will fail as the `admin-openrc.sh` file is not generated during the deployment process.

This in turn causes the workflow to be reported as a failure even though the service deployment succeeded.

### DIFF
--- a/etc/kayobe/ansible/deploy-os-capacity-exporter.yml
+++ b/etc/kayobe/ansible/deploy-os-capacity-exporter.yml
@@ -15,59 +15,61 @@
   tags: os_capacity
   gather_facts: false
   tasks:
-    - name: Create os-capacity directory
-      ansible.builtin.file:
-        path: /opt/kayobe/os-capacity/
-        state: directory
-      when: stackhpc_enable_os_capacity
-
-    - name: Read admin-openrc credential file
-      ansible.builtin.command:
-        cmd: "cat {{ lookup('ansible.builtin.env', 'KOLLA_CONFIG_PATH') }}/admin-openrc.sh"
+    - name: Check if admin-openrc.sh exists
+      ansible.builtin.stat:
+        path: "{{ lookup('ansible.builtin.env', 'KOLLA_CONFIG_PATH') }}/admin-openrc.sh"
       delegate_to: localhost
-      register: credential
-      when: stackhpc_enable_os_capacity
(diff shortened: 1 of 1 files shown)


assumptionsandg (dismissed):
LGTM, Nice catch!

etc/kayobe/ansible/deploy-os-capacity-exporter.yml
```diff
@@ -15,59 +15,61 @@
...
+              OS_CLIENT_CONFIG_FILE: /etc/openstack/clouds.yaml
+            mounts:
+              - type: bind
+                source: /opt/kayobe/os-capacity/
+                target: /etc/openstack/
+            network_mode: host
+            restart: "{{ clouds_yaml_result is changed or cacert_result is changed }}"
+            restart_policy: unless-stopped
+          become: true
+      when: stackhpc_enable_os_capacity and openrc_file_stat.stat.exists
```
holliefae: Think the extra line needs to be removed to make the linter happy

jackhodgkiss: Thanks. Line has been removed.

assumptionsandg (dismissed):

etc/kayobe/ansible/deploy-os-capacity-exporter.yml
```diff
@@ -15,59 +15,60 @@
...
-    - name: Create os-capacity directory
-      ansible.builtin.file:
-        path: /opt/kayobe/os-capacity/
-        state: directory
-      when: stackhpc_enable_os_capacity
-
-    - name: Read admin-openrc credential file
-      ansible.builtin.command:
-        cmd: "cat {{ lookup('ansible.builtin.env', 'KOLLA_CONFIG_PATH') }}/admin-openrc.sh"
+    - name: Check if admin-openrc.sh exists
```
Alex-Welsh, line 18: Could also do
```yaml
run_once: true
```
So it doesn't do it three times

jackhodgkiss: Done.

Alex-Welsh (approved):
lgtm

PR #1347, opened 2024-10-28
CI: Bump AIO root volume size to 50GB

seunghun1ee:
Tempest test on https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/11516474614/job/32145413990?pr=1312 is keep failing because the aio is running out of space. Bumping it to 50 GB.


jovial (approved):
We've noticed CI failing when the storage fills up. This will give us a bit more head room.

PR #1352, opened 2024-10-30
Caracal upgrade: document Octavia bugfix

MoteHue:


Alex-Welsh (approved):
nit: should probably use proper links here, and maybe also mention that KA will complain if you have local changes

seunghun1ee (approved):

MoteHue:
Fix is merging now: https://review.opendev.org/c/openstack/kolla-ansible/+/932408
Let's get this synced into our KA fork instead

PR #1354, opened 2024-10-30
Use python3 -m venv for advise venv create

MoteHue:


Alex-Welsh (approved):

Alex-Welsh:
Technically we should merge this into Antelope first then sync up but idk if it's that important

MoteHue:
> Technically we should merge this into Antelope first then sync up but idk if it's that important

All our new deployments will be on Caracal, so we don't need it in Antelope anymore :)

PR #1358, opened 2024-11-01
Redfish exporter: Decrease sensitivity of alert

jovial:
The redfish exporter is prone to failed scrapes. Lets wait for mulitple failed scrapes before triggering an alert. This should still catch the case where it is completely dead, but reduce the false positives from failed scrapes.

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/prometheus.rules
+++ b/etc/kayobe/kolla/config/prometheus/prometheus.rules
@@ -7,8 +7,17 @@ groups:
   rules:
 
   - alert: PrometheusTargetMissing
-    expr: up == 0
+    expr: up{job!="redfish-exporter-seed"} == 0
     for: 5m
+    labels:
+      severity: critical
+    annotations:
+      summary: "Prometheus target missing (instance {{ $labels.instance }})"
+      description: "A Prometheus target has disappeared. An exporter might have crashed."
+
+  - alert: PrometheusTargetMissing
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla/config/prometheus/prometheus.rules
```diff
@@ -7,14 +7,23 @@ groups:
...
 
+  - alert: PrometheusTargetMissing
+    expr: up{job="redfish-exporter-seed"} == 0
+    for: 15m
+    labels:
+      severity: critical
+    annotations:
+      summary: "Prometheus target missing (instance {{ $labels.instance }})"
+      description: "A Prometheus target has disappeared. An exporter might have crashed."
+
```
Alex-Welsh, line 26: Duplicate of the rule above?

JohnGarbutt: Its more splitting it into two alerts, so the redfish one waits for 15m before alerting, I think?

JohnGarbutt (approved):

etc/kayobe/kolla/config/prometheus/prometheus.rules
Alex-Welsh: Ah, I hadn't noticed the expression was different. Makes sense now

Alex-Welsh (approved):

PR #1360, opened 2024-11-04
Add scripts to simplify Tempest testing

Alex-Welsh:
Added two scripts:

`run-tempest.sh` makes it easy to run Tempest tests
`tempest-cleanup.sh` does a reasonable job of clearing up openstack resources that have been left behind by Tempest failures. It is not perfect but something is better than nothing and we can iterate on it over time.

They've been hanging around on a few client deployments I've used so I'm adding them here for all to enjoy.

### DIFF
--- a/.automation.conf/run-tempest.sh
+++ b/.automation.conf/run-tempest.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# Script based on Tempest section of Kayobe multinode deployment script
+# plus https://wiki.stackhpc.com/doc/using-kayobe-automation-to-run-tempest-7yAEJw2eHb
+
+set -euo pipefail
+
+KAYOBE_CONFIG_BASE_PATH=$KAYOBE_CONFIG_PATH/../../
+
+# Set Tempest env vars if not already set
+set +x
+: "${KAYOBE_AUTOMATION_SSH_PRIVATE_KEY:=$(cat ~/.ssh/id_rsa)}"
+: "${TEMPEST_OPENRC:=$(cat $KAYOBE_CONFIG_PATH/../kolla/public-openrc.sh)}"
+set -x
+
+# Ensure timestamped output dir exists
+OUTPUT_DIR=$KAYOBE_CONFIG_BASE_PATH/tempest-artifacts/$KAYOBE_ENVIRONMENT/$(date +%Y-%m-%d--T%H-%M-%S)
+echo Creating output directory: $OUTPUT_DIR
+mkdir -p $OUTPUT_DIR
+
+
+# Set base image for Kayobe container. Use Rocky 9 for zed+ CentOS otherwise
+if grep -Eq "(202|zed)" $KAYOBE_CONFIG_BASE_PATH/.gitreview; then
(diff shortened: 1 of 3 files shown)


.automation.conf/tempest-cleanup.sh
jackhodgkiss, line 1: I think it could be worth exploring https://docs.openstack.org/tempest/latest/cleanup.html in combination with `--resource-list` to automatically clean up resources.

.automation.conf/run-tempest.sh
```diff
@@ -0,0 +1,57 @@
...
+# Script based on tempest section of Kayobe multinode deployment script
+# plus https://wiki.stackhpc.com/doc/using-kayobe-automation-to-run-tempest-7yAEJw2eHb
+
+set -euxo pipefail
+
+KAYOBE_CONFIG_BASE_PATH=$KAYOBE_CONFIG_PATH/../../
+
+set +x
+export KAYOBE_AUTOMATION_SSH_PRIVATE_KEY=$(cat ~/.ssh/id_rsa)
+export TEMPEST_OPENRC="$(< $KAYOBE_CONFIG_PATH/../kolla/admin-openrc.sh)"
```
jackhodgkiss: This would be problematic for environments that don't use the default `id_rsa` key.

.automation.conf/run-tempest.sh
jackhodgkiss, line 1: Is this only intended for making multinodes easier to test? Or for end-users? I think ideally we would use Kayobe Automation integrated into either GitHub or Gitlab.

If this is only for multinodes it could be placed in that environment.

.automation.conf/run-tempest.sh
seunghun1ee: We can try setting this conditionally. If environment variable is already defined, use the value of it. Otherwise, use these default values.
But don't we usually use ``public-openrc.sh`` as a default?

.automation.conf/run-tempest.sh
Alex-Welsh: Yes, it should be usable in most production environments. I agree that kayobe automation is superior, but it just isn't available everywhere

.automation.conf/tempest-cleanup.sh
Alex-Welsh: Interesting, I'd not seen that before. It looks useful, especially with the `--prefix` option, which could mirror the operation of the script. That being said, there is a big red warning at the top of the page saying **Do not use in production**

.automation.conf/run-tempest.sh
Alex-Welsh: Switched to the public openrc and changed it so it won't override the variables if they're already defined. That's as good as we can get I think without converting it all to a playbook or something

.automation.conf/tempest-cleanup.sh
Alex-Welsh: I think for now this script is a good first step, and we should see if we can replace it in the future

Alex-Welsh:
Found a couple of issues in testing, converting to draft for now

Alex-Welsh:
Not worth the effort for now

PR #1361, opened 2024-11-04

jackhodgkiss:

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -202,21 +202,12 @@ jobs:
       - name: Write Terraform network config
         run: |
           cat << EOF > etc/kayobe/environments/$KAYOBE_ENVIRONMENT/tf-networks.yml
-
-          admin_oc_net_name: admin
-          admin_cidr: "{{ access_cidr.value }}"
-          admin_allocation_pool_start: 0.0.0.0
-          admin_allocation_pool_end: 0.0.0.0
-          admin_gateway: "{{ access_gw.value }}"
-          admin_bootproto: dhcp
-          admin_ips:
+          admin_oc_net_name: ethernet
+          ethernet_cidr: "{{ access_cidr.value }}"
+          ethernet_allocation_pool_start: 0.0.0.0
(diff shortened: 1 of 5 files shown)


jovial:
Guess the conflict comes from: https://github.com/stackhpc/stackhpc-kayobe-config/blob/95182a6b3c700041c89b58ba2414360ec17c1790/.github/workflows/stackhpc-all-in-one.yml#L229-L233

Alex-Welsh:
I like the approach but one comment, can we keep the original `admin` name, or maybe switch to `admin_oc`? It's a bit clearer than `ethernet` imo

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -202,23 +202,14 @@ jobs:
       - name: Write Terraform network config
         run: |
           cat << EOF > etc/kayobe/environments/$KAYOBE_ENVIRONMENT/tf-networks.yml
-
-          admin_oc_net_name: admin
```
jovial, line 206: Now that I've seen this, I'm wondering why adding:

```
admin_zone: trusted
```

and https://github.com/stackhpc/stackhpc-kayobe-config/pull/1361/files#diff-bc3d997969357b16390ebf2a9aa7c65f86030f4eb5fc5323194c095e5d483c46L75

isn't enough?

jackhodgkiss:
> I like the approach but one comment, can we keep the original `admin` name, or maybe switch to `admin_oc`? It's a bit clearer than `ethernet` imo

I prefer ethernet because it the port plugged into the VM or instance. It is only called admin_oc here because of how AIO within the context of CI is configured.

.github/workflows/stackhpc-all-in-one.yml
jovial: @jackhodgkiss Explained that this works in CI but not running https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/yoga/etc/kayobe/environments/ci-aio/automated-setup.sh. This is essentially making CI use the same mechanism.

jovial (approved):

PR #1362, opened 2024-11-05
Update octavia.rst

MaxBed4d:
Update to the Octavia docs to provide some details on the deployment.


doc/source/operations/octavia.rst
```diff
@@ -2,6 +2,45 @@
...
+``${KAYOBE_CONFIG_PATH}/inventory/group_vars``) to see the network which Octavia management
+network will be using. If it is configured to use a VLAN then ``kolla_enable_neutron_provider_networks: true``
+should also be set in ``kolla.yml``.
+
+.. note::
+
+      It of course goes without saying that the network configured for ``octavia_net_interface``
+      should also exist in ``networks.yml``.
+
+By default Octavia will deploy an Amphorae (a single Amphora VM) per load balancing service, this should be
```
MoteHue: Amphorae is the plural of Amphora.
```suggestion
By default Octavia will deploy an Amphora (a single Ubuntu VM running HAProxy) per load balancing service. This should be
```

MoteHue (changes requested):

doc/source/operations/octavia.rst
```diff
@@ -2,6 +2,45 @@
...
+Octavia is known as an *operator-scaled* load balancing service, whose primary goal is to
+distribute incoming network traffic, destined for a single VM, across multiple VMs instead;
+thus preventing overloading the VM and improving the overall performance of the VM's outbound
+service.
+
+For more information on Octavia, please refer to the `official documentation <https://docs.openstack.org/octavia/latest/reference/introduction.html>`_.
+
+.. _Deploying Octavia:
+
+Deploying Octavia
```
Alex-Welsh, line 19: Could you restructure this into a sequence of individual steps? See for example the [Multinode page](https://stackhpc-kayobe-config.readthedocs.io/en/stackhpc-2023.1/contributor/environments/ci-multinode.html#multinode-test-environment) which goes through a few things step-by-step

doc/source/operations/octavia.rst
```diff
@@ -2,6 +2,45 @@
...
+.. note::
+
+      It of course goes without saying that the network configured for ``octavia_net_interface``
+      should also exist in ``networks.yml``.
+
+By default Octavia will deploy an Amphorae (a single Amphora VM) per load balancing service, this should be
+changed to make them highly available and have two Amphora VM per service. Done by setting
+``octavia_loadbalancer_topology: "ACTIVE_STANDBY"`` in ``${KAYOBE_CONFIG_PATH}/kolla/globals.yml``,
+this will ensure that if the master Amphora VM were to go down, the other would be able to
+take over the load balancing functions.
```
Alex-Welsh: Only applies when using the Amphora driver

doc/source/operations/octavia.rst
```diff
@@ -2,6 +2,45 @@
...
+
+Further configuration options and details on installation can be found in the
+`Octavia documentation <https://docs.openstack.org/octavia/latest/install/index.html>`_.
```
Alex-Welsh: The kolla-ansible documentation will be more relevant here than the raw Octavia documentation
```suggestion
`Octavia documentation <https://docs.openstack.org/kolla-ansible/latest/reference/networking/octavia.html>`_.
```

Alex-Welsh (changes requested):

doc/source/operations/octavia.rst
MaxBed4d: It is, indeed, don't know why I changed it originally.

doc/source/operations/octavia.rst
MaxBed4d: Agreed

doc/source/operations/octavia.rst
MaxBed4d: Made this clearer.

doc/source/operations/octavia.rst
MaxBed4d: This seems a little unnecessary for this because it would end up being 2 steps; this is more to make people aware of the various configs which could snag the deployment.

Alex-Welsh: I think at the moment it misses a few important steps, in particular "It of course goes without saying that the network should exist in networks.yml" is a pretty big thing to gloss over.

So for example you could have:
1. Where to enable the service in kolla.yml
2. How to define the octavia network in networks.yml
3. How to define network interfaces for controllers
4. What to check for neutron provider networks
5. How to select the right driver(s), and the differences between Amphora vs OVN
6. How to properly configure the Amphora driver

Alex-Welsh: Please also keep in mind that simple, clear instructions make for the best docs. try to stay away from _wordyness_.

For example:

> Much like any other Kolla managed service, the method of deploying Octavia is as simple
as ...

could be written as:

> To enable Octavia, set ...

MaxBed4d: I have an idea on how to make a nice comprehensive list. Regarding the *wordyness*, in some cases it is intentional, in the example provided mentioning its similarity to other Kolla managed services will give users a lot of context as to what to expect, as long as they have deployed at least one service before; in addition to reinforcing the steps for deploying services. In other cases, I agree and some sentences are wordy without adding anything.

doc/source/operations/octavia.rst
```diff
@@ -2,6 +2,74 @@
...
+
+   - USING VLAN  
+
+     4.1. Set ``kolla_enable_neutron_provider_networks: true`` in ``kolla.yml``.  
+   
+   - USING AMPHORA  
+  
+     4.1. Set ``octavia_loadbalancer_topology: "ACTIVE_STANDBY"`` in ``${KAYOBE_CONFIG_PATH}/kolla/globals.yml``.
+
+#. Run ``kayobe overcloud service configure``.
```
MoteHue:
```suggestion
#. Run ``kayobe overcloud service reconfigure``.
```

MoteHue (changes requested):

PR #1363, opened 2024-11-05

jackhodgkiss:


Alex-Welsh:
Need this on `stackhpc/zed` first, no SLURP till Caracal

jackhodgkiss:
> Need this on `stackhpc/zed` first, no SLURP till Caracal

https://github.com/stackhpc/stackhpc-kayobe-config/pull/1364

Alex-Welsh (dismissed):
LGTM as long as CI passes once https://github.com/stackhpc/stackhpc-kayobe-config/pull/1364 merges

Alex-Welsh (approved):

PR #1365, opened 2024-11-05
Pin os-capacity to v0.5 release

JohnGarbutt:
Version v0.3 was broken an automatically pulled
on a fresh controller. Lets start pinning this version.

Note v0.5 fixes 842 CVEs vs the previous version:
https://github.com/stackhpc/os-capacity/security/code-scanning

### DIFF
--- a/etc/kayobe/ansible/deploy-os-capacity-exporter.yml
+++ b/etc/kayobe/ansible/deploy-os-capacity-exporter.yml
@@ -60,7 +60,7 @@
     - name: Ensure os_capacity container is running
       community.docker.docker_container:
         name: os_capacity
-        image: ghcr.io/stackhpc/os-capacity:v0.5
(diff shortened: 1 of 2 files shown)


Alex-Welsh (commented):
Since it's arguably a bugfix, could we aim this at `stackhpc/2023.1` first? We'll then merge it up from there.

I'd also prefer it if we could slurp this out to a variable and set it somewhere like `stackhpc-monitoring.yml` so it's a bit easier to find

Both of those suggestions are easy to do retroactively so if this is a more urgent fix I'm happy to merge now and patch later

Alex-Welsh (approved):
Looks good. For future reference, I would normally say we need to have release notes when we change pins but in this case, we're pinning it to the release it's already on so nothing functionally changes.

PR #1366, opened 2024-11-06
CIS: Stop recursively setting permissions on logs files

jovial:
This suffers from a time-of-check to time-of-use race condition and in general recursively setting file permissions seems to be a bad idea.


Alex-Welsh:
@jovial you might want to update your local `stackhpc/2023.1` checkout

jackhodgkiss (approved):

MoteHue (approved):

PR #1369, opened 2024-11-08
Allow admins to list all baremetal nodes

MoteHue:


etc/kayobe/kolla/config/ironic/policy.yaml
```diff
@@ -0,0 +1,6 @@
+# Retrieve multiple Node records
+# GET  /nodes
+# GET  /nodes/detail
+# Intended scope(s): system, project
+# Overridden: added role:admin
+"baremetal:node:list_all": "role:admin or (role:reader and system_scope:all) or (role:service and system_scope:all) or rule:service_role"
```
jovial, line 6: I would say this is probably undesirable on public clouds as you probably don't want users knowing how many nodes you have and what state they are in. There is the caveat that we can't really grant users project admin at the moment since this also grants them a bunch of cloud admin based rights (not all services support system scope), but just thinking toward the future. Could this be solved by using a non-standard role, such that it is opt in e.g  the baremetal_node_list role that you grant to specific users?

MoteHue: That sounds like a decent idea, I certainly like future-proofing here. My motivation for this patch was really to make sure we don't break existing Ironic use-cases during an upgrade to Caracal. If we automatically create this new role and make it clear in our upgrade docs it needs to be assigned to the appropriate users, then that fits the bill imo.

jovial: Would be good to hear what @JohnGarbutt thinks

jovial: Thought about this a little bit more, and as system scope admin is not fully implemented, I think it makes sense to go with what you currently have as we do not really know how things will play out. We should avoid introducing roles that could potentially break in the future.

MoteHue: Makes sense, I'd be happy to merge this as-is then. (Also, it's not really causing any _new_ concerns with our public-cloud customers as they already have this behaviour on 2023.1 anyway. Will be nice to eventually tighten this, as you say once the features stabilise.)

Alex-Welsh (approved):

PR #1375, opened 2024-11-14
Update docs for client.cinder to support COW snapshots

jovial:
This is to support copy on write snapshots:

```
Performing standard snapshot because direct snapshot failed: no write permission on storage pool images: nova.exception.Forbidden: no write permission on storage pool images
```

when using ceph for nova ephemeral storage. My preferrence is for a standardised configuration rather than another if you use this feature, do this.

### DIFF
--- a/releasenotes/notes/adjust-ceph-permissions-on-images-for-cinder-user-3154b995916ee27e.yaml
+++ b/releasenotes/notes/adjust-ceph-permissions-on-images-for-cinder-user-3154b995916ee27e.yaml
@@ -1,0 +1,11 @@
+---
+features:
+  - |
+    Adds a ``stackhpc_enable_ceph_cow_optimisations`` feature flag to enable
+    copy on write optimisations when using Ceph. Please see the ``Copy on write
+    optimisations`` section under `Configuraton Guide` > `Cephadm and Kayobe`
+    in the documentation.
+
+    The feature is currently opt-in. Note the documented permissions for the
+    images pool for the Cinder user have been adjusted to make this easier to
+    apply for future deployments.
--- a/doc/source/configuration/cephadm.rst
+++ b/doc/source/configuration/cephadm.rst
@@ -259,6 +259,25 @@ for Cinder, Cinder backup, Glance, and Nova in Kolla Ansible.
          mgr: "profile rbd pool=images"
        state: present
 
+.. note::
+
+   By default, the ``client.cinder`` user is configured with read-only access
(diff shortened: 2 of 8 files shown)


releasenotes/notes/adjust-ceph-permissions-on-images-for-cinder-user-3154b995916ee27e.yaml
```diff
@@ -0,0 +1,10 @@
+---
+fixes:
+  - |
+    Adjusts the documented permissions for the images pool for the cinder user.
+    This is to support copy on write snaphots of vms using ceph backed
+    ephemeral storage. Existing deployments will need to manually adjust these
+    values using the CLI (the collection is currently not smart enough to do this):
+    ``ceph auth caps client.cinder mon 'profile rbd' osd 'profile rbd
+    pool=volumes, profile rbd pool=vms, profile rbd pool=images' mgr 'profile
+    rbd pool=volumes, profile rbd pool=vms'``.
```
Alex-Welsh: A few very minor nits, otherwise LGTM but someone with more Ceph knowledge than me should review before we merge
```suggestion
    Adjusts the documented permissions for the images pool for the Cinder user.
    This is to support copy on write snapshots of VMs using Ceph backed
    ephemeral storage. Existing deployments will need to manually adjust these
    values using the CLI (the collection is currently not smart enough to do this):
    ``ceph auth caps client.cinder mon 'profile rbd' osd 'profile rbd
    pool=volumes, profile rbd pool=vms, profile rbd pool=images' mgr 'profile
    rbd pool=volumes, profile rbd pool=vms'``.
```

jovial:
Thanks Alex - Hoping @MoteHue might be able to test this (particularly the command in the release note). Make sure you grab a copy of the current capabilities before changing them.

MoteHue (changes requested):
> Thanks Alex - Hoping @MoteHue might be able to test this (particularly the command in the release note). Make sure you grab a copy of the current capabilities before changing them.

This command does work, but we should also bring to attention two things:

1. You need to update the ceph keyrings under cinder and nova configs to remove the read-only flag too.
2. This will break snapshots of boot-from-image instances unless the glance option `show_multiple_locations` is enabled.

mnasiadka:
Should we post an update to the Ceph docs at https://docs.ceph.com/en/latest/rbd/rbd-openstack/#setup-ceph-client-authentication ?

cityofships (approved):
LGTM

MoteHue (approved):

PR #1377, opened 2024-11-14
Add playbook for updating Vault API TLS certificate

seunghun1ee:
Unlike ``vault-deploy-seed.yml`` playbook, this playbook only updates ``overcloud.crt`` and ``overcloud.key`` which are used as Vault API TLS certificate.


seunghun1ee:
But are ``overcloud.crt`` and ``overcloud.key`` only used as Vault API TLS certificate? (Or ``{{ kolla_internal_fqdn }}.crt`` and ``{{ kolla_internal_fqdn }}.key`` ?)

Alex-Welsh:
This looks pretty handy @seunghun1ee

What do we need to do to push it forward?

seunghun1ee:
I tested this playbook but needs some fix. The draft PR was made because I wasn't sure about my approach.
I'll continue when I'm available to work on this.

PR #1382, opened 2024-11-22
Amending paths to match Ark

jjo93sa:
- Treasuredata needs to reference v5 not v4 and the Ark path includes `lts`
- RabbitMQ path should reference `noarch` rather than `x86_64`

Found during the upgrade of Graphcore CL0 from Antelope to Caracal

### DIFF
--- a/etc/kayobe/pulp.yml
+++ b/etc/kayobe/pulp.yml
@@ -42,8 +42,8 @@ stackhpc_release_pulp_domain: "ark.stackhpc.com"
 stackhpc_release_pulp_url: "{{ stackhpc_release_pulp_scheme }}://{{ stackhpc_release_pulp_domain }}"
 
 # Credentials used to access the StackHPC Pulp service.
-stackhpc_release_pulp_username: graphcore
-stackhpc_release_pulp_password: "{{ secrets_stackhpc_release_pulp_password }}"
+stackhpc_release_pulp_username:
+stackhpc_release_pulp_password:
 
 # Content URL of the StackHPC Pulp service.
 stackhpc_release_pulp_content_url: "{{ stackhpc_release_pulp_url }}/pulp/content"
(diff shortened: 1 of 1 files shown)


etc/kayobe/pulp.yml
```diff
@@ -42,8 +42,8 @@ stackhpc_release_pulp_domain: "ark.stackhpc.com"
 stackhpc_release_pulp_url: "{{ stackhpc_release_pulp_scheme }}://{{ stackhpc_release_pulp_domain }}"
 
 # Credentials used to access the StackHPC Pulp service.
-stackhpc_release_pulp_username:
-stackhpc_release_pulp_password:
+stackhpc_release_pulp_username: graphcore
```
m-bull: Please could you remove this change?

etc/kayobe/pulp.yml
```diff
@@ -42,8 +42,8 @@ stackhpc_release_pulp_domain: "ark.stackhpc.com"
...
+stackhpc_release_pulp_password: "{{ secrets_stackhpc_release_pulp_password }}"
```
m-bull: and this one

etc/kayobe/pulp.yml
```diff
@@ -56,7 +56,7 @@ stackhpc_release_pulp_content_url_with_auth: "{{ stackhpc_release_pulp_scheme }}
 # Sync all repositories required for building Kolla images from the
 # StackHPC Ark Pulp service to local Pulp.
 # NOTE: Only RPM repositories are supported.
-stackhpc_pulp_sync_for_local_container_build: false
+stackhpc_pulp_sync_for_local_container_build: true
```
m-bull: and this one!

m-bull (changes requested):
Thanks @jjo93sa - just a couple of downstream changes from your config that could do with a clean up please.

etc/kayobe/pulp.yml
jjo93sa: Done

etc/kayobe/pulp.yml
jjo93sa: And this one is done too

etc/kayobe/pulp.yml
jjo93sa: As is this one

m-bull (approved):

MoteHue:
Unfortunately it looks like our CI can't handle external contributors, we don't get very many of these. I've pushed to a new branch under my auth here: https://github.com/stackhpc/stackhpc-kayobe-config/pull/1426

Thanks for the fix @jjo93sa! We'll look into fixing external contributions in the future

PR #1386, opened 2024-11-26
Provide support for `ssh-sk` for `Nova`

jackhodgkiss:
Provide `ssh-sk` based keys for use within `Nova` providing users with the ability to further secure `SSH` access to their instances.

This has been achieved with by bumping the `cryptography` package to version `43.0.0` which includes https://github.com/pyca/cryptography/pull/10608.

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -346,6 +346,11 @@ kolla_build_blocks:
         && grafana-cli plugins install grafana-opensearch-datasource
   ironic_inspector_header: |
     ADD additions-archive /
+  nova_base_footer: |
+    {% raw %}
+    RUN {{ macros.upper_constraints_version_change('cryptography', '42.*', '43.0.3') }}
+    RUN {{ macros.upper_constraints_version_change('eventlet', '0.35.*', '0.36.1') }}
+    {% endraw %}
 
 # Dict mapping image customization variable names to their values.
 # Each variable takes the form:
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla.yml
```diff
@@ -346,6 +346,11 @@ kolla_build_blocks:
     {% set magnum_capi_packages = ['magnum-capi-helm==1.1.0'] %}
     RUN {{ macros.install_pip(magnum_capi_packages | customizable("pip_packages")) }}
     {% endraw %}
+  nova_base_footer: |
+    {% raw %}
+    {% set nova_packages = ['cryptography==43.0.0'] %}
```
mnasiadka: I assume cryptography is pulled in as a dependency to Nova? Just bump it up in upper-constraints using a macro?

jackhodgkiss: I wasn't sure what the best approach is as adjusting upper-constraints would impact more than just Nova.

mnasiadka: Well, we use u-c fork as a repo, but there are macros in Kolla that mangle u-c entries - you could use that only for Nova.

mnasiadka (dismissed):

jackhodgkiss:
@mnasiadka not sure why it is failing to build. I see `INFO:kolla.common.utils:Added next task BuildTask(nova-compute) to queue` and then nothing from `nova-compute` until the end when it says `"failed": [{"name": "nova-compute", "status": "building"}]`

PR #1387, opened 2024-11-27
Restore old UUID of the Redfish dashboard

technowhizz:
The redfish dashboard was updated to use a new UUID, however this caused errors in the Grafana logs and Grafana didn't update the dashboard. Reverting this to the old UUID fixes the issue.


dougszumski (approved):
Thanks @technowhizz

PR #1390, opened 2024-11-29
Generalise host image playbooks

assumptionsandg:
Generalise image upload playbook so it can be used with Octavia and IPA image uploads.


.github/workflows/overcloud-host-image-build.yml
Alex-Welsh, line 1: Still got a non-generalised name on this file

holliefae: Sorry I didn't mean to make changes in this workflow, only updating the upload playbook name.

holliefae: Just to clarify, we're only updating the upload playbook as it's applicable to other images, we will use new workflow files for the IPA and Amphora builds.

.github/workflows/overcloud-host-image-build.yml
```diff
@@ -239,7 +239,7 @@ jobs:
           source venvs/kayobe/bin/activate &&
           source src/kayobe-config/kayobe-env --environment ci-builder &&
           kayobe playbook run \
-          src/kayobe-config/etc/kayobe/ansible/pulp-host-image-upload.yml \
+          src/kayobe-config/etc/kayobe/ansible/pulp-image-upload.yml \
           -e image_path=/opt/kayobe/images/overcloud-rocky-9 \
           -e host_image_tag=${{ steps.host_image_tag.outputs.host_image_tag }} \
```
Alex-Welsh: There's a few invocations of `host_image_tag` in this workflow that do to be change, since the playbook now calls it `image_tag`

etc/kayobe/ansible/pulp-image-upload.yml
```diff
@@ -7,15 +7,16 @@
     remote_pulp_password: "{{ stackhpc_image_repository_password }}"
     repository_name: "kayobe-images-{{ openstack_release }}-{{ os_distribution }}-{{ os_release }}"
     pulp_base_path: "kayobe-images/{{ openstack_release }}/{{ os_distribution }}/{{ os_release }}"
+    file_extension: '*.qcow2'
```
Alex-Welsh: `file_extension` is somewhat misleading since it really a regex to search for. You could just leave the `*.` out of it then add that on inside the playbook?

etc/kayobe/ansible/pulp-image-upload.yml
```diff
@@ -7,15 +7,16 @@
     remote_pulp_password: "{{ stackhpc_image_repository_password }}"
     repository_name: "kayobe-images-{{ openstack_release }}-{{ os_distribution }}-{{ os_release }}"
     pulp_base_path: "kayobe-images/{{ openstack_release }}/{{ os_distribution }}/{{ os_release }}"
```
Alex-Welsh: In general, we may want to customise this path a bit more, I think it might need one more variable at the start so you can change `kayobe-images` to instead be something like `ipa-images` for example

Alex-Welsh (changes requested):

.github/workflows/overcloud-host-image-build.yml
holliefae: Think they're all updated now.

etc/kayobe/ansible/pulp-image-upload.yml
holliefae: Moved the regex out, now just defaulting to `qcow2`. IPA will pass in `initramfs` and `kernel`.

etc/kayobe/ansible/pulp-image-upload.yml
holliefae: Setting the image_name extra var to `kayobe-images` in the host image build. Set to `ipa` in the IPA workflow.

etc/kayobe/ansible/pulp-image-upload.yml
```diff
@@ -5,17 +5,18 @@
     remote_pulp_url: "{{ stackhpc_release_pulp_url }}"
     remote_pulp_username: "{{ stackhpc_image_repository_username }}"
     remote_pulp_password: "{{ stackhpc_image_repository_password }}"
-    repository_name: "kayobe-images-{{ openstack_release }}-{{ os_distribution }}-{{ os_release }}"
-    pulp_base_path: "kayobe-images/{{ openstack_release }}/{{ os_distribution }}/{{ os_release }}"
+    repository_name: "{{ image_name }}-{{ openstack_release }}-{{ os_distribution }}-{{ os_release }}"
+    pulp_base_path: "{{ image_name }}/{{ openstack_release }}/{{ os_distribution }}/{{ os_release }}"
```
Alex-Welsh: I think the term `image_name` is a bit misleading. Something like `image_path` or `image_directory` might make more sense

Alex-Welsh (commented):
Could we actually generalise this further? I think it would be quite easy to remove references to "image" and make the playbook usable for all file/artifact/object uploads

etc/kayobe/ansible/pulp-image-upload.yml
Alex-Welsh: Superseded by other comments

.github/workflows/overcloud-host-image-build.yml
```diff
@@ -291,9 +292,10 @@ jobs:
           source venvs/kayobe/bin/activate &&
           source src/kayobe-config/kayobe-env --environment ci-builder &&
           kayobe playbook run \
-          src/kayobe-config/etc/kayobe/ansible/pulp-host-image-upload.yml \
+          src/kayobe-config/etc/kayobe/ansible/pulp-image-upload.yml \
           -e image_path=/opt/kayobe/images/overcloud-ubuntu-jammy \
-          -e host_image_tag=${{ steps.host_image_tag.outputs.host_image_tag }} \
+          -e image_tag=${{ steps.host_image_tag.outputs.host_image_tag }} \
+          -e image_name="kayobe-images" \
```
Alex-Welsh: image_name is now artifact_path (there are two of these in this file)

etc/kayobe/ansible/pulp-image-upload.yml
Alex-Welsh, line 1: pulp-artifact-upload.yml

etc/kayobe/ansible/pulp-image-upload.yml
```diff
@@ -1,21 +1,22 @@
...
       debug:
-        msg: "Image tag: {{ host_image_tag }}"
+        msg: "artifact tag: {{ artifact_tag }}"
 
     - name: Get filename
       find:
-        paths: "{{ image_path }}"
-        patterns: '*.qcow2'
+        paths: "{{ artifact_path }}"
+        patterns: "*.{{ file_extension }}"
```
Alex-Welsh: Now that we support all sorts of artifacts, this might be better as just a variable for `file_regex` so the user can more tightly control what files are found.

Alex-Welsh (changes requested):
Once you've made these changes, could you give it a test? I think we can them merge

.github/workflows/overcloud-host-image-build.yml
holliefae: This needs to be `artifact_name` I think, so we don't conflict with the regex search.

assumptionsandg:
Tested:
- Host image: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/12159394894
- IPA: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/12160444988
- IPA (promote): https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/12162747981

Alex-Welsh (approved):
Nice work @assumptionsandg

PR #1392, opened 2024-12-02
Fix Bifrost templating bug on httpboot config

seunghun1ee:


priteau (dismissed):

etc/kayobe/kolla-image-tags.yml
```diff
@@ -22,8 +22,9 @@ kolla_image_tags:
     ubuntu-jammy: 2024.1-ubuntu-jammy-20241004T094540
   horizon:
     rocky-9: 2024.1-rocky-9-20240909T144917
-  bifrost_deploy:
-    rocky-9: 2024.1-rocky-9-20240725T165045
+  bifrost:
```
Alex-Welsh, line 25: @seunghun1ee have you checked this is correct?

Alex-Welsh: You might need to either change the mapping in tools/kolla-images.py or change this back to bifrost_deploy

seunghun1ee: Yes this correctly deployed in customer's system

Alex-Welsh (approved):

PR #1393, opened 2024-12-02
Update Horizon container images to fix bug 2055784

priteau:
This bug would cause Horizon to be unable to retrieve resource providers statistics in deployments with VGPU resources.


priteau:
I have tested the `rocky-9` image and it resolves the bug.

seunghun1ee (approved):

Alex-Welsh (approved):

PR #1406, opened 2024-12-05
Reduce network interruption of quorum migration

priteau:
We can reduce the potential network connectivity interruption caused by quorum migration by stopping Keystone and Neutron last and starting them first, at the expense of longer API downtime (because each kayobe invocation first generates configuration).

### DIFF
--- a/tools/rabbitmq-quorum-migration.sh
+++ b/tools/rabbitmq-quorum-migration.sh
@@ -2,7 +2,8 @@
 
 set -ex
 
-RABBITMQ_SERVICES_TO_RESTART=barbican,blazar,cinder,cloudkitty,designate,heat,ironic,keystone,magnum,manila,neutron,nova,octavia
+RABBITMQ_CRITICAL_SERVICES_TO_RESTART=keystone,neutron
+RABBITMQ_SERVICES_TO_RESTART=barbican,blazar,cinder,cloudkitty,designate,heat,ironic,magnum,manila,nova,octavia
 RABBITMQ_CONTAINER_NAME=rabbitmq
 
 if [[ ! $KAYOBE_CONFIG_PATH ]]; then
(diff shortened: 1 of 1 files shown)


MoteHue:
Any reason this is still a draft @priteau? It looks very helpful for OVS-based systems, would be good to get it in :)

priteau:
I set it to draft because I wanted to hear thoughts on the approach. The code change itself is ready.

MoteHue:
I think this approach is fine, although it does mean there will be a longer downtime on the non-critical services.
Perhaps we can just split the tasks when OVS is in use?

Alex-Welsh:
@MoteHue didn't you try this recently? Can we merge it now?

MoteHue:
> @MoteHue didn't you try this recently? Can we merge it now?

I haven't tried this change personally. I'd prefer that we change this split to be done via a flag, or lookup if OVS is in use. Otherwise we're just extending the outage window on OVN systems with no benefit.

Alex-Welsh:
@grzegorzkoper You have an upcoming OVS upgrade right? Would you mind testing this change when you do that?

grzegorzkoper:
@Alex-Welsh : I can discuss it with our client, if they don't mind I can run it like that.

grzegorzkoper:
> @Alex-Welsh : I can discuss it with our client, if they don't mind I can run it like that.

They prefer shorter networking downtime vs API downtime. I can test it next week.

assumptionsandg:
Potentially unrelated, but this is the approach we used for the quorum migration for OVS at Cambridge. https://gitlab.developers.cam.ac.uk/rcs/platforms/cloud-services/arcus-kayobe-config/-/merge_requests/537/diffs#699dcf399709f4a7d55ccaeed5717da26b459bb8

grzegorzkoper:
> @grzegorzkoper You have an upcoming OVS upgrade right? Would you mind testing this change when you do that?

Tested, worked like a charm.
We did hit an issue while upgrading RMQ, API downtime was long enough to trigger dhcp issues (leases expired, no new leases since services were down) It was nice to be able to bring keystone and neutron back faster and resolve them.

👍

Alex-Welsh:
Looks like a good tradeoff on OVS clouds, but it's a bit of an unnecessary slowdown otherwise. Could we make a `--ovs` flag to enable these tweaks?

PR #1164, opened 2024-07-17
Rocky Linux 9.4 support

grzegorzkoper:

PR #1413, opened 2024-12-09
Update for latest RL 9.5 image tags

MaxBed4d:


etc/kayobe/kolla-image-tags.yml
```diff
@@ -4,7 +4,7 @@
 # where the key is the OS distro and the value is the tag to deploy.
 kolla_image_tags:
   openstack:
-    rocky-9: 2024.1-rocky-9-20240903T113235
+    rocky-9: 2024.1-rocky-9-20241126T102910
```
Alex-Welsh, line 7: If you've rebuilt all the images, you can delete all the other rocky-9 tags in this file

Alex-Welsh (changes requested):
Can you tell us what testing you've done for these images? Minor version bumps like this are one of the things we need to be most careful with.

Missing some changes around [here](https://github.com/stackhpc/stackhpc-kayobe-config/blame/stackhpc/2024.1/etc/kayobe/pulp.yml#L215)

Needs a release note

See this PR for a good example of how we've done all this before: https://github.com/stackhpc/stackhpc-kayobe-config/pull/1164/files

Alex-Welsh:
@MaxBed4d I don't want this PR to get forgotten and stale, we should try to finish it off as soon as possible. If you need a hand getting it over the line let me know

seunghun1ee:
I agree. We have new Caracal deployment ongoing that is planning to use RL 9.5 as host image. This support will be needed very soon.

cityofships:
Has this not been superseded by #1430?

PR #1416, opened 2024-12-10
OC host image modifications

g0rgamesh:
- grub timeout should always be there to troubleshoot if needed
- Openstack data source fixed cloud-init issues

### DIFF
--- a/etc/kayobe/stackhpc-overcloud-dib.yml
+++ b/etc/kayobe/stackhpc-overcloud-dib.yml
@@ -37,7 +37,9 @@ stackhpc_overcloud_dib_elements:
 stackhpc_overcloud_dib_env_vars:
   DIB_BLOCK_DEVICE_CONFIG: "{{ stackhpc_overcloud_dib_block_device_config_uefi_lvm }}"
   DIB_BOOTLOADER_DEFAULT_CMDLINE: "nofb nomodeset gfxpayload=text net.ifnames=1 rd.auto"
-  DIB_CLOUD_INIT_DATASOURCES: "ConfigDrive"
+  DIB_GRUB_TIMEOUT: "5"
+  DIB_GRUB_TIMEOUT_STYLE: "countdown"
+  DIB_CLOUD_INIT_DATASOURCES: "OpenStack, ConfigDrive"
   DIB_CONTAINERFILE_RUNTIME: "docker"
   DIB_CONTAINERFILE_NETWORK_DRIVER: "host"
   DIB_CONTAINERFILE_DOCKERFILE: "/opt/kayobe/src/stackhpc-image-elements/elements/rocky-container-stackhpc/containerfiles/9-stackhpc"
--- a/etc/kayobe/stackhpc-overcloud-dib.yml
+++ b/etc/kayobe/stackhpc-overcloud-dib.yml
@@ -38,7 +38,7 @@ stackhpc_overcloud_dib_env_vars:
   DIB_BLOCK_DEVICE_CONFIG: "{{ stackhpc_overcloud_dib_block_device_config_uefi_lvm }}"
   DIB_BOOTLOADER_DEFAULT_CMDLINE: "nofb nomodeset gfxpayload=text net.ifnames=1 rd.auto"
   DIB_GRUB_TIMEOUT: "5"
-  DIB_GRUB_TIMEOUT_STYLE: "countdown"
+  DIB_GRUB_TIMEOUT_STYLE: "menu"
   DIB_CLOUD_INIT_DATASOURCES: "OpenStack, ConfigDrive"
   DIB_CONTAINERFILE_RUNTIME: "docker"
   DIB_CONTAINERFILE_NETWORK_DRIVER: "host"
--- a/etc/kayobe/pulp-host-image-versions.yml
+++ b/etc/kayobe/pulp-host-image-versions.yml
@@ -1,5 +1,5 @@
 ---
 # Overcloud host image versioning tags
 # These images must be in SMS, since they are used by our AIO CI runners
-stackhpc_rocky_9_overcloud_host_image_version: "2024.1-20240912T145502"
-stackhpc_ubuntu_jammy_overcloud_host_image_version: "2024.1-20240911T124950"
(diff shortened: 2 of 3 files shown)


etc/kayobe/stackhpc-overcloud-dib.yml
```diff
@@ -37,7 +37,9 @@ stackhpc_overcloud_dib_elements:
 stackhpc_overcloud_dib_env_vars:
   DIB_BLOCK_DEVICE_CONFIG: "{{ stackhpc_overcloud_dib_block_device_config_uefi_lvm }}"
   DIB_BOOTLOADER_DEFAULT_CMDLINE: "nofb nomodeset gfxpayload=text net.ifnames=1 rd.auto"
-  DIB_CLOUD_INIT_DATASOURCES: "ConfigDrive"
+  DIB_GRUB_TIMEOUT: "5"
+  DIB_GRUB_TIMEOUT_STYLE: "countdown"
+  DIB_CLOUD_INIT_DATASOURCES: "OpenStack, ConfigDrive"
```
mnasiadka, line 42: Default from Kayobe is ConfigDrive (https://github.com/openstack/kayobe/blob/3c178846a4e40bfc1c5f6f3bed2f5f70bf226cf0/ansible/inventory/group_vars/all/overcloud-dib#L59) - add a reno please

etc/kayobe/stackhpc-overcloud-dib.yml
```diff
@@ -37,7 +37,9 @@ stackhpc_overcloud_dib_elements:
 stackhpc_overcloud_dib_env_vars:
   DIB_BLOCK_DEVICE_CONFIG: "{{ stackhpc_overcloud_dib_block_device_config_uefi_lvm }}"
   DIB_BOOTLOADER_DEFAULT_CMDLINE: "nofb nomodeset gfxpayload=text net.ifnames=1 rd.auto"
-  DIB_CLOUD_INIT_DATASOURCES: "ConfigDrive"
+  DIB_GRUB_TIMEOUT: "5"
+  DIB_GRUB_TIMEOUT_STYLE: "countdown"
```
mnasiadka: That's the default (well, the default is hidden, but the docs say countdown is an alias) - do we need to set that?

etc/kayobe/stackhpc-overcloud-dib.yml
mnasiadka: Actually we did set that here, but still deserves a reno

etc/kayobe/stackhpc-overcloud-dib.yml
g0rgamesh: The ones that we've built before had DIB_GRUB_TIMEOUT set to 1 and DIB_GRUB_TIMEOUT_STYLE set to hidden.

mnasiadka: I asked about DIB_GRUB_TIMEOUT_STYLE, because DIB bootloader element  readme says:

* ``DIB_GRUB_TIMEOUT_STYLE`` sets the visibility of the ``grub`` menu.
  It defaults to ``hidden`` (or ``countdown`` as an alias). Set this to
  ``menu`` to display the menu and then wait for the timeout set by
  ``DIB_GRUB_TIMEOUT`` to expire before booting the default entry.
  
  That makes me think hidden is an alias for countdown - which is the default.

g0rgamesh: It's true that in both cases Grub will wait for the timer to expire, however countdown will provide additional visual cue by displaying the countdown timer. Here's the definition from gnu.org:
```
‘GRUB_TIMEOUT_STYLE’
If this option is unset or set to ‘menu’, then GRUB will display the menu and then wait for the timeout set by ‘GRUB_TIMEOUT’ to expire before booting the default entry. Pressing a key interrupts the timeout.
If this option is set to ‘countdown’ or ‘hidden’, then, before displaying the menu, GRUB will wait for the timeout set by ‘GRUB_TIMEOUT’ to expire. If ESC or F4 are pressed, or SHIFT is held down during that time, it will display the menu and wait for input. If a hotkey associated with a menu entry is pressed, it will boot the associated menu entry immediately. If the timeout expires before either of these happens, it will boot the default entry. In the ‘countdown’ case, it will show a one-line indication of the remaining time.

```
https://www.gnu.org/software/grub/manual/grub/html_node/Simple-configuration.html

mnasiadka: Let's use 'menu' for now since that's the default if it's unset - and as discussed - let's change the default upstream so it's unset.

MoteHue: Just my opinion, but I quite like the menu option. It gives a much longer window to edit boot options when we hit issues with booting. Particularly helpful on some of the more sluggish virtual consoles that drop inputs.

etc/kayobe/stackhpc-overcloud-dib.yml
g0rgamesh: Reno added.

mnasiadka:
Should we rebuild overcloud host images now and push new versions in that PR?

g0rgamesh:
> Should we rebuild overcloud host images now and push new versions in that PR?

Rocky 9: 2024.1-20241209T151515
Ubuntu: 2024.1-20250116T133659
Promoted to Ark.

mnasiadka (approved):

etc/kayobe/stackhpc-overcloud-dib.yml
```diff
@@ -37,7 +37,9 @@ stackhpc_overcloud_dib_elements:
 stackhpc_overcloud_dib_env_vars:
   DIB_BLOCK_DEVICE_CONFIG: "{{ stackhpc_overcloud_dib_block_device_config_uefi_lvm }}"
   DIB_BOOTLOADER_DEFAULT_CMDLINE: "nofb nomodeset gfxpayload=text net.ifnames=1 rd.auto"
-  DIB_CLOUD_INIT_DATASOURCES: "ConfigDrive"
+  DIB_GRUB_TIMEOUT: "5"
```
cityofships, line 40: It's working great 👍

PR #1424, opened 2024-12-12
Alert rule for active routers on ML2/OVS

jovial:
Can detect issues with HA routers on an ML2/OVS deployment. On OVN deployments we shouldn't have these metrics so the alert will have no effect.

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/openstack.rules
+++ b/etc/kayobe/kolla/config/prometheus/openstack.rules
@@ -19,6 +19,6 @@ groups:
       labels:
         severity: alert
       annotations:
-        summary: "The router {{ $labels.router_id }} is not active on eactly one agent"
-        description: "The router {{ $labels.router_id }} should be active on exactly one agent. It can either active on multiple agents or not active at all."
+        summary: "The router {{ $labels.router_id }} is not active on exactly one agent"
+        description: "The router {{ $labels.router_id }} should be active on exactly one agent. It can either be active on multiple agents or not active at all."
 {% endraw %}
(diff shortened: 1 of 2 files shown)


etc/kayobe/kolla/config/prometheus/openstack.rules
```diff
@@ -11,5 +11,14 @@ groups:
...
-
+- name: Routers
+  rules:
+    - alert: OpenStackRouterDown
+      expr: count by (router_id) (openstack_neutron_l3_agent_of_router{ha_state="active"}) != 1
+      for: 1m
+      labels:
+        severity: alert
+      annotations:
+        summary: "The router {{ $labels.router_id }} is not active on eactly one agent"
```
mnasiadka:
```suggestion
        summary: "The router {{ $labels.router_id }} is not active on exactly one agent"
```

etc/kayobe/kolla/config/prometheus/openstack.rules
```diff
@@ -11,5 +11,14 @@ groups:
...
+        description: "The router {{ $labels.router_id }} should be active on exactly one agent. It can either active on multiple agents or not active at all."
```
mnasiadka:
```suggestion
        description: "The router {{ $labels.router_id }} should be active on exactly one agent. It can either be active on multiple agents or not active at all."
```

mnasiadka (dismissed):

releasenotes/notes/adds-alert-for-active-routers-91281912213692c9.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Adds an alert to check that there is at least one active router on ML2/OVS based
```
MoteHue:
```suggestion
    Adds an alert to check that there is exactly one active router on ML2/OVS based
```

JohnGarbutt (dismissed):
That is a cool thing to check!

When the vxlan mesh is broken, they can all go active, which its a bit nuts, but good to check that.

seunghun1ee (dismissed):

MoteHue (approved):

seunghun1ee (approved):

PR #1427, opened 2024-12-13
INFRA-839 Add playbooks, config & docs for enabling Pulp tls with vault

technowhizz:


doc/source/configuration/vault.rst
```diff
@@ -298,6 +298,31 @@ Enable the required TLS variables in kayobe and kolla
...
+   .. code-block::
+
+      # Whether to enable TLS for Pulp.
+      pulp_enable_tls: true
+
+3. Redeploy pulp
+
+   .. code-block::
+
+      kayobe seed service reconfigure -t seed-deploy-containers
```
MoteHue:
```suggestion
      kayobe seed service reconfigure -t seed-deploy-containers -kt none
```
Avoids kicking Bifrost by mistake

technowhizz: Thanks

etc/kayobe/ansible/vault-generate-pulp-tls.yml
```diff
@@ -0,0 +1,53 @@
+---
+- name: Generate TLS certificate for pulp
+  hosts: controllers
+  run_once: true
+  vars:
+    vault_api_addr: "https://{{ internal_net_name | net_ip(groups['controllers'][0]) }}:8200"
```
jovial, line 6: Only thing to point out here is that we have a bootstrapping issue where we have to bring the controllers up and deploy vault before generating the certs. You could use the vault on the seed to generate the certificate for pulp instead, but is it better to always use the overcloud vault? What do people think?

mnasiadka: If we want to use vault from the seed (which might make more sense, since this is being deployed on the seed) then we'll need a second intermediate CA?

MoteHue: It would make sense to have TLS for Pulp before we need to use Pulp at all, so even before Bifrost. So +1 to not relying on the overcloud vault

technowhizz: Should we do that as a part 2? and perhaps call this 'Deploying pulp with TLS on an existing cloud?'

seunghun1ee (dismissed):
Don't we also need to add the cert to the docker config directory? /etc/docker/certs/<pulp endpoint> I think.
If we can restart docker, this is not needed though.

technowhizz:
> Don't we also need to add the cert to the docker config directory? /etc/docker/certs/ I think. If we can restart docker, this is not needed though.

Yeah, so I think docker will use the system CA trust but you might be right about needing to restart docker for that. Also if we did need to add the CA cert to docker I realised @seunghun1ee that we have a variable in kayobe for that https://github.com/stackhpc/stackhpc-kayobe-config/blob/1b04d7445755b932644a3ab943569addccd366c9/etc/kayobe/docker.yml#L33

etc/kayobe/ansible/vault-generate-pulp-tls.yml
Alex-Welsh: That's fine by me as long as we have a clear warning in the docs

doc/source/configuration/vault.rst
```diff
@@ -298,6 +298,35 @@ Enable the required TLS variables in kayobe and kolla
...
+
+   These steps are intended for enabling tls for pulp on an existing deployment as on a new deployment the overcloud vaults may not be up at this point.
+
+To enable tls for pulp using vault generated certificates, we first need to generate the certificates using vault and then configure the seed + seed-hypervisor + overcloud nodes to add the root CA to their trust.
+
+1. Run the playbook which will generate the certificates and add the root CA to the seed + seed-hypervisor + overcloud nodes
+
+   .. code-block::
+
+      kayobe playbook run $KAYOBE_CONFIG_PATH/ansible/vault-generate-pulp-tls.yml
```
Alex-Welsh, line 390: Could you add a step to encrypt the secrets that come out of this?

doc/source/configuration/vault.rst
Alex-Welsh, line 1: Could you capitalise TLS throughout this section?

Alex-Welsh:
@technowhizz what is the expiry on the certs? Do we need a note on cert rotation in the docs?

technowhizz:
@Alex-Welsh once checks pass this should be good to go too. Unless also you want this in 2025.1

re cert rotation and length - its 1 year I believe like all the other certs. Nothing out of the ordinary, hence nothing worth mentioning. As far as rotation is concerned I think we need a note for all types not just pulp. Perhaps can be done as a follow up to this

Alex-Welsh:
>re cert rotation and length - its 1 year I believe like all the other certs

This has recently changed actually, https://github.com/stackhpc/stackhpc-kayobe-config/pull/1775/files
It's been increased to 2 years so we can switch them as part of annual upgrades

>Unless also you want this in 2025.1

Yeah, if you could just rebase & change the target to 2025.1 that'd be great.

PR #1430, opened 2024-12-19
Add RL9.5 support

Alex-Welsh:
Final testing is still under way. Finishing off a multinode deployment


releasenotes/notes/rl95-6cbdf902e30502bf.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Added support for Rocky Linux 9.5, including host packages and a full
+    container image refresh.
+  - |
```
priteau, line 6: Nit: the second part should probably be an `upgrade` note instead.

Alex-Welsh: I'm not sure about this. I think in the past we've discussed what upgrade notes should be used for and decided that they are for major changes to be aware of when changing openstack release. Perhaps @mnasiadka can advise?

mnasiadka: Looking at reno docs - (https://docs.openstack.org/reno/latest/user/usage.html#editing-a-release-note) - we agreed in the past it's features most probably - but maybe when we change defaults - it should be upgrade: because most probably that's what fits in the default sections (as in people updating SKC will read upgrade section).
I would probably recommend we rethink if default sections is what we want - because we could add something like "update:" or similar, which would fit for updates inside an OpenStack release.

priteau: OK, we can leave it as `features` for now.

cityofships (approved):

PR #1431, opened 2024-12-19
Add support for Amphora images in StackHPC Release Train

seunghun1ee:
The workflow of Amphora image building, uploading, promoting, downloading, registering were tested.
~But the image itself was not.~
Image is now tested

Opening this PR because workflows are ready to review.

### DIFF
--- a/.github/workflows/amphora-image-build.yml
+++ b/.github/workflows/amphora-image-build.yml
@@ -1,0 +1,256 @@
+---
+name: Build Amphora image
+on:
+  workflow_dispatch:
+    secrets:
+      KAYOBE_VAULT_PASSWORD:
+        required: true
+      CLOUDS_YAML:
+        required: true
+      OS_APPLICATION_CREDENTIAL_ID:
+        required: true
+      OS_APPLICATION_CREDENTIAL_SECRET:
+        required: true
+
+env:
+  ANSIBLE_FORCE_COLOR: True
+  KAYOBE_ENVIRONMENT: ci-builder
+  KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
+jobs:
+  amphora-image-build:
+    name: Build Amphora image
+    if: github.repository == 'stackhpc/stackhpc-kayobe-config'
+    runs-on: arc-skc-host-image-builder-runner
+    permissions: {}
+    steps:
+
+      - name: Install Package
+        uses: ConorMacBride/install-package@main
+        with:
+          apt: git unzip nodejs python3-pip python3-venv openssh-server openssh-client jq
+
+      - name: Start the SSH service
+        run: |
+          sudo /etc/init.d/ssh start
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          path: src/kayobe-config
+
+      - name: Determine OpenStack release
+        id: openstack_release
+        run: |
+          BRANCH=$(awk -F'=' '/defaultbranch/ {print $2}' src/kayobe-config/.gitreview)
+          echo "BRANCH=$BRANCH" >> $GITHUB_OUTPUT
+          echo "openstack_release=${BRANCH}" | sed -E "s,(stable|unmaintained)/,," >> $GITHUB_OUTPUT
+
+      # Generate a tag to apply to all built Amphora image.
+      - name: Generate Amphora image tag
+        id: image_tag
+        run: |
+          echo "image_tag=$(date +${{ steps.openstack_release.outputs.openstack_release }}-%Y%m%dT%H%M%S)" >> $GITHUB_OUTPUT
+
+      - name: Display Amphora image tag
+        run: |
+          echo "${{ steps.image_tag.outputs.image_tag }}"
+
+      - name: Install Kayobe
+        run: |
+          mkdir -p venvs &&
+          pushd venvs &&
+          python3 -m venv kayobe &&
+          source kayobe/bin/activate &&
+          pip install -U pip &&
+          pip install -r ../src/kayobe-config/requirements.txt
+
+      - name: Install terraform
+        uses: hashicorp/setup-terraform@v2
+
+      - name: Initialise terraform
+        run: terraform init
+        working-directory: ${{ github.workspace }}/src/kayobe-config/terraform/aio
+
+      - name: Generate SSH keypair
+        run: ssh-keygen -f id_rsa -N ''
+        working-directory: ${{ github.workspace }}/src/kayobe-config/terraform/aio
+
+      - name: Generate clouds.yaml
+        run: |
+          cat << EOF > clouds.yaml
+          ${{ secrets.CLOUDS_YAML }}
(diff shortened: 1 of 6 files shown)


etc/kayobe/ansible/octavia-amphora-image-build.yml
```diff
@@ -15,21 +16,22 @@
     - name: Ensure packages are installed
       become: true
       vars:
+        packages_common:
+          - debootstrap
+          - git
+          - python3-venv
+          - python3-diskimage-builder
```
priteau: Why install DIB from packages when we use it from a virtual environment?

etc/kayobe/stackhpc.yml
```diff
@@ -166,3 +166,7 @@ stackhpc_docker_registry_password: "{{ pulp_password }}"
 
 # Whether or not to run CIS benchmark hardening playbooks. Default is false.
 #stackhpc_enable_cis_benchmark_hardening_hook:
+
+###############################################################################
+# Octavia Amphora image version
+stackhpc_amphora_image_version: "placeholder" # TODO(seunghun1ee): Test the image
```
priteau: Should we keep this for another PR?

priteau (dismissed):

etc/kayobe/stackhpc.yml
seunghun1ee: This was added because the amphora image download playbook relies on the tag value. Should we separate the PR into build part and download & register part?

etc/kayobe/ansible/octavia-amphora-image-build.yml
seunghun1ee: That wasn't removed from testing I think. I'll remove that.

etc/kayobe/stackhpc.yml
seunghun1ee: Now added tested image tag

etc/kayobe/ansible/octavia-amphora-image-register.yml
```diff
@@ -13,21 +11,57 @@
         lookup('env', 'OS_USERNAME') != 'octavia' or
         lookup('env', 'OS_PROJECT_NAME') != 'service'
 
+- name: Download Amphora image from Ark
+  ansible.builtin.import_playbook: pulp-amphora-image-download.yml
+  when: download_from_ark | default(true) | bool
```
MoteHue:
```suggestion
  when: download_amphora_from_ark | default(true) | bool
```
This will make it clear what the variable does if we want to override it in globals for a customer.

MoteHue: Also, I think it would be good to define this in `stackhpc.yml`  as true, instead of putting the default statement here.

MoteHue (dismissed):

.github/workflows/amphora-image-build.yml
```diff
@@ -0,0 +1,261 @@
...
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          path: src/kayobe-config
+
+      - name: Output image tag of the builder
+        id: builder_image_tag
+        run: |
+          echo image_tag=$(grep stackhpc_rocky_9_overcloud_host_image_version: etc/kayobe/pulp-host-image-versions.yml | awk '{print $2}') >> $GITHUB_OUTPUT
```
Alex-Welsh: Did you copy this in from another workflow? I don't think it is used/relevant here and can be removed

.github/workflows/amphora-image-build.yml
```diff
@@ -0,0 +1,261 @@
...
+          -e artifact_path=/opt/kayobe/images/amphora \
+          -e artifact_tag=${{ steps.image_tag.outputs.image_tag }} \
+          -e file_regex="*.qcow2" \
+          -e repository_name="amphora-images-${{ steps.openstack_release.outputs.openstack_release }}" \
+          -e pulp_base_path="amphora-images/${{ steps.openstack_release.outputs.openstack_release }}"
+        env:
+          KAYOBE_VAULT_PASSWORD: ${{ secrets.KAYOBE_VAULT_PASSWORD }}
+        if: steps.build_amphora.outcome == 'success'
+
+      - name: Copy logback
```
Alex-Welsh:
```suggestion
      - name: Copy logs back to runner
```

etc/kayobe/ansible/octavia-amphora-image-build.yml
```diff
@@ -89,15 +90,15 @@
         - name: Create the Amphora image
           ansible.builtin.shell:
             cmd: >-
-              source {{ venv_path }}/bin/activate &&
+              . {{ venv_path }}/bin/activate &&
```
Alex-Welsh: What's the issue with source here?

etc/kayobe/ansible/octavia-amphora-image-register.yml
```diff
@@ -13,21 +11,57 @@
...
+      vars:
+        os_family_required_packages:
+          Debian:
+            - python3-venv
+            - python3-dev
+          RedHat: []
+        common_required_pacakges:
+          - gcc
+        required_packages: "{{ common_required_pacakges + os_family_required_packages[ansble_facts.os_family] }}"
+        ansible_host: "{{ hostvars[groups['controllers'][0]].ansible_host }}"
```
Alex-Welsh, line 57: I think typically we use `delegate_to` and `run_once` instead of this variable

.github/workflows/amphora-image-build.yml
seunghun1ee: oops

etc/kayobe/ansible/octavia-amphora-image-build.yml
seunghun1ee: I'm not completely sure but the workflow was failing to find the command "source"

etc/kayobe/ansible/octavia-amphora-image-register.yml
seunghun1ee: This is a copy from https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2024.1/etc/kayobe/ansible/nova-compute-disable.yml#L22 

``delegate_to`` cannot be used here because we're using environment variables (openrc) from the ansible control host but want to do the openstack query from the controller.

I used ``delegate_to`` when it's possible.

.github/workflows/amphora-image-build.yml
seunghun1ee: removed

etc/kayobe/ansible/octavia-amphora-image-build.yml
Alex-Welsh: Adding `executable: "/bin/bash"` should fix that

Alex-Welsh (commented):
LGTM once we wrap up https://github.com/stackhpc/stackhpc-kayobe-config/pull/1431/files#r1946743044

Alex-Welsh (approved):

PR #1432, opened 2024-12-20
Use Caracal's Magnum

cityofships:
* Switch to Caracal release;
* Remove `magnum-cluster-api` driver.

To be merged after #1430.

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -342,12 +342,6 @@ kolla_build_blocks:
         && grafana-cli plugins install grafana-opensearch-datasource
   ironic_inspector_header: |
     ADD additions-archive /
-  magnum_base_footer: |
-    RUN curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | head -n -1 | bash
-    {% raw %}
-    {% set magnum_capi_packages = ['magnum-capi-helm==1.1.0'] %}
-    RUN {{ macros.install_pip(magnum_capi_packages | customizable("pip_packages")) }}
-    {% endraw %}
 
 # Dict mapping image customization variable names to their values.
 # Each variable takes the form:
@@ -360,6 +354,9 @@ kolla_build_customizations_common:
     - /additions/*
   ironic_inspector_pip_packages_append:
     - /additions/*
+  magnum_base_pip_packages_override:
+    - /magnum[osprofiler]
+    - magnum-capi-helm==1.2.0
   neutron_server_packages_append:
     - python3-libvirt
     - python3-ethtool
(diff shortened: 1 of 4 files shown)


cityofships:
Perhaps should be merged without waiting for CI - Magnum's not tested.

mnasiadka (dismissed):

etc/kayobe/kolla.yml
```diff
@@ -360,6 +356,9 @@ kolla_build_customizations_common:
     - /additions/*
   ironic_inspector_pip_packages_append:
     - /additions/*
+  magnum_base_pip_packages_override:
+    - /magnum[osprofiler]
+    - magnum-capi-helm
```
mnasiadka: We've lost version pinning here

etc/kayobe/kolla.yml
```diff
@@ -344,10 +344,6 @@ kolla_build_blocks:
     ADD additions-archive /
   magnum_base_footer: |
     RUN curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | head -n -1 | bash
```
mnasiadka: This is not required anymore - see https://review.opendev.org/c/openstack/kolla/+/937838

releasenotes/notes/update-magnum-containers-8814b1425ec45fc0.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Updates Magnum to a proper Caracal release version following critical bug
```
mnasiadka, line 3: Current version will also bump magnum-capi-helm to latest pypi version i.e. 1.2.0

mnasiadka (changes requested):

sd109:
FWIW unpinning the driver version here feels wrong to me. Since we're not 100% in control of upstream driver releases, we could in theory be broken if other upstream Magnum core reviewers start becoming more active in the driver lifecycle and publishing releases themselves. Maybe we could pin to minor version so that we still get automatic patch releases when rebuilding containers?

cityofships:
> FWIW unpinning the driver version here feels wrong to me. Since we're not 100% in control of upstream driver releases, we could in theory be broken if other upstream Magnum core reviewers start becoming more active in the driver lifecycle and publishing releases themselves. Maybe we could pin to minor version so that we still get automatic patch releases when rebuilding containers?

The goal of this PR was to remove Antelope's pin on Magnum itself and it scope crept a little bit. Whether to unpin the driver should be a separate discussion, so I'm just bumping its version here.

mnasiadka (approved):

PR #1434, opened 2024-12-23
Fix NONE/503 errors in Squid proxy

cityofships:


mnasiadka (changes requested):
Reno missing, otherwise looks good

mnasiadka (approved):

PR #1442, opened 2025-01-08
Fix dangerous wazuh secrets templating

Alex-Welsh:
Replace double quotes with singles in wazuh secrets template, since passwords can contain  special characters that break Ansible's parsing

### DIFF
--- a/etc/kayobe/ansible/templates/wazuh-secrets.yml.j2
+++ b/etc/kayobe/ansible/templates/wazuh-secrets.yml.j2
@@ -3,12 +3,12 @@
 # Store these securely and use lookups here
 secrets_wazuh:
   # Wazuh agent authd pass
-  authd_pass: "{{ secrets_wazuh.authd_pass | default(lookup('password', '/dev/null'), true) }}"
+  authd_pass: '{{ secrets_wazuh.authd_pass | default(lookup("password", "/dev/null"), true) }}'
   # Strengthen default wazuh api user pass
   wazuh_api_users:
     - username: "wazuh"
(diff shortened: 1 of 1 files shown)


etc/kayobe/ansible/templates/wazuh-secrets.yml.j2
```diff
@@ -3,12 +3,12 @@
 # Store these securely and use lookups here
 secrets_wazuh:
   # Wazuh agent authd pass
-  authd_pass: "{{ secrets_wazuh.authd_pass | default(lookup('password', '/dev/null'), true) }}"
+  authd_pass: '{{ secrets_wazuh.authd_pass | default(lookup('password', '/dev/null'), true) }}'
```
MoteHue: `default(lookup('password', '/dev/null')` the single quotes in here will need replacing with doubles now. Same for the rest of the file

MoteHue (changes requested):

etc/kayobe/ansible/templates/wazuh-secrets.yml.j2
Alex-Welsh: Good spot. Not sure why that didn't come up when I tested it

Alex-Welsh: Retested and same result. I guess it's cause it's a proper jinja template not an ansible playbook, so it parses differently. I'll update it anyway for consistency

Alex-Welsh: Done

MoteHue: Funky that the behaviour differs between inline and separate templates

MoteHue (approved):

PR #1445, opened 2025-01-08
feat(wazuh): Add JVM proxy configuration for Slack notifications

grzegorzkoper:
Add JVM proxy settings to wazuh-indexer configuration to enable OpenSearch Dashboard Slack notifications when running behind a proxy. The configuration:
- Sets both HTTP and HTTPS proxy host and port from http_proxy_url
- Triggers wazuh-indexer restart when proxy settings change

http_proxy_url needs to be defined before running wazuh-manager.yml

### DIFF
--- a/etc/kayobe/ansible/wazuh-manager.yml
+++ b/etc/kayobe/ansible/wazuh-manager.yml
@@ -116,6 +116,23 @@
       notify:
         - Restart wazuh
 
+    - name: Add JVM proxy settings to wazuh-indexer
+      blockinfile:
+        path: "/etc/wazuh-indexer/jvm.options"
+        state: present
+        owner: root
+        group: wazuh
+        marker: "# {mark} ANSIBLE MANAGED BLOCK JVM PROXY SETTINGS"
+        block: |
+          -Dhttp.proxyHost={{ http_proxy_url | urlsplit('hostname') }}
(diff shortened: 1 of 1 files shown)


Alex-Welsh:
Great timing, I came across the same problem today. I've tried this patch but still get errors trying to trigger a test alert in the dashboard:
image
Which is the same error as trying to send a test alert with curl and no proxy.

grzegorzkoper:
> Great timing, I came across the same problem today. I've tried this patch but still get errors trying to trigger a test alert in the dashboard: image Which is the same error as trying to send a test alert with curl and no proxy.

This looks like in your case you have a problem with resolving the name first ? 
"hooks.slack.com : Name or service not known ? "

Alex-Welsh:
> This looks like in your case you have a problem with resolving the name first ? "hooks.slack.com : Name or service not known ? "

In my case, that's the error I get when the proxy isn't working e.g.
```
$  curl example.com --noproxy '*'
curl: (6) Could not resolve host: example.com
```
(works fine when proxy is in use)

Alex-Welsh (approved):
Templating works, just didn't work my case because DNS wasn't being proxied.

(For future reference, that was fixed by adding `3.68.124.95 hooks.slack.com` to /etc/hosts)

PR #1454, opened 2025-01-14
Added parameters to AIO and image build workflows to be reused by external CI

wtripp180901:
Version of stackhpc-cloud-tests used by AIO workflow can now be overridden to allow tests on non-default branches to be validated in CI

Repository for checkout actions can now be overridden to allow the repo to be checked out by external caller workflows

also unpinned the stackhpc-cloud-tests version as otherwise requires keeping pinned version consistent across several places

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -59,7 +59,7 @@ on:
         description: Whether to perform an upgrade
         type: boolean
         default: false
-      openstack-tests-version:
+      openstack_tests_version:
         description: Git version of https://github.com/stackhpc/stackhpc-openstack-tests to use for testing
         type: string
         default: v0.0.1
@@ -452,7 +452,7 @@ jobs:
             -v $(pwd)/sot-results:/stack/sot-results \
             -e KAYOBE_ENVIRONMENT -e KAYOBE_VAULT_PASSWORD -e KAYOBE_AUTOMATION_SSH_PRIVATE_KEY \
             $KAYOBE_IMAGE \
-            /stack/kayobe-automation-env/src/kayobe-config/.automation/pipeline/playbook-run.sh '$KAYOBE_CONFIG_PATH/ansible/stackhpc-openstack-tests.yml -e sot_version=${{ inputs.openstack-tests-version }}'
+            /stack/kayobe-automation-env/src/kayobe-config/.automation/pipeline/playbook-run.sh '$KAYOBE_CONFIG_PATH/ansible/stackhpc-openstack-tests.yml -e sot_version=${{ inputs.openstack_tests_version }}'
         env:
           KAYOBE_AUTOMATION_SSH_PRIVATE_KEY: ${{ steps.ssh_key.outputs.ssh_key }}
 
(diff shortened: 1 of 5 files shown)


.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -59,6 +59,10 @@ on:
         description: Whether to perform an upgrade
         type: boolean
         default: false
+      openstack_tests_version:
```
Alex-Welsh: nit: this name is a bit deceptive, since it's not really testing openstack itself. It's a bad repo name to begin with though, I definitely see why you called it that. If `stackhpc_openstack_tests_version` fits, I'd use that

MoteHue: Would it be a big change to rename the repo?

Alex-Welsh: At this point, probably not... might be best to do it now while it's early days

etc/kayobe/ansible/stackhpc-openstack-tests.yml
Alex-Welsh, line 1: There's a few more old references to stackhpc-openstack-tests, including in the name of this file, but more generally they're out of the scope of this PR.

For now, just focus on the new things you're changing and I'll go through and make the other changes

Alex-Welsh (commented):
There's still some `stackhpc-{openstack|cloud}-tests` changes here, but it's missing a lot. I've created https://github.com/stackhpc/stackhpc-kayobe-config/pull/1456 to rename everything. I think we should either wait for that to merge first, or remove the out-of-scope changes here before merging this one

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -94,6 +102,7 @@ jobs:
       - name: Checkout ${{ inputs.upgrade && 'previous release' || 'current' }} config
         uses: actions/checkout@v4
         with:
+          repository: ${{ inputs.repository }}
           ref: ${{ inputs.upgrade && env.PREVIOUS_BRANCH || github.ref }}
```
Alex-Welsh: Will this `github.ref` point to the right branch when we are calling from a different repo?

wtripp180901: hmm that probably will try to use the caller ref, I'll see if there's a sensible way to parameterise this too

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -94,7 +110,8 @@ jobs:
       - name: Checkout ${{ inputs.upgrade && 'previous release' || 'current' }} config
         uses: actions/checkout@v4
         with:
-          ref: ${{ inputs.upgrade && env.PREVIOUS_BRANCH || github.ref }}
+          repository: ${{ inputs.repository }}
+          ref: ${{ inputs.upgrade && env.PREVIOUS_BRANCH || inputs.called_from_external && inputs.external_ref_override || github.ref }}
```
Alex-Welsh: Just to minimise the number of variables we're adding here, do you think we can rename `external_ref_override` to `github_ref` and default it to `github.ref`, then just say `${{ inputs.upgrade && env.PREVIOUS_BRANCH || github_ref }}`

It then makes this conditional a bit more simple and removes the need for `called_from_external`

Alex-Welsh (approved):

PR #1457, opened 2025-01-14
Correct permissions on Wazuh alerts manifest

MoteHue:
Wazuh instance wasnt creating wazuh-alerts-* indices and the "Check alerts index pattern" health-check fails.

Error and solution from logs:

```
Error reading manifest file: config file ("/usr/share/filebeat/module/wazuh/alerts/manifest.yml") can only be writable by the owner but the permissions are "-rw-rw-rw-"
to fix the permissions use: 'chmod go-w /usr/share/filebeat/module/wazuh/alerts/manifest.yml'
```

### DIFF
--- a/etc/kayobe/ansible/wazuh-manager.yml
+++ b/etc/kayobe/ansible/wazuh-manager.yml
@@ -130,6 +130,11 @@
       changed_when: false
       retries: 2
 
+    - name: Correct permissions on alerts manifest
+      ansible.builtin.file:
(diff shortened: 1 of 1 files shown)


MoteHue:
Would be good to test this on a fresh Wazuh deploy if anyone has the time :)

seunghun1ee (dismissed):

Alex-Welsh (approved):

PR #1458, opened 2025-01-14
Bump to wazuh-ansible v4.10.0

MoteHue:
This brings in SCA CIS checks for RL9 by default.
Keep using our fork for unmerged bugfixes

### DIFF
--- a/releasenotes/notes/wazuh-ansible-v4.10.0-ed5209199194cddf.yaml
+++ b/releasenotes/notes/wazuh-ansible-v4.10.0-ed5209199194cddf.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
+    Upgrades the version of wazuh-ansible to v4.10.0. This brings in the SCA CIS
+    checks for Rocky Linux 9 by default.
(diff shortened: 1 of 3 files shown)


MoteHue:
I've tested upgrading to this from an existing Wazuh deployment on v4.3.0 (the old `stackhpc` branch of our `wazuh-ansible` fork). Saw no issues, CIS checks appeared right away

releasenotes/notes/wazuh-ansible-v4.10.0-ed5209199194cddf.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Upgrades the version of wazuh-ansible to v4.10.0. This brings in the SCA CIS
+    checks for Rocky Linux 9 by default.
```
Alex-Welsh, line 5: Does this mean we need to update the docs for custom SCA?

MoteHue: Yes, good point

Alex-Welsh (approved):
Ideally I'd like to see this tested for Ubuntu as well but the toil required for that just ain't worth it right now. I found that the old branch just didn't work when I tried it a few days ago so definitely fixing rocky outweighs _maybe_ breaking ubuntu for me.

We should totally add Wazuh to our periodic mutinodes, and build out our tests more, but that's a problem for another day.

PR #1459, opened 2025-01-15
Bump Cephadm collection in Antelope

Alex-Welsh:
Cherry pick of https://github.com/stackhpc/stackhpc-kayobe-config/pull/1444 to mitigate multinode deployment issues such as [this](https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/12780574397/job/35627067895#step:30:227)


priteau (approved):

Alex-Welsh:
We're also going to see the same issue in Zed/Antelope upgrades, but I think we're past the point of support for Zed

PR #1464, opened 2025-01-16
feat: disable `TMOUT` for development environments

jackhodgkiss:
To avoid disruption to development work, disable the shell timeout for inactivity in the CIS hardening configuration for the `ci-aio` and `ci-multinode` environments.

### DIFF
--- a/etc/kayobe/environments/ci-aio/inventory/group_vars/cis-hardening/cis
+++ b/etc/kayobe/environments/ci-aio/inventory/group_vars/cis-hardening/cis
@@ -1,0 +1,16 @@
+---
+##############################################################################
+# Rocky 9 CIS Hardening Configuration
+
+# Disable shell timeout for inactivity which can be disruptive to
+# development work.
+rhel9cis_rule_5_4_3_2: false
+
+##############################################################################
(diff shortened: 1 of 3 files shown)


Alex-Welsh:
Seems okay to me but since it's changing a security feature I'd like to get input from others. @priteau or @mnasiadka ?

jackhodgkiss:
> Seems okay to me but since it's changing a security feature I'd like to get input from others. @priteau or @mnasiadka ?

Changing security only in specific environments for development.

Alex-Welsh:
> > Seems okay to me but since it's changing a security feature I'd like to get input from others. @priteau or @mnasiadka ?
> 
> Changing security only in specific environments for development.

They're also the only environments that we regularly test, so any difference to production is a risk

technowhizz (approved):
LGTM and makes sense as being logged out is quite disruptive. Perhaps if others aren't fond of this we can bump the timeout value

priteau (approved):
Let's merge this for now, but it would be nice to find a solution that fixes the issue for interactive environments while still applying the rule in unattended CI jobs (and potentially checking that it works as expected too).

PR #1468, opened 2025-01-21
Rebuild nova-compute image to address bug 2091033

priteau:
This bug would cause nova-compute services to freeze, which would result in frequent monitoring alerts.


Alex-Welsh (approved):
Do we need the same for Caracal?

priteau:
> Do we need the same for Caracal?

We do, but I am hoping that it will be backported upstream quickly, so we wouldn't need to use our fork: https://review.opendev.org/c/openstack/nova/+/939695

PR #1476, opened 2025-01-27
Fix permissions of the SMART metrics textfiles

cityofships:


dougszumski (approved):
Many thanks @cityofships

PR #1479, opened 2025-01-30
Make runs-on a variable

MaxBed4d:
Change runs-on to allow external calling workflows to change the runner it runs on.

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -86,7 +86,7 @@ jobs:
   all-in-one:
     name: All in one
     if: ${{ inputs.if && !cancelled() }}
-    runs-on: arc-skc-aio-runner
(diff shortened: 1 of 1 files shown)


.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -96,6 +101,7 @@ jobs:
       PREVIOUS_KAYOBE_IMAGE: ghcr.io/stackhpc/stackhpc-kayobe-config:stackhpc-2023.1
       # NOTE(upgrade): Reference the PREVIOUS release branch here.
       PREVIOUS_BRANCH: stackhpc/2023.1
+
```
Alex-Welsh: None of the other workflows have whitespace here, could you remove it?
```suggestion
```
And no, we shouldn't add whitespace to all the other workflows

Alex-Welsh (changes requested):

Alex-Welsh (approved):

PR #1481, opened 2025-01-30
Add OVN SB Relays

mnasiadka:


bbezak (approved):

Alex-Welsh:
Promotion failed: https://github.com/stackhpc/stackhpc-release-train/actions/runs/13055468937/job/36425309182
Is the container name the same as the image? if not we need another exception [here](https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2024.1/tools/kolla-images.py#L41-L78)

PR #1485, opened 2025-02-03
Bump neutron to fix PID clean up

priteau:


Alex-Welsh:
Does this affect OVN, OVS, or both?

priteau:
> Does this affect OVN, OVS, or both?

This is just for ML2/OVS.

Alex-Welsh (approved):
Multinode Rocky OVS deployed just fine: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/13112438238/job/36579084813#step:1:19

If you think this is a dangerous change we can try more combinations (i.e. minor upgrade to these containers, or Ubuntu deployments) but otherwise lgtm

priteau:
> Multinode Rocky OVS deployed just fine: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/13112438238/job/36579084813#step:1:19
> 
> If you think this is a dangerous change we can try more combinations (i.e. minor upgrade to these containers, or Ubuntu deployments) but otherwise lgtm

I don't think it is dangerous, since stable/2023.1 is unmaintained there should be only few changes.

PR #1486, opened 2025-02-03
Master: switch from Reef to Squid

Alex-Welsh:


etc/kayobe/apt.yml
```diff
@@ -33,8 +33,6 @@ apt_config:
 apt_keys:
   - url: "https://download.docker.com/linux/ubuntu/gpg"
     filename: docker.asc
-  - url: "https://download.ceph.com/keys/release.asc"
-    filename: ceph.asc
```
priteau, line 37: Not needed anymore?

Alex-Welsh: The Ceph repo was only added for Jammy because it didn't have reef in the standard ubuntu mirrors. From what I've read, the Noble mirrors do have Squid. I removed most other references to the standalone ceph repo when I switched everything from Jammy to Noble, but missed this one

priteau (approved):

PR #1487, opened 2025-02-03
Master: Rename Caracal/2024.1 to Master

Alex-Welsh:

PR #1491, opened 2025-02-04
Update rabbitmq-reset for Epoxy

MoteHue:
A couple more tweaks:
- We need to restart cinder containers on the storage nodes now, as they make use of previously-transient queues.
- Matching on the name was failing when services are in an unhealthy state, as special characters get put before the systemctl status. The description at the end always matches the container name


Alex-Welsh (dismissed):

Alex-Welsh:
Needs a rebase on latest stackhpc/master to fix CI

Alex-Welsh:
Apologies, it's still stuck on requirements.txt having the old caracal version of kayobe in it. Probably just best to wait for https://github.com/stackhpc/stackhpc-kayobe-config/pull/1487 then rebase

Alex-Welsh (approved):

PR #1495, opened 2025-02-05
Pin remaining Ansible dependencies

Alex-Welsh:
Pins remaining unpinned ansible dependencies. Pinning to the latest available version so there should be no change in behaviour

### DIFF
--- a/etc/kayobe/ansible/requirements.yml
+++ b/etc/kayobe/ansible/requirements.yml
@@ -14,6 +14,7 @@ collections:
     version: 1.1.0
 roles:
   - src: stackhpc.vxlan
+    version: 1.1.0
   - name: ansible-lockdown.ubuntu22_cis
     src: https://github.com/ansible-lockdown/UBUNTU22-CIS
     version: 1.4.1
@@ -29,3 +30,4 @@ roles:
     version: 1.18.5
   - src: https://github.com/stackhpc/ansible-role-docker.git
     name: geerlingguy.docker
+    version: stackhpc/7.0.1.1
--- a/releasenotes/notes/pin-deps-dbe52c49e945daf5.yaml
+++ b/releasenotes/notes/pin-deps-dbe52c49e945daf5.yaml
@@ -1,0 +1,6 @@
+---
+features:
+  - |
+    All Ansible dependencies are now pinned to specific versions. The
+    ``stackhpc.vxlan`` role was pinned to 1.1.0, and ``ansible-role-docker``
+    was pinned to stackhpc/7.0.1.1.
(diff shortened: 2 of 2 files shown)


etc/kayobe/ansible/requirements.yml
```diff
@@ -29,3 +30,4 @@ roles:
     version: 1.18.5
   - src: https://github.com/stackhpc/ansible-role-docker.git
     name: geerlingguy.docker
+    version: 0.0.1
```
jovial: The only strange thing about this version is that has no relation to the original versions on the role prefork: https://galaxy.ansible.com/ui/standalone/roles/geerlingguy/docker/versions/

jovial: I'm wondering if we should follow  our usual `stackhpc/<upstream major>.<upstream minor>.<upstream patch>.<# stackhpc commits ontop>` namescheme e.g https://github.com/stackhpc/kolla-ansible/tags, but agree that pinning is definitely a good idea

Alex-Welsh: Good idea, I've made a new tag and pinned to that

priteau: Thanks for highlighting our use of a fork of geerlingguy.docker, I asked @jackhodgkiss if he could update his pull request: https://github.com/geerlingguy/ansible-role-docker/pull/423

releasenotes/notes/pin-deps-dbe52c49e945daf5.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    All Ansible dependencies are now pinned to specific versions. The
+    ``stackhpc.vxlan`` role was pinned to 1.1.0, and ``ansible-role-docker``
+    was pinned to  stackhpc/7.0.1.1.
```
priteau: Nit: double space

Alex-Welsh: Well spotted, fixed

releasenotes/notes/pin-deps-dbe52c49e945daf5.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    All Ansible dependencies are now pinned to specific versions. The
+    ``stackhpc.vxlan`` role was pinned to 1.1.0, and ``ansible-role-docker``
+    was pinned to stackhpc/7.0.1.1.
```
MoteHue: Very nitpicky, but the tense here is inconsistent
```suggestion
    All Ansible dependencies are now pinned to specific versions. The
    ``stackhpc.vxlan`` role is pinned to 1.1.0, and ``ansible-role-docker``
    is pinned to stackhpc/7.0.1.1.
```

seunghun1ee (approved):

MoteHue (approved):
LGTM

PR #1503, opened 2025-02-13
Bump neutron images to fix #2097770

bbezak:
Reverts "Track all interfaces in Keepalived" so only HA interfaces are tracked. This prevents L3 HA router flapping when detaching floating IP addresses, because non-HA router interfaces did not include "no_track". 

Closes-Bug: #2097770


jackhodgkiss (approved):

MoteHue (approved):
LGTM

PR #1508, opened 2025-02-13
Re-enable CI for stackhpc/master

seunghun1ee:
We need CI to test the kolla images


Alex-Welsh:
I'd rather enable these in the same PR that we test container and host images, and probably enable them one at a time too. That way we're not repeatedly running CI that is known to fail

PR #1514, opened 2025-02-14
Remove custom blackbox exporter template

technowhizz:
This file is the same as in KA, no longer needed


Alex-Welsh (changes requested):
The [upstream version](https://github.com/stackhpc/kolla-ansible/blob/stackhpc/2024.1/ansible/roles/prometheus/templates/prometheus-blackbox-exporter.yml.j2) still doesn't have the custom modules we use e.g. `http_2xx_prometheus` here: https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2024.1/etc/kayobe/kolla/inventory/group_vars/prometheus-blackbox-exporter#L126

Alex-Welsh:
I think these changes are in master but not Caracal

Alex-Welsh:
@technowhizz is there anything you want to add before I close this?

PR #1519, opened 2025-02-17
Update docs to reference Ubuntu Noble, not Jammy

Alex-Welsh:


jackhodgkiss (approved):

MoteHue:
Do we need to add ubuntu-noble here? Or is that for a separate PR? https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/master/tools/kolla-images.py#L100-L103

MoteHue (approved):

PR #1527, opened 2025-02-19
Fix and re-enable image tag check

Alex-Welsh:


seunghun1ee (approved):

etc/kayobe/kolla-image-tags.yml
```diff
@@ -7,3 +7,6 @@ kolla_image_tags:
   openstack:
     rocky-9: master-rocky-9-20250213T114650
     ubuntu-noble: master-ubuntu-noble-20250213T091307
+  bifrost_deploy:
+    rocky-9: master-rocky-9-20250219T145044
+    ubuntu-noble: master-ubuntu-noble-20250219T145044
```
MoteHue, line 12: Why do we need a new Bifrost image for this?

Alex-Welsh: Because it wasn't built before. We had no image so the check failed

MoteHue: I see, thanks.

MoteHue (approved):

PR #1530, opened 2025-02-22
Configuration for the StackHPC fork of Redfish Exporter

m-bull:
- Use updated container image
- Update scrape jobs to not collect logs during frequent scrapes, then collect logs once per hour
- Clean up Redfish dashboard to work with Lenovo hardware and remove deprecated panel types

Dashboard needs testing for compatibility with metrics produced by other manufacturer's Redfish implementations.

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
+++ b/etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
@@ -2,10 +2,34 @@
 ---
 {% if seed_redfish_exporter_container_enabled | bool %}
 scrape_configs:
-  - job_name: redfish-exporter-seed
+  - job_name: redfish-exporter
+    params:
+      collectlogs: ['false']
     metrics_path: /redfish
-    scrape_timeout: 120s
-    scrape_interval: {{ [8 * groups['redfish_exporter_targets'] | length, 120] | max }}s
+    scrape_timeout: 300s
+    scrape_interval: {{ [8 * groups['redfish_exporter_targets'] | length, 300] | max }}s
+    relabel_configs:
+     - source_labels: [__address__]
+       target_label: __param_target
+     - source_labels: [__param_target]
+       target_label: instance
+     - target_label: __address__
+       replacement: "{{ lookup('vars', admin_oc_net_name ~ '_ips')[groups.seed.0] }}:9610"
+    static_configs:
+{% for host in groups.get('redfish_exporter_targets', []) %}
+      - targets:
+          - '{{ hostvars[host]["redfish_exporter_target_address"] }}'
+        labels:
+          server: '{{ host }}'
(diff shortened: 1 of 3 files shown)


etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
```diff
@@ -2,10 +2,34 @@
...
+    static_configs:
+{% for host in groups.get('redfish_exporter_targets', []) %}
+      - targets:
+          - '{{ hostvars[host]["redfish_exporter_target_address"] }}'
+        labels:
+          server: '{{ host }}'
+          env: "{{ kayobe_environment | default('openstack') }}"
+          group: "{{ hostvars[host]['redfish_exporter_scrape_group'] | default('overcloud') }}"
+{% endfor %}
+  - job_name: redfish-exporter-collectlog
```
dougszumski, line 27: I wondered if we should put this behind a `redfish_exporter_collect_logs` flag so we can easily disable it at sites if it causes issues. Having said that, it should be a lot more robust now it lives in a separate scrape job. Many thanks for adding it.

dougszumski (approved):
Very nice, many thanks for adding this.

etc/kayobe/kolla/config/prometheus/prometheus.yml.d/60-redfish.yml
m-bull: I think its more nuanced than that, and I couldn't quite get my brain around it when I made this PR, but I think its a bit clearer to me now...

There's two cases of scrape style (currently anyway!):
1. [iDRAC style] Scrape normally in a single job with collectlog not present in the job, just use the defaults - this is what we've always done and should be the default IMO
2. [Lenovo XCC style] Two jobs, one with collectlog=true and the other more frequent with collectlog=false

I think we should put the second style behind a feature flag as you suggest.

dougszumski: That sounds good. A limp mode flag (2) for cases when the logs are taking too long to fetch, and 1 as the default. On SMSlab (iDRAC) I noticed that most of the logs fetched are actually logs from logging into and out of the BMC - so I am hoping that once we switch to using persistent sessions, the scrape time will improve. I see about 5 minutes for an iDRAC there, which easily causes trouble.

jovial:
For me a bunch of stuff doesn't work with dell. I will try and fix up a few bits. We also have lost the health summary. Did that not work on lenovo? That was one of more useful bits for me.

m-bull:
This is just up as a record of things that worked on Lenovo,  I don't have the systems to be able to coalesce the dashboards to work on both types of hardware unfortunately :(. Added to metrics and panel names not really matching up, I didn't make any real attempt to remain compatible with the Dell metrics. 

I also had to remove some bits of the dashboard because of the Angular deprecation, though I don't remember if the health summary was one of those.

PR #1536, opened 2025-02-27
Deploy OS capacity after overcloud service upgrade

MoteHue:
We only started enabling this by default in Caracal, so this needs to get deployed as part of an Antelope->Caracal upgrade. Otherwise, the haproxy and grafana config gets put in place without the exporter being present.


Alex-Welsh:
Could you just add a note to the upgrade docs to say this will happen?

Alex-Welsh (approved):

PR #1532, opened 2025-02-24
Automate Kolla version updates

assumptionsandg:
PR tests:

- Kayobe: https://github.com/assumptionsandg/stackhpc-kayobe-config/pull/11
- Kolla-Ansible: https://github.com/assumptionsandg/stackhpc-kayobe-config/pull/12
- Kolla: https://github.com/assumptionsandg/stackhpc-kayobe-config/pull/10

This workflow will require enabling GitHub Actions pull requests (already enabled):
https://docs.github.com/en/enterprise-server@3.10/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#preventing-github-actions-from-creating-or-approving-pull-requests

PR #1548, opened 2025-02-28
Automate Kolla version updates (2023.1)

assumptionsandg:


Alex-Welsh (approved):

Alex-Welsh:
:facepalm: This won't work https://github.com/stackhpc/stackhpc-kayobe-config/pull/1532#issuecomment-2687842210

Alex-Welsh:
From external discussion:
A new iteration of this is in the works using a matrix to trigger a reusable workflow across multiple branches. Should be fixed in the next week or so. This change is fine as we can trigger it manually when we need to

PR #1551, opened 2025-03-03
Kolla-Ansible/Kayobe version enforcement

assumptionsandg:


etc/kayobe/ansible/check-kayobe-version.yml
Alex-Welsh, line 1: Should we provide a variable to skip the check?

etc/kayobe/ansible/check-kayobe-version.yml
```diff
@@ -0,0 +1,44 @@
+---
+- name: Check kayobe version
+  hosts: localhost
+  gather_facts: false
+  vars:
+    requirements_path: "{{ kayobe_config_path }}/../../requirements.txt"
+  tasks:
+    - name: Get installed kayobe commit
+      ansible.builtin.shell:
+        cmd: set -o pipefail && pip freeze | grep kayobe | cut -d @ -f 3 -s
```
Alex-Welsh: What's `-s` bringing to the table?

etc/kayobe/ansible/check-kayobe-version.yml
```diff
@@ -0,0 +1,44 @@
...
+      ansible.builtin.assert:
+        that: "kayobe_latest_version.stdout in kayobe_current_version.stdout"
+        fail_msg: |
+          Kayobe must be updated to the latest version before continuing.
+
+          Current Kayobe version: {{ kayobe_current_version.stdout }}
+          Latest Kayobe version: {{ kayobe_latest_version.stdout }}
+
+          Recreate the Kayobe environment, or install the latest version
+          by running: pip install --force-reinstall -r {{ requirements_path }}
```
Alex-Welsh: What happens to this when you install from source, then reinstall from git? Does it override properly or do you end up with two overlapping versions?

etc/kayobe/ansible/check-kayobe-version.yml
holliefae: I think we can drop it in this case.

etc/kayobe/ansible/check-kayobe-version.yml
Alex-Welsh: We can skip the check with --skip-hooks so I genuinely don't know if we should also add a var or not

etc/kayobe/ansible/check-kayobe-version.yml
holliefae: If I understand you correctly, I think if we're installing from a local checkout, then do a force re-install from the git version in the config it should overwrite the existing version.

I ran it on my local and the version seems different in pip freeze:
- Installing from source: `kayobe @ file:///home/assumptions/src/kayobe-stackhpc`
- Following reinstall from git: `kayobe @ git+https://github.com/stackhpc/kayobe@4474fb512f7bc4fe466c51f72795d44a58fa6466`

etc/kayobe/ansible/check-kayobe-version.yml
holliefae: I've added the tag option to skip-tags for kayobe-version-check in case.

assumptionsandg:
Skipping the linter as I don't think it's possible to use `git describe` as part of the Ansible git module.

etc/kayobe/ansible/check-kayobe-version.yml
Alex-Welsh, line 1: nit: Kayobe capitalisation across this file

assumptionsandg:
Requires pip 24.0 or newer. (precheck added in [88dba74](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1551/commits/88dba74a8d7621c6ce44b0326892ca79d23004ec))

etc/kayobe/ansible/check-kayobe-version.yml
```diff
@@ -0,0 +1,57 @@
...
+      community.general.pip_package_info:
+      register: packages
+
+    - name: Check if pip is version 24.0 or newer
+      ansible.builtin.assert:
+        that: "{{ packages.packages.pip.pip[0].version is version('24.0', '>=') }}"
+        fail_msg: |
+          Pip must be 24.0 or newer to run this check. Upgrade pip by running
+          pip install -U pip and reinstall Kayobe by running:
+          pip install --force-reinstall -r {{ requirements_path }}
```
holliefae: Note the order here must be observed as Kayobe installed on an older verison of pip will still show the incorrect version output in pip freeze, breaking the check.

mnasiadka:
Out of my curiosity - why are we doing that here jnstead in the Kayobe python CLI?

assumptionsandg:
> Out of my curiosity - why are we doing that here jnstead in the Kayobe python CLI?

This is more of a workaround for Kayobe/Kolla CLI version returning a different value from our downstream tagging, which don't follow pbr versioning.

See upstream bug https://bugs.launchpad.net/kayobe/+bug/2101051

etc/kayobe/ansible/check-kayobe-version.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Get installed Kayobe commit
+      ansible.builtin.shell:
+        cmd: set -o pipefail && pip freeze | grep kayobe | cut -d @ -f 3
+        executable: /usr/bin/bash
+      register: kayobe_git_commit
+
+    - name: Clone Kayobe
```
jovial: I wonder if instead of doing all this manually, could you use:

```
pip install . --dry-run --report /tmp/test.json
```

Which i think will tell you the new commit of kayobe it would install.

holliefae: I'm not sure if this approach will work here, as we're cloning Kayobe to get the commit that matches up with the currently installed tag from the git SHA outputted in the venv, rather than the newest commit from Kayobe.

jovial: So just to clarify, I was suggesting doing a dry run pip install of kayobe-config. That would work with any repo specified in requirements.txt, whereas this solution has some custom parsing of that and is hardcoded to:

```
       - name: Clone Kayobe
          ansible.builtin.git:
            repo: https://github.com/stackhpc/kayobe.git
            dest: /tmp/kayobe-git
            version: stackhpc/{{ openstack_release }}
```  

But was just a thought and can accept it might not do what you want.

holliefae: I'm not sure I understand from the context of cloning Kayobe, as we're using this purely for referencing the installed version, not the one currently in requirements.

As for getting the latest version currently defined in requirements, while this approach could work for that I think, we're using the same parsing in the update-kolla-versions [workflow](https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2024.1/.github/workflows/update-dependencies.yml#L32) so it should be fine

releasenotes/notes/kolla-kayobe-version-check-f44d43c9c34d1b89.yaml
```diff
@@ -0,0 +1,12 @@
+---
+features:
+  - |
+    Ansible playbooks to check the installed Kayobe/Kolla-Ansible versions
+    with the expected versions in Kayobe configuration. These checks
+    will run on Kayobe bootstrap, host and service operations.
```
Alex-Welsh: nit: wording could be improved
```suggestion
    Added playbooks to check the installed Kayobe/Kolla-Ansible versions
    against the expected versions in Kayobe configuration. These checks
    will run on Kayobe bootstrap, host and service operations.
```

releasenotes/notes/kolla-kayobe-version-check-f44d43c9c34d1b89.yaml
```diff
@@ -0,0 +1,12 @@
...
+upgrade:
+  - |
+    Checks are enabled by default which may break existing deployments using
+    custom forks or branches for Kayobe and Kolla-Ansible, to disable version
+    checks in configuration set ``stackhpc_enable_kayobe_check`` and
+    ``stackhpc_enable_kolla_ansible_check`` to false
```
Alex-Welsh: nit: wording could be improved
```suggestion
upgrade:
  - |
    Kolla Ansible and Kayobe version checks are enabled by default which
    may fail on existing deployments using custom forks or branches for
    Kayobe and Kolla-Ansible. To disable version checks in configuration
    set ``stackhpc_enable_kayobe_check`` and
    ``stackhpc_enable_kolla_ansible_check`` to false.
```

Alex-Welsh (commented):
Wording could be improved on the release note, but otherwise I'd be happy to merge this. We can always tweak it once it's merged. @jovial is that okay with you?

jovial:
> Wording could be improved on the release note, but otherwise I'd be happy to merge this. We can always tweak it once it's merged. @jovial is that okay with you?

Sure, lets give it a whirl and see how well it works.

etc/kayobe/ansible/check-kayobe-version.yml
jovial: So I think the virtualenv will be tracking the installed version and the dry run will download the latest version from requirements (which could be pinned) and compare.

Resolving as a possible future improvement.

jovial (approved):

PR #1571, opened 2025-03-11
Fix brackets in fluentd opensearch config

assumptionsandg:


releasenotes/notes/fix-fluentd-opensearch-template-14bde9b6d5a723f8.yaml
```diff
@@ -0,0 +1,4 @@
+---
+fixes:
+  - |
+    Fix missing bracket in fluentd opensearch config.
```
priteau: It could have been useful to mention that it only impacts configurations where caso is enabled.

priteau (dismissed):

priteau (approved):

PR #1576, opened 2025-03-14
Add a confirmation prompt to reboot.yml

MoteHue:
This is a risky playbook to run without confirmation, especially as it targets all hosts when not using a limit.

### DIFF
--- a/etc/kayobe/ansible/reboot.yml
+++ b/etc/kayobe/ansible/reboot.yml
@@ -9,9 +9,26 @@
     ansible_user: "{{ bootstrap_user if reboot_with_bootstrap_user | bool else kayobe_ansible_user }}"
     ansible_ssh_common_args: "{{ '-o StrictHostKeyChecking=no' if reboot_with_bootstrap_user | bool else '' }}"
     ansible_python_interpreter: /usr/bin/python3
+    confirm_reboot: False
   tags:
     - reboot
   tasks:
+    - name: Prompt to confirm reboot
+      ansible.builtin.pause:
+        prompt: >
+          The following hosts will be rebooted:
+          {{ play_hosts | join(', ') }}
+          If you want to proceed type: yes
+      register: pause_prompt
+      when: not confirm_reboot
+
+    - name: Fail if reboot is not confirmed
+      ansible.builtin.assert:
+        that: confirm_reboot == 'yes' or pause_prompt.user_input == 'yes'
+        msg: >
+          Reboot has not been confirmed. You must either type 'yes' when
+          prompted, or set ``confirm_reboot=yes``.
+
     - name: Reboot and wait
       become: true
       ansible.builtin.reboot:
--- a/etc/kayobe/ansible/reboot.yml
+++ b/etc/kayobe/ansible/reboot.yml
@@ -9,7 +9,7 @@
     ansible_user: "{{ bootstrap_user if reboot_with_bootstrap_user | bool else kayobe_ansible_user }}"
     ansible_ssh_common_args: "{{ '-o StrictHostKeyChecking=no' if reboot_with_bootstrap_user | bool else '' }}"
     ansible_python_interpreter: /usr/bin/python3
-    confirm_reboot: False
+    confirm_reboot: false
   tags:
     - reboot
   tasks:
@@ -27,7 +27,7 @@
         that: confirm_reboot | bool or pause_prompt.user_input == 'yes'
         msg: >
           Reboot has not been confirmed. You must either type 'yes' when
-          prompted, or set ``confirm_reboot=yes``.
+          prompted, or set ``confirm_reboot: true``.
 
     - name: Reboot and wait
       become: true
--- a/etc/kayobe/ansible/reboot.yml
+++ b/etc/kayobe/ansible/reboot.yml
@@ -24,7 +24,7 @@
 
     - name: Fail if reboot is not confirmed
       ansible.builtin.assert:
-        that: confirm_reboot == 'yes' or pause_prompt.user_input == 'yes'
+        that: confirm_reboot | bool or pause_prompt.user_input == 'yes'
         msg: >
           Reboot has not been confirmed. You must either type 'yes' when
           prompted, or set ``confirm_reboot=yes``.
--- a/etc/kayobe/environments/aufn-ceph/globals.yml
+++ b/etc/kayobe/environments/aufn-ceph/globals.yml
@@ -13,3 +13,9 @@ os_distribution: "{{ lookup('pipe', '. /etc/os-release && echo $ID') | trim }}"
 os_release: >-
           {{ (lookup('pipe', '. /etc/os-release && echo $VERSION_CODENAME') | trim) if os_distribution == 'ubuntu' else
              (lookup('pipe', '. /etc/os-release && echo $VERSION_ID') | trim | split('.') | first) if os_distribution == 'rocky' }}
+
+###############################################################################
+# Extra vars.
+
+# Don't prompt when rebooting hosts.
+confirm_reboot: "yes"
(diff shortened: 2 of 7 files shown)


etc/kayobe/ansible/fix-hostname.yml
```diff
@@ -27,4 +27,5 @@
   vars:
     reboot_hosts: fix-hostname
     reboot_with_bootstrap_user: true
+    confirm_reboot: "yes"
```
MoteHue: Open decision on if we should set this here. If we don't ci-multinode will need to be updated as it symlinks this as a hook: https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2024.1/etc/kayobe/environments/ci-multinode/hooks/overcloud-host-configure/pre.d/10-fix-hostname.yml
Otherwise, auto-deployments of ci-multinode won't work.

jovial: I'd be tempted to set confirm_reboot as an extra var in all CI environments

MoteHue: Sounds fine, means it behaves the same as it does right now. 
Done

etc/kayobe/ansible/reboot.yml
```diff
@@ -9,9 +9,26 @@
     ansible_user: "{{ bootstrap_user if reboot_with_bootstrap_user | bool else kayobe_ansible_user }}"
     ansible_ssh_common_args: "{{ '-o StrictHostKeyChecking=no' if reboot_with_bootstrap_user | bool else '' }}"
     ansible_python_interpreter: /usr/bin/python3
+    confirm_reboot: False
   tags:
     - reboot
   tasks:
+    - name: Prompt to confirm reboot
```
Alex-Welsh, line 16: I've not used this before, will it spam for every host if you don't have `run_once` and `delegate_to: localhost`?

releasenotes/notes/add-confirmation-prompt-to-reboot.yml-4fd1ae8e8d360e57.yaml
```diff
@@ -0,0 +1,11 @@
...
+features:
+  - |
+    A confirmation prompt has been added to ``reboot.yml`` to help avoid
+    rebooting the wrong hosts by mistake. This check can be skipped by setting
+    ``confirm_reboot=yes``.
+upgrade:
+  - |
+    A confirmation prompt has been added to ``reboot.yml`` to help avoid
+    rebooting the wrong hosts by mistake. This check can be skipped by setting
+    ``confirm_reboot=yes``.
```
Alex-Welsh: I probably wouldn't include this as an upgrade note, since it's not directly relevant to performing an upgrade

releasenotes/notes/add-confirmation-prompt-to-reboot.yml-4fd1ae8e8d360e57.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
+  - |
+    A confirmation prompt has been added to ``reboot.yml`` to help avoid
+    rebooting the wrong hosts by mistake. This check can be skipped by setting
+    ``confirm_reboot=yes``.
```
Alex-Welsh: nit: yamlise
```suggestion
    ``confirm_reboot: yes``.
```

etc/kayobe/ansible/reboot.yml
```diff
@@ -9,9 +9,26 @@
...
+        prompt: >
+          The following hosts will be rebooted:
+          {{ play_hosts | join(', ') }}
+          If you want to proceed type: yes
+      register: pause_prompt
+      when: not confirm_reboot
+
+    - name: Fail if reboot is not confirmed
+      ansible.builtin.assert:
+        that: confirm_reboot == 'yes' or pause_prompt.user_input == 'yes'
```
Alex-Welsh: Can we make confirm_reboot a bool rather than a str?

Alex-Welsh (changes requested):

etc/kayobe/ansible/reboot.yml
MoteHue: We set `serial: "{{ lookup('env', 'ANSIBLE_SERIAL') | default(1, true) }}"` at the top of this playbook, so it will go through each host one-by-one and prompt for each. 

> The pause module integrates into async/parallelized playbooks without any special considerations (see Rolling Updates). When using pauses with the serial playbook parameter (as in rolling updates) you are only prompted once for the current group of hosts.

The pause module does support serial, so if we set ANSIBLE_SERIAL=0 it will only prompt once for all hosts.

etc/kayobe/ansible/reboot.yml
MoteHue:
```suggestion
        that: confirm_reboot | bool or pause_prompt.user_input == 'yes'
```

releasenotes/notes/add-confirmation-prompt-to-reboot.yml-4fd1ae8e8d360e57.yaml
MoteHue:
```suggestion
```

etc/kayobe/ansible/reboot.yml
```diff
@@ -9,9 +9,26 @@
...
+          If you want to proceed type: yes
+      register: pause_prompt
+      when: not confirm_reboot
+
+    - name: Fail if reboot is not confirmed
+      ansible.builtin.assert:
+        that: confirm_reboot | bool or pause_prompt.user_input == 'yes'
+        msg: >
+          Reboot has not been confirmed. You must either type 'yes' when
+          prompted, or set ``confirm_reboot=yes``.
```
Alex-Welsh:
```suggestion
          prompted, or set ``confirm_reboot: true``.
```

etc/kayobe/environments/aufn-ceph/globals.yml
```diff
@@ -13,3 +13,9 @@ os_distribution: "{{ lookup('pipe', '. /etc/os-release && echo $ID') | trim }}"
 os_release: >-
           {{ (lookup('pipe', '. /etc/os-release && echo $VERSION_CODENAME') | trim) if os_distribution == 'ubuntu' else
              (lookup('pipe', '. /etc/os-release && echo $VERSION_ID') | trim | split('.') | first) if os_distribution == 'rocky' }}
+
+###############################################################################
+# Extra vars.
+
+# Don't prompt when rebooting hosts.
+confirm_reboot: "yes"
```
Alex-Welsh: Same for other instances but only commenting once
```suggestion
confirm_reboot: true
```

releasenotes/notes/add-confirmation-prompt-to-reboot.yml-4fd1ae8e8d360e57.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    A confirmation prompt has been added to ``reboot.yml`` to help avoid
+    rebooting the wrong hosts by mistake. This check can be skipped by setting
+    ``confirm_reboot: yes``.
```
Alex-Welsh:
```suggestion
    ``confirm_reboot: true``.
```

Alex-Welsh (changes requested):

Alex-Welsh:
Some CI errors, looks like Ark is having a bad time

Alex-Welsh (approved):

Alex-Welsh:
Part of me wants to separate this out into a task list so we can use it in other playbooks with `include_tasks`, but that seems like more of a nasty workaround than elegant solution

PR #1579, opened 2025-03-17
fix: use `rabbitmq` length for `RabbitMQNodeDown`

jackhodgkiss:
The `RabbitMQNodeDown` made the assumption that all deployments involve three controllers. However, this is not always the case as we do support deployments with a single controller or more than three controllers.

Before this would have caused false alerts in deployments with a single controller. Whilst also concealing alerts in deployments with more than three controllers.

### DIFF
--- a/etc/kayobe/kolla/config/prometheus/rabbitmq.rules
+++ b/etc/kayobe/kolla/config/prometheus/rabbitmq.rules
@@ -6,7 +6,7 @@ groups:
 - name: rabbitmq.rules
   rules:
   - alert: RabbitMQNodeDown
-    expr: sum(rabbitmq_build_info{instance!=""}) < 3
+    expr: sum(rabbitmq_build_info{instance!=""}) < {% endraw %}{{ groups['rabbitmq'] | length }}{% raw %}
     for: 30m
     labels:
       severity: critical
--- a/etc/kayobe/kolla/config/prometheus/rabbitmq.rules
+++ b/etc/kayobe/kolla/config/prometheus/rabbitmq.rules
@@ -6,7 +6,7 @@ groups:
 - name: rabbitmq.rules
   rules:
   - alert: RabbitMQNodeDown
-    expr: sum(rabbitmq_build_info{instance!=""}) < {% endraw %}{{ groups['rabbitmq'] | length }}{% raw %}
+    expr: sum(rabbitmq_build_info{instance!=""}) < {{ groups['rabbitmq'] | length }}
     for: 30m
     labels:
       severity: critical
--- a/etc/kayobe/kolla/config/prometheus/rabbitmq.rules
+++ b/etc/kayobe/kolla/config/prometheus/rabbitmq.rules
@@ -6,7 +6,7 @@ groups:
 - name: rabbitmq.rules
   rules:
   - alert: RabbitMQNodeDown
-    expr: sum(rabbitmq_build_info{instance!=""}) < {{ groups['rabbitmq'] | length }}
+    expr: sum(rabbitmq_build_info{instance!=""}) < {% endraw %}{{ alertmanager_number_of_rabbitmq_nodes }}{% raw %}
     for: 30m
     labels:
       severity: critical
--- a/releasenotes/notes/use-length-for-rabbitmq-node-down-rule-c9e9c6b09f57954d.yaml
+++ b/releasenotes/notes/use-length-for-rabbitmq-node-down-rule-c9e9c6b09f57954d.yaml
@@ -1,6 +1,8 @@
 ---
 features:
   - |
-    Use the length of the ``rabbitmq`` group to determine if any RabbitMQ
-    nodes are down. This is benefical for deployments that do not use a
-    standard three node setup.
+    Allow for easy customisation of the number of expected `RabbitMQ`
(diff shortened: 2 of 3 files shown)


etc/kayobe/kolla/config/prometheus/rabbitmq.rules
```diff
@@ -6,7 +6,7 @@ groups:
 - name: rabbitmq.rules
   rules:
   - alert: RabbitMQNodeDown
-    expr: sum(rabbitmq_build_info{instance!=""}) < 3
+    expr: sum(rabbitmq_build_info{instance!=""}) < {% endraw %}{{ groups['controllers'] | length }}{% raw %}
```
MoteHue: Is it possible to use the `rabbitmq` group from here instead? https://github.com/openstack/kayobe/blob/master/ansible/roles/kolla-ansible/templates/overcloud-components.j2#L62

Just thinking this wouldn't work if anyone has moved RabbitMQ to a different group

MoteHue (changes requested):

etc/kayobe/kolla/config/prometheus/rabbitmq.rules
jovial: I think you could potentially use the rabbitmq group here since the templating is within a raw tag and thus templated by kolla-ansible

MoteHue (approved):
LGTM now, thanks!

etc/kayobe/kolla/config/prometheus/rabbitmq.rules
```diff
@@ -6,7 +6,7 @@ groups:
 - name: rabbitmq.rules
   rules:
   - alert: RabbitMQNodeDown
-    expr: sum(rabbitmq_build_info{instance!=""}) < 3
+    expr: sum(rabbitmq_build_info{instance!=""}) < {% endraw %}{{ groups['rabbitmq'] | length }}{% raw %}
```
MoteHue: Sorry as I hit send I realised the raw tags are different here.

``rabbitmq`` is a group defined in Kolla-Ansible, rather than Kayobe, so it think we want this to be included in the whole file's raw tags. 
```suggestion
    expr: sum(rabbitmq_build_info{instance!=""}) < {{ groups['rabbitmq'] | length }}
```

MoteHue (changes requested):

MoteHue (dismissed):

jackhodgkiss:
This fails to template correctly.

```
  - alert: RabbitMQNodeDown
    expr: sum(rabbitmq_build_info{instance!=""}) < {{ groups['rabbitmq'] | length }}
    for: 30m
    labels:
```

Alex-Welsh:
Kolla-Ansible uses copy, not template, for rules files [1], so they can either be hard-coded or templated by Kayobe.

Possible Kayobe groups are: all, ungrouped, seed, seed-hypervisor, container-image-builders, hypervisors, infra-vms, wazuh-manager, wazuh-agent, github-runners, github-writer, controllers, network, monitoring, storage, compute-vgpu, compute, overcloud, vgpu, iommu, mlnx, docker, docker-registry, ntp, baremetal-compute, mgmt-switches, ctl-switches, hs-switches, switches, ceph, mons, mgrs, osds, rgws, cis-hardening, redfish_exporter_targets, fix-hostname, tempest_runner, controllers_with_ironic_enabled_False

Short term I'd say we make a new variable in SKC and default it to the length of the controller group, and have a backlog task to make the prometheus rules files templatable in KA
[1] https://github.com/openstack/kolla-ansible/blob/master/ansible/roles/prometheus/tasks/config.yml#L38

seunghun1ee:
Good idea. Happy to +1 once it's in ready-to-review state

releasenotes/notes/use-length-for-rabbitmq-node-down-rule-c9e9c6b09f57954d.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Use the length of the ``rabbitmq`` group to determine if any RabbitMQ
+    nodes are down. This is benefical for deployments that do not use a
+    standard three node setup.
```
Alex-Welsh: Needs an update in line with the latest changes

Alex-Welsh (changes requested):
Thanks Jack, I think this solution works well. Just need to update the release note

releasenotes/notes/use-length-for-rabbitmq-node-down-rule-c9e9c6b09f57954d.yaml
jackhodgkiss: Updated. Thanks.

etc/kayobe/kolla/config/prometheus/rabbitmq.rules
Alex-Welsh: tested, but didn't work, see later discussions

Alex-Welsh: tested, but didn't work, see later discussions

Alex-Welsh (approved):

seunghun1ee (approved):

PR #1583, opened 2025-03-24
Nit: Add .git to all Github Repos in kolla.yml

technowhizz:
Makes it consistent with the other URLs


priteau (changes requested):
If you are going to do this, then please fix all occurrences, both in `kolla_build_customizations_common` and `kolla_sources`.

priteau (approved):

PR #1597, opened 2025-04-07
feat: enable `skyline` build

jackhodgkiss:
Customer(s) have requested and in some cases started deploying `Skyline` using upstream images. Before we can decide to support it we should be probably build the container images for testing and evaluation purposes.


Alex-Welsh:
Probably best to just build the images and add some tags in one PR. If you only want to build images at the moment, it's easy enough to do from a feature branch

Alex-Welsh (approved):

PR #1601, opened 2025-04-08
Fix ssh in Ubuntu overcloud host images

MoteHue:
Bumps the ``stackhpc_ubuntu_jammy_overcloud_host_image_version`` to fix an issue where ``authorized_keys`` for the ``ubuntu`` user was not populated on newly provisioned nodes.


seunghun1ee (dismissed):

MoteHue:
Converted to draft: it looks like the image hasn't been built off of the pulp repo versions in SKC, instead it's from upstream sources. As such, this PR will also need to bump the repo versions so package installs don't have unhandleable dependencies.

Alex-Welsh (approved):

PR #1604, opened 2025-04-09
Fix selinux error in upgrade jobs

Alex-Welsh:
We're now checking the selinux state in stackhpc-cloud-tests. The expected state changes between 2023.1 and 2024.1, but isn't applied (reboots fail in AIOs). This PR changes the expected state when doing upgrades


seunghun1ee:
You'll need to remove extra variable at https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2024.1/.github/workflows/stackhpc-all-in-one.yml#L418 I think

seunghun1ee (approved):

jackhodgkiss (approved):

MoteHue (approved):

PR #1610, opened 2025-04-16
Fix RADOS gateway usage exporter deployment

m-bull:
credential_check.stdout is a string representation of a list, and as such will never return true when compared to an empty Python list.

Passing credential_check.stdout through the from_json filter casts the string to a list, and the equality can be properly evaluated.


Alex-Welsh:
Needs a reno, otherwise looks good

MoteHue (approved):
LGTM, thanks :)

PR #1534, opened 2025-02-27
Caracal: Update Prometheus stack

cityofships:
Depends on: https://github.com/stackhpc/kolla/pull/381

PR #1611, opened 2025-04-18
Caracal: Use Prometheus v3

cityofships:
Depends on: 
https://github.com/stackhpc/kolla/pull/394
https://github.com/stackhpc/kolla-ansible/pull/708


dougszumski (changes requested):
this should be 2.55.1

cityofships:
> this should be 2.55.1

It's being implemented in #1534

PR #1622, opened 2025-04-24
feat: add support for `TLS` generation with `OpenBao`

jackhodgkiss:


etc/kayobe/ansible/requirements.yml
```diff
@@ -9,7 +9,7 @@ collections:
   - name: stackhpc.pulp
     version: 0.5.5
   - name: stackhpc.hashicorp
-    version: 2.5.1
+    version: 2.7.0
```
jovial:
```suggestion
    version: 2.6.1
```

jovial:
Should we deprecate hashicorp vault and refer people to the openbao page? I know there is some duplication between the playbooks, but I don't see a massive problem if we remove support for vault in the near future. Any migration path for your old data?

jackhodgkiss:
> Should we deprecate hashicorp vault and refer people to the openbao page? I know there is some duplication between the playbooks, but I don't see a massive problem if we remove support for vault in the near future. Any migration path for your old data?

I think for all new deployments we should use OpenBao. It terms of migrating from vault to openbao I think TLS could be discarded and you still deploy from scratch say in the run up to a certificate replacement. However the problem is Barbican that data is valuable and should be imported.

My understanding though I haven't tested is OpenBao should be an inplace upgrade of the container image. Though we need to handle the different configuration file and docker volumes. Something that could be tested in a AIO.

jovial:
> My understanding though I haven't tested is OpenBao should be an inplace upgrade of the container image. Though we need to handle the different configuration file and docker volumes. Something that could be tested in a AIO.

Sounds promising. I'd suggest adding a notice in the vault docs noting that it is deprecated and new deployments should use OpenBao (also worth mentioning it in the release note). We can add the migration procedure once we have worked it out :)

mnasiadka:
I think we need some more write up why do we need to duplicate all playbooks - in theory we use the same client. Can't we just update the vault playbooks with some variables when openbao is enabled?

jackhodgkiss:
> I think we need some more write up why do we need to duplicate all playbooks - in theory we use the same client. Can't we just update the vault playbooks with some variables when openbao is enabled?

I decided to clone the playbooks so there is clear distinction between Vault and OpenBao I think it would be confusing to some that the playbook for deploying OpenBao is called `vault-deploy-overcloud.yml`. We also want to deprecate the use of Vault all together so at some point the vault functionaility would have to be removed it seems it would be easier to do so if they were kept as separate playbooks.

jovial (approved):
Looks good to me. Decision to merge was discussed in 8/06/2025 infra meeting.

PR #1623, opened 2025-04-24
feat: `lint` and `tox` run before `AIO` tests

jackhodgkiss:
Avoid running more resource intensive tests before linting and unit tests. As subsequent patches to fix linting and tox issues will require running AIOs again.

image


Alex-Welsh (approved):
Nice idea, this could save a lot of resources

PR #1624, opened 2025-04-25
Support Jammy to Noble upgrade

seunghun1ee:
Improving/adapting ubuntu-upgrade playbook used on Yoga Ubuntu Focal to Jammy

### DIFF
--- a/etc/kayobe/ansible/ubuntu-upgrade.yml
+++ b/etc/kayobe/ansible/ubuntu-upgrade.yml
@@ -1,11 +1,13 @@
 ---
 # To prevent Ansible role dependency errors, this playbook requires that environment variable
 # ANSIBLE_ROLES_PATH is defined and includes '$KAYOBE_PATH/ansible/roles' on the Ansible control host.
-- name: Migrate hosts from Ubuntu Jammy 22.04 to Noble 24.04
+# Where KAYOBE_PATH is the path to the source of kayobe that the environment uses.
+- name: Prepare upgrade from Ubuntu Jammy 22.04 to Noble 24.04
   hosts: overcloud:infra-vms:seed:seed-hypervisor
   vars:
     ansible_python_interpreter: /usr/bin/python3
     reboot_timeout_s: "{{ 20 * 60 }}"
+  tags: pre
   tasks:
     - name: Assert that hosts are running Ubuntu Jammy
       ansible.builtin.assert:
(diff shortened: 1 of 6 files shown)


etc/kayobe/ansible/ubuntu-upgrade.yml
```diff
@@ -63,10 +65,64 @@
           deb {{ stackhpc_repo_ubuntu_noble_security_url }} noble-security main restricted universe multiverse
       become: true
 
-    - name: Do release upgrade
-      ansible.builtin.command: do-release-upgrade -f DistUpgradeViewNonInteractive
+    - name: Enusre /tmp is mounted with exec
+      ansible.builtin.command: sudo mount -o remount,exec /tmp
       become: true
```
Alex-Welsh: There is a [mount module](https://docs.ansible.com/ansible/latest/collections/ansible/posix/mount_module.html) which would work well here

etc/kayobe/ansible/ubuntu-upgrade.yml
```diff
@@ -121,6 +207,23 @@
...
 
+    - name: Ensure only Kayobe defined apt repositories are defined
+      ansible.builtin.file:
+        path: "/etc/apt/{{ item }}"
+        state: absent
+      loop:
+        - sources.list.distUpgrade
+        - sources.list.d/third-party.sources
+        - sources.list.d/ubuntu.sources
```
Alex-Welsh, line 230: For previous versions we've renamed the file to keep it around but unused. Should we do the same here? We'd have to move it out of `sources.list.d`

Alex-Welsh (commented):
There's a few issues with whitespace in the playbooks, worth just going through them all once you're done tweaking them just to make sure it all matches

etc/kayobe/ansible/ubuntu-upgrade.yml
seunghun1ee: Those files were created during the upgrade. ``ubuntu.sources`` has upstream sources inside and other two have duplicated pulp sources that are already in ``kayobe.sources``. So, I think it's better to remove them to avoid confusing ``apt``.
They're removed only when the repositories of the host is controlled by us because of the when condition of this task, so we wouldn't need to worry about the hosts that directly uses upstream sources (Like seed hosts)

seunghun1ee:
CI wouldn't test the Ubuntu upgrade.
However, they were tested on OVS AIO, OVN AIO, OVS multinode and OVN multinode environments.

etc/kayobe/ansible/ubuntu-upgrade.yml
Alex-Welsh: I think it would be best to back up `ubuntu.sources`. Otherwise, if something goes wrong with the pulp sources, it can be difficult to revert back to upstream. I've had to do so a few times before.

It's the same as we do in kayobe when configuring apt repos: https://github.com/stackhpc/kayobe/blob/smslab/2024.1/ansible/roles/apt/tasks/repos.yml#L26

Alex-Welsh (approved):

Alex-Welsh:
We should definitely add CI for this, though it should probably go in the master branch where we run a full upgrade

PR #1633, opened 2025-05-02
Implement Jammy to noble upgrade test for CI

seunghun1ee:


.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -64,6 +64,10 @@ on:
         description: Whether to perform an upgrade
         type: boolean
         default: false
+      host_os_upgrade:
```
Alex-Welsh: Does this need to be a new var? I think we can just trigger it every time we do an openstack upgrade on ubuntu

seunghun1ee: We don't know yet if G release is going to support Ubuntu 26.04. So, I'd like to keep it flexible.

Alex-Welsh: `if: inputs.host_os_upgrade && inputs.os_release == jammy`?

seunghun1ee: for the condition of host OS upgrade task?

Alex-Welsh: Yeah, i think that's the only place it gets used

Alex-Welsh: Sorry, I've just realised I got the condition wrong. What I meant was this:
`if: inputs.upgrade && inputs.os_release == jammy?`
Then we don't need `host_os_upgrade`

seunghun1ee:
depends on https://github.com/stackhpc/stackhpc-kayobe-config/pull/1643

seunghun1ee:
depends on https://github.com/stackhpc/stackhpc-kayobe-config/pull/1656

seunghun1ee:
depends on https://github.com/stackhpc/stackhpc-kayobe-config/pull/1671 and https://github.com/stackhpc/stackhpc-kayobe-config/pull/1672

Alex-Welsh (approved):
RL9 OVS upgrade CI error is a known unrelated issue. LGTM

PR #1638, opened 2025-05-07
Fix Neutron endpoint used in Octavia OVN provider

priteau:
See related bug: https://bugs.launchpad.net/octavia/+bug/2049551


Alex-Welsh (approved):

Alex-Welsh (changes requested):
I've thought about it more, and we should build images, add tags, and add a release note in the same PR. That way our containers match the configuration in SKC

priteau:
> I've thought about it more, and we should build images, add tags, and add a release note in the same PR. That way our containers match the configuration in SKC

That's what I usually do, by updating the pull request once the images are built.

Alex-Welsh (approved):

MoteHue (approved):

PR #1641, opened 2025-05-12
CI: Remove Ubuntu OVS, add RL9 OVS upgrade cases

Alex-Welsh:
Ubuntu OVS jobs are being removed as we no longer have any clients with Ubuntu OVS systems.
Since we are removing some jobs, we can add in a RL9 OVS upgrade task without increasing resource usage.


Alex-Welsh:
Ubuntu OVS checks have already been removed as gating jobs

MoteHue (approved):

PR #1646, opened 2025-05-14
Make SMS the default runner environment

MaxBed4d:


Alex-Welsh:
It obviously works, but do you know why this says the env is SMS for an upgrade job? Just a visual bug maybe? https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/15069754840/job/42364136409?pr=1646#step:1:18

MaxBed4d:
> It obviously works, but do you know why this says the env is SMS for an upgrade job? Just a visual bug maybe? [stackhpc/stackhpc-kayobe-config/actions/runs/15069754840/job/42364136409?pr=1646#step:1:18](https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/15069754840/job/42364136409?pr=1646#step:1:18)

Because the default input for the chosen environment, I don't think, can be defined by a ternary operator dependent on another input variable (upgrade) being defined at the same time. Therefore, the default for the environment input (SMS) corresponds appropriately to the upgrade variable default (false), and so only when the upgrade is set to ``true`` do we need to change the environment from the default.

We could remove the input all together and define the environment in the ternary operator but, considering we are planning a full move to SMS of all tasks, this will be required.

Alex-Welsh (approved):

PR #1648, opened 2025-05-14
Fix PCI passthrough default templates

seunghun1ee:
Even though the variable ``gpu_group_map`` has default value of {} defined at stackhpc-compute.yml, it can be set to anything by users. Therefore, Using dict2items filter with it can cause unexpected templating error.

Fixing this by setting {} as a default of ``gpu_group_map`` before using dict2items filter.
So even if ``gpu_group_map`` becomes undefined for some reason, it doesn't fail in run time.

### DIFF
--- a/etc/kayobe/ansible/pci-passthrough.yml
+++ b/etc/kayobe/ansible/pci-passthrough.yml
@@ -11,7 +11,7 @@
     vfio_pci_ids: |-
       {% set gpu_list = [] %}
       {% set output = [] %}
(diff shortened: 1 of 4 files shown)


MoteHue (approved):
Looks sensible to me, thanks.

MoteHue (changes requested):
On second thought, it would be good to mention the fix in a release note

Alex-Welsh (changes requested):
reno

Alex-Welsh (approved):

MoteHue (approved):

PR #1649, opened 2025-05-15
CI: Replace next_nth_usable with FQCN

Alex-Welsh:
Ansible Netcommon 8.0.0 dropped support for next_nth_usable, which must now come from ansible.utils.


Alex-Welsh:
Will cherry-pick to later branches once this has merged

priteau (approved):

PR #1656, opened 2025-05-19
Manually set openstack_release to master for stackhpc/master

seunghun1ee:
With the Kayobe 2025.1 release [1], openstack_release is no longer "master" by default

[1] https://review.opendev.org/c/openstack/kayobe/+/949499

### DIFF
--- a/etc/kayobe/openstack.yml
+++ b/etc/kayobe/openstack.yml
@@ -2,14 +2,16 @@
 ###############################################################################
 # OpenStack release configuration.
 
-# Name of the current OpenStack release. Default is "master".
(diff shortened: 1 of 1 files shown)


Alex-Welsh:
This will be fixed soon upstream, I'm hoping to just use the upstream version so we don't end up with merge conflicts and anything messy downstream.

It's annoying because it does break things downstream for us but should be resolved tomorrow

seunghun1ee:
With https://review.opendev.org/c/openstack/kayobe/+/950512 being merged, this is no longer needed.

PR #1661, opened 2025-05-20
Fix whitespace issues

Alex-Welsh:
I normally wouldn't worry about this but we need to push _something_ to stackhpc/zed to rebuild the kayobe image and fix upgrade CI in Antelope, and this is better than an empty commit


MoteHue (approved):

MoteHue:
> I normally wouldn't worry about this but we need to push _something_ to stackhpc/zed to rebuild the kayobe image and fix upgrade CI in Antelope, and this is better than an empty commit

Never a wasted movement, so efficient

priteau (approved):

PR #1643, opened 2025-05-12
Rebuild kolla container images for Ubuntu Noble

seunghun1ee:
Previous images are too old that they cause downgrade of dependencies when performing upgrading from Caracal.

PR #1671, opened 2025-05-29
Bump Ubuntu package repo versions

seunghun1ee:
These were older than what stackhpc/2024.1 currently have and causing problems when packages are installed or upgraded.
This PR matches the tags with what stackhpc/2024.1 has


Alex-Welsh:
If we're bumping packages, we should probably rebuild containers as well, but given this is on master I'm not that worried about it

seunghun1ee:
It's 2 weeks ago but the rebuild was done for kolla container images too
https://github.com/stackhpc/stackhpc-kayobe-config/pull/1643
But I guess you mean to rebuild again with newer pulp repo tags?

seunghun1ee:
depends on https://github.com/stackhpc/stackhpc-kayobe-config/pull/1672

priteau (approved):

PR #1673, opened 2025-05-29
Add doc about big audit.log file after Ubuntu upgrade

seunghun1ee:


seunghun1ee:
Not needed for production system as /var/log/audit will have much larger than what CIs have

PR #1674, opened 2025-05-30
Fix release note link format

Alex-Welsh:
Currently causing CI failures https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/15342279739/job/43171651146


priteau (changes requested):
You also need a line break before the beginning of the bullet list to fix formatting of the list items.

Alex-Welsh:
Looks like this will need to bypass branch protection to merge

priteau (approved):

PR #1675, opened 2025-05-30
fix: do not attempt to install `python3-venv` on non `Ubuntu` systems

jackhodgkiss:


etc/kayobe/ansible/smartmon-tools.yml
```diff
@@ -15,10 +15,8 @@
 
     - name: Ensure Python 3, venv, and pip are installed
       ansible.builtin.package:
-        name:
-          - python3
-          - python3-venv
-          - python3-pip
+        name: >
+          {{ ['python3', 'python3-pip'] + (['python3-venv'] if ansible_facts['distribution'] == 'Ubuntu' else []) }}
```
dougszumski, line 19: It's a little quirky seeing distro conditionals used with the package module, but no big deal.

You cold define the package list in distro specific tasks as an alternative, or go back to yum/apt ansible modules, but I'm find with merging as-is.

dougszumski (approved):

seunghun1ee (approved):

PR #1678, opened 2025-06-03
CI: Add upstream-sync workflow

priteau:
We tried adding upstream-sync using stackhpc-release-train [1], but the workflow addition is skipped because our default branch is a StackHPC release branch [2].

Add workflow manually to synchronise releases that still exist upstream.

[1] https://github.com/stackhpc/stackhpc-release-train/pull/410
[2] https://github.com/stackhpc/stackhpc-release-train/commit/49c06d14375a90a61af7d4a4cb765111c1e6b7b3

### DIFF
--- a/.github/workflows/upstream-sync.yml
+++ b/.github/workflows/upstream-sync.yml
@@ -1,0 +1,34 @@
+---
+name: Upstream Sync
+'on':
+  schedule:
+    - cron: "15 8 * * 1"
+  workflow_dispatch:
+permissions:
+  contents: write
+  pull-requests: write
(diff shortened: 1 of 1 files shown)


Alex-Welsh (commented):
Since this is on a "production" branch, could we add a check so it doesn't run on customer systems?

priteau:
> Since this is on a "production" branch, could we add a check so it doesn't run on customer systems?

Good catch. I updated the workflow, hopefully with the right syntax?

Alex-Welsh (approved):

PR #1683, opened 2025-06-05
feat: add `stack` user for `Pulp`

jackhodgkiss:
The `stack` user is intended to be used for accessing `Pulp` from the `overcloud` hosts rather than using the `admin` user.

### DIFF
--- a/doc/source/configuration/release-train.rst
+++ b/doc/source/configuration/release-train.rst
@@ -52,16 +52,29 @@ The Pulp container is deployed on the seed by default, but may be disabled by
 setting ``seed_pulp_container_enabled`` to ``false`` in
 ``etc/kayobe/seed.yml``.
 
-The URL and credentials of the local Pulp server are configured in
-``etc/kayobe/pulp.yml`` via ``pulp_url``, ``pulp_username`` and
-``pulp_password``. In most cases, the default values should be sufficient.
-An admin password must be generated and set as the value of a
-``secrets_pulp_password`` variable, typically in an Ansible Vault encrypted
-``etc/kayobe/secrets.yml`` file. This password will be automatically set on
-Pulp startup.
+The URL for the local Pulp server is configured by ``pulp_url`` within
+``etc/kayobe/pulp.yml``.
 
-If a proxy is required to access the Internet from the seed, ``pulp_proxy_url``
-may be used.
+The Pulp service can be configured with two sets of credentials; one for
+administrator operations and another read-only for overcloud hosts
+to use.
+The administrator credentials can be configured ``pulp_username``,
+``pulp_password``
+The basic user account credentials can be configured with ``pulp_stack_username``
+and ``pulp_stack_password``.
+Both sets of credentials can be found within ``etc/kayobe/pulp.yml``.
+
+Both the ``pulp_password`` and ``pulp_stack_password`` are intended to be
+configured via their ``secrets_*`` counterparts, i.e.
+``secrets_pulp_password`` and ``secrets_pulp_stack_password``. These variables
+are expected to be set in an Ansible Vault encrypted
+``etc/kayobe/secrets.yml`` file.
+
+Passwords can be generated using ``OpenSSL``
+
+.. code-block:: console
+
+  openssl rand -base64 32
 
 Host images are not synchronised to the local Pulp server, since they should
 only be pulled to the seed node once. More information on host images can be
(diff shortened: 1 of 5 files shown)


etc/kayobe/pulp.yml
```diff
@@ -26,6 +26,10 @@ pulp_enable_tls: false
 pulp_username: admin
 pulp_password: "{{ secrets_pulp_password }}"
 
+# Credentials for non-admin user within Pulp.
+pulp_stack_username: stack
+pulp_stack_password: "{{ secrets_pulp_stack_password | default('omit') }}"
```
Alex-Welsh: Our "traditional" way of doing this is just to make the default an empty string i.e. `''`. I'd prefer to use that just for consistency.

etc/kayobe/stackhpc.yml
```diff
@@ -178,8 +178,8 @@ stackhpc_kolla_ansible_source_version: stackhpc/18.6.0.10
 stackhpc_docker_registry: "{{ pulp_url | regex_replace('^https?://', '') }}"
 
 # Username and password of container registry.
-stackhpc_docker_registry_username: "{{ pulp_username }}"
-stackhpc_docker_registry_password: "{{ pulp_password }}"
+stackhpc_docker_registry_username: "{{ pulp_stack_username if pulp_stack_password is defined and pulp_stack_password != 'omit' else pulp_username }}"
+stackhpc_docker_registry_password: "{{ pulp_stack_password if pulp_stack_password is defined and pulp_stack_password != 'omit' else pulp_password }}"
```
Alex-Welsh: See previous comments on default value

doc/source/configuration/release-train.rst
```diff
@@ -52,16 +52,29 @@ The Pulp container is deployed on the seed by default, but may be disabled by
...
+configured via their ``secrets_*`` counterparts, i.e.
+``secrets_pulp_password`` and ``secrets_pulp_stack_password``. These variables
+are expected to be set in an Ansible Vault encrypted
+``etc/kayobe/secrets.yml`` file.
+
+Passwords can be generated using ``OpenSSL``
+
+.. code-block:: console
+
+  openssl rand -base64 32
```
Alex-Welsh, line 77: Have you checked the password requirements? It might not like all the special chars.

doc/source/configuration/release-train.rst
```diff
@@ -52,16 +52,29 @@ The Pulp container is deployed on the seed by default, but may be disabled by
...
-Pulp startup.
-
-If a proxy is required to access the Internet from the seed, ``pulp_proxy_url``
-may be used.
+The URL for the local Pulp server is configured by ``pulp_url`` within
+``etc/kayobe/pulp.yml``.
+
+The Pulp service can be configured with two sets of credentials; one for
+administrator operations and another basic user account for overcloud hosts
+to use.
```
Alex-Welsh, line 60: Would be nice to also say that the overcloud creds are specifically read-only.

Alex-Welsh (changes requested):
Very nice change. We should make this a requirement in 2025.1.

doc/source/configuration/release-train.rst
jackhodgkiss: Pulp makes no mention of password requirements and this approach has been used elsewhere before, just has not been documented before.

etc/kayobe/pulp.yml
jackhodgkiss: The filter `default('omit')` has been used throughout SKC and would preferable to an empty string as this would mean users would only have to define `secrets_pulp_stack_password` in the `secrets.yml` and that is it. Whereas an empty string here would need to be assigned a value and then added to `secrets.yml` also this enforces consistency between deployments.

Alex-Welsh: `default('omit') != default(omit)`
I might be missing something, but I don't see any usage of 'omit' (as a string) in SKC, Kayobe, or Kolla-Ansible

>would preferable to an empty string as this would mean users would only have to define secrets_pulp_stack_password in the secrets.yml and that is it. Whereas an empty string here would need to be assigned a value and then added to secrets.yml

I'm not sure I understand the difference. If we just update the conditions, wouldn't the behaviour be identical?
To be clear, my suggestion is to replace it with this:
```suggestion
pulp_stack_password: "{{ secrets_pulp_stack_password | default('') }}"
```
Though admittedly, I've checked and we use that less in SKC than I thought, it's Kayobe and Kolla-Ansible where we use the pattern mostly.

Alex-Welsh (approved):

PR #1684, opened 2025-06-06
Remove usage of temporary subnet as external net

seunghun1ee:
This change makes the script aio-init.sh create external neutron network with existing AIO net.
With this there is no need to add additional ip address to AIO for creating external net which is not persistent during a reboot or ``kayobe overcloud host configure``.


Alex-Welsh (approved):

Alex-Welsh:
@seunghun1ee should we cherry-pick this back to any other branches?

seunghun1ee:
> @seunghun1ee should we cherry-pick this back to any other branches?

Yes

PR #1687, opened 2025-06-06
Caracal sync master

MaxBed4d:
General resync with caracal including CI fixes


etc/kayobe/stackhpc.yml
```diff
@@ -160,15 +160,23 @@ stackhpc_kolla_ansible_source_version: stackhpc/{{ openstack_release }}
...
 
 ###############################################################################
 # Feature flags
 
 # Whether or not to run CIS benchmark hardening playbooks. Default is false.
 #stackhpc_enable_cis_benchmark_hardening_hook:
 
+# Whether or not to check the installed Kayobe version when running Kayobe commands.
+# Default is true.
+stackhpc_enable_kayobe_check: true
```
seunghun1ee: This needs to be set to false for stackhpc/master
```suggestion
stackhpc_enable_kayobe_check: false
```

etc/kayobe/ansible/check-kayobe-version.yml
Alex-Welsh, line 1: These checks are going to cause problems because we don't pin our kayobe/KA versions in Master. Probably just best to globally set `stackhpc_enable_kayobe_check: false` and `stackhpc_enable_kolla_ansible_check: false` on this branch

etc/kayobe/kolla.yml
```diff
@@ -144,8 +144,24 @@ kolla_sources:
...
+    location: https://github.com/stackhpc/ovn-octavia-provider.git
+    reference: stackhpc/{{ openstack_release }}
+  octavia-driver-agent-plugin-ovn-octavia-provider:
+    type: git
+    location: https://github.com/stackhpc/ovn-octavia-provider.git
+    reference: stackhpc/{{ openstack_release }}
+  blazar-base:
+    type: git
+    location: https://github.com/stackhpc/blazar.git
+    reference: stackhpc/master
```
seunghun1ee: These need to be reverted. These are 2024.1 specific

seunghun1ee (changes requested):

etc/kayobe/kolla.yml
MaxBed4d: Sweet, will patch that now

MaxBed4d:
The Failed Ubuntu Noble OVN action is a false positive, the SCT alert is a warning for storage fill rate and can be ignored here.

seunghun1ee (approved):

etc/kayobe/kolla-image-tags.yml
```diff
@@ -15,6 +15,9 @@ kolla_image_tags:
     rocky-9: master-rocky-9-20250305T111730
   prometheus:
     rocky-9: master-rocky-9-20250430T112026
+  proxysql:
+    rocky-9: master-rocky-9-20250609T102146
+    ubuntu-noble: master-ubuntu-noble-20250609T102146
```
Alex-Welsh, line 20: Any idea why these are being updated? I'm not sure how they would be related to a Caracal sync. Did you rebuild them?

MaxBed4d: Yeah, they needed rebuilding because they were [updated](https://github.com/stackhpc/kolla-ansible/commit/aecefb1b8714c9e2b3fee6996f3e2f2276389d56) in kolla-ansible I believe

Alex-Welsh (approved):

PR #1692, opened 2025-06-09
Bump proxysql to be compatible with K-A bug #2112339 fix

seunghun1ee:
https://github.com/stackhpc/kolla-ansible/commit/aecefb1b8714c9e2b3fee6996f3e2f2276389d56 and https://github.com/stackhpc/kolla/pull/422/commits/b643e8f9fd3e9d8e881e112e53183db743f1e5d3 made old proxysql images obsolete.

PR #1693, opened 2025-06-09
Fix kolla-images.py for latest master Kolla Ansible

seunghun1ee:
Kolla-ansible change [1] introduced ``docker_image_url`` variable and it replaced the prefix for all kolla images.
Changed replacement in kolla-images.py accordingly to fix ``image_map`` not being able to be constructed well as before.

[1] https://review.opendev.org/c/openstack/kolla-ansible/+/951747


seunghun1ee:
This was included to https://github.com/stackhpc/stackhpc-kayobe-config/pull/1692

PR #1697, opened 2025-06-10
Put the RadosGW Usage Exporter behind HAProxy

MoteHue:
Metrics were coming out patchy because multiple prometheus-server instances would try to scrape from the same exporter, which would return empty if a collection is already in progress. Fix this by round-robining the requests through haproxy.

A nice graph to show it works:
image

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -560,6 +560,8 @@ kolla_build_args: {}
 kolla_overcloud_inventory_pass_through_host_vars_extra:
   - stackhpc_gpu_data
   - gpu_group_map
+  - stackhpc_radosgw_usage_exporter_frontend_port
+  - stackhpc_radosgw_usage_exporter_backend_port
 
 # List of names of host variables to pass through from kayobe hosts to
 # kolla-ansible hosts, if set. See also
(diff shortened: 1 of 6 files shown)


etc/kayobe/kolla.yml
```diff
@@ -560,6 +560,8 @@ kolla_build_args: {}
 kolla_overcloud_inventory_pass_through_host_vars_extra:
   - stackhpc_gpu_data
   - gpu_group_map
+  - "stackhpc_radosgw_usage_exporter_frontend_port"
+  - "stackhpc_radosgw_usage_exporter_backend_port"
```
Alex-Welsh: nit: Can we either drop the quotes or quote everything in the list? It's going to look odd if there's no pattern

seunghun1ee (commented):
All looks good but I think it's better to explain this change with release note

releasenotes/notes/radosgw-usage-exporter-behind-haproxy-e4371d2732f2a081.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
```
seunghun1ee, line 9: Why not also adding new feature of using HAProxy and two new variables?

MoteHue: Sure, done

seunghun1ee (approved):

Alex-Welsh (approved):

PR #1698, opened 2025-06-10
feat: increase default number of retries for checking `Pulp` status

jackhodgkiss:
It has been noticed in recent deployments that Pulp takes longer than the current default number of retries which causes `kayobe seed service deploy` to fail


releasenotes/notes/increase-pulp-retries-79cc258da9aabb4f.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
```
Alex-Welsh, line 2: very minor nit: I'd say this is a fix rather than a feature

Alex-Welsh (approved):

PR #1701, opened 2025-06-11
Fix Critical CVEs on Epoxy Kolla container images

seunghun1ee:
CVE-2024-45337 of InfluxDB was allowed as there is no upstream fix available.


priteau (approved):

priteau:
Is this using Rocky Linux 9.6?

seunghun1ee:
> Is this using Rocky Linux 9.6?

Kolla images for rocky should be built on top of 9.5 as ``stackhpc_pulp_repo_rocky_9_minor_version`` is set to 5.

priteau:
> > Is this using Rocky Linux 9.6?
> 
> Kolla images for rocky should be built on top of 9.5 as `stackhpc_pulp_repo_rocky_9_minor_version` is set to 5.

We can merge this if it works, but let's rebuild with 9.6 separately. We don't want to ship an already outdated distribution.

Alex-Welsh:
CI fix: https://github.com/stackhpc/stackhpc-kayobe-config/pull/1704

PR #1705, opened 2025-06-12
Refactor AIO script and fix for Epoxy

Alex-Welsh:


Alex-Welsh:
I'd like to remove the "manual setup" section of the docs. It's duplicated effort trying to maintain two versions (script and docs). There's a few useful bits at the bottom about Ironic testing and Tenks, but I don't think they're tested/maintained well.

Any objections?

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -1,109 +1,333 @@
...
+    KAYOBE_AUTOMATION_SSH_PRIVATE_KEY=$(cat ~/.ssh/id_rsa)
+    export KAYOBE_AUTOMATION_SSH_PRIVATE_KEY
     mkdir -p tempest-artifacts
-    sudo -E docker run --name kayobe-automation --detach -it --rm --network host \
-    -v $(pwd):/stack/kayobe-automation-env/src/kayobe-config -v $(pwd)/tempest-artifacts:/stack/tempest-artifacts \
-    -e KAYOBE_ENVIRONMENT -e KAYOBE_VAULT_PASSWORD -e KAYOBE_AUTOMATION_SSH_PRIVATE_KEY kayobe:latest \
-    /stack/kayobe-automation-env/src/kayobe-config/.automation/pipeline/tempest.sh -e ansible_user=stack
+    
+    activate_kayobe_env
+    sudo -E docker run \
```
jovial, line 220: You could run everything inside kayobe-automation or were you going the other way and removing kayobe-automation? 

The nice thing about using the kayobe automation execution environment is that would be the same execution environment as the CI jobs. 

There was a script to make things easier to run:

https://github.com/stackhpc/kayobe-automation/blob/8a24af6f8ba1447382804a2b3e30a16f5b40bace/run-local.sh#L7

Although maybe it needs extending to add some additional variables. Then it would be something like:

```
.automation/run-local.sh .automation/scripts/runbook-overcloud-deploy.sh
```
and you wouldn't need to set up the kayobe environment. You can also use it to interactively. The gist is:

```
docker run -v $PWD:/stack/kayobe-automation-env/src/kayobe-config --network host -v /tmp:/tmp  -v ~/backups:/stack/backups -d --name kayobe-zed-staging-$USER -v /usr/bin/docker:/usr/bin/docker -v /var/run/docker.sock:/var/run/docker.sock -v /etc/pki/:/etc/pki/:ro -v /etc/ssl/certs/:/etc/ssl/certs/:ro -v $HOME/.ssh/gitlab:/stack/.ssh/id_rsa:ro -v $HOME/.ssh/gitlab.pub:/stack/.ssh/id_rsa.pub:ro --privileged kayobe:zed-$USER sleep infinity
docker exec -it kayobe-zed-staging-$USER kayobe-automation-activate
```

Alex-Welsh: That's an interesting Idea

Alex-Welsh: That's an interesting idea, I hadn't really considered using Kayobe automation for more of the deployment.

Like you say, the environment would be more CI-like, which could be handy when resolving CI-related issues. It might also make it quicker to deploy if we could have an option to pull an existing kayobe automation container, rather than create it fresh each time.
The flipside is that it becomes less production-like, so it really depends on what you want to do with the environment. I guess it might also make it harder to develop something like Kayobe or Kolla Ansible themselves, since it gets wrapped up inside the container.

Either way, my focus with this was just to improve the formatting and usability of the script we have, rather than change how it actually works. I think if we do something like that it should be a follow-up PR once we decide how it looks.

Alex-Welsh: That's an interesting idea, I hadn't really considered using Kayobe automation for more of the deployment.

Like you say, the environment would be more CI-like, which could be handy when resolving CI-related issues. It might also make it quicker to deploy if we could have an option to pull an existing kayobe automation container, rather than create it fresh each time.
The flipside is that it becomes less production-like, so it really depends on what you want to do with the environment. I guess it might also make it harder to develop something like Kayobe or Kolla Ansible themselves, since it gets wrapped up inside the container.

Either way, my focus with this was just to improve the formatting and usability of the script we have, rather than change how it actually works. I think if we do something like that it should be a follow-up PR once we decide how it looks.

Alex-Welsh: We may of course also want to use this script (or something like it) in other situations.

We could make it the definitive way to run an AIO and use it directly in CI, which might make sense when we're moving to zuul. In that case we'd definitely want kayobe automation support. 

It might also be a good starting point for an Azimuth app, but I'm less sure of how they work behind the scenes

Alex-Welsh:
CI error can be ignored. Known upstream issue, should be fixed soonish

seunghun1ee (dismissed):

doc/source/contributor/environments/ci-aio.rst
```diff
@@ -29,13 +29,13 @@ Download the setup script:
 
 .. parsed-literal::
 
-   wget https://raw.githubusercontent.com/stackhpc/stackhpc-kayobe-config/stackhpc/2025.1/etc/kayobe/environments/ci-aio/automated-setup.sh
+   curl -o automated-setup.sh https://raw.githubusercontent.com/stackhpc/stackhpc-kayobe-config/stackhpc/2025.1/etc/kayobe/environments/ci-aio/automated-setup.sh
```
priteau: Nit: you can use `curl -O` to use the remote name.

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -88,22 +88,43 @@ kayobe overcloud host configure
...
 kayobe overcloud service deploy
 
+export KAYOBE_CONFIG_SOURCE_PATH=$BASE_PATH/src/kayobe-config
+export KAYOBE_VENV_PATH=$BASE_PATH/venvs/kayobe
+pushd $BASE_PATH/src/kayobe
+./dev/overcloud-test-vm.sh
+
 if $AIO_RUN_TEMPEST; then
+    if type apt; then
+        sudo apt install docker-buildx-plugin
```
priteau: Don't you need `apt -y`?

priteau (dismissed):

doc/source/contributor/environments/ci-aio.rst
Alex-Welsh: Cool, I didn't know that!

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: You appear to have an old version of this change. This issue is already fixed

priteau: Sorry, I just realised I only looked at the first commit.

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: Marking as resolved. The focus of this PR is to just improve on what we have, rather than change the approach

MoteHue:
> I'd like to remove the "manual setup" section of the docs. It's duplicated effort trying to maintain two versions (script and docs). There's a few useful bits at the bottom about Ironic testing and Tenks, but I don't think they're tested/maintained well.
> 
> Any objections?

I think it's worth keeping the testing section, but I don't really see a reason to keep the setup instructions. I'd imagine everyone just uses the script.

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -1,109 +1,333 @@
...
+}
+
+# Remove LVM configuration if not using LVM
+remove_lvm_configuration() {
+    if ! sudo vgdisplay | grep -q lvm2; then
+        echo "Removing LVM configuration..."
+        rm -f "$BASE_PATH/src/kayobe-config/etc/kayobe/environments/ci-aio/inventory/group_vars/controllers/lvm.yml" || true
+        sed -i -e '/controller_lvm_groups/,+2d' "$BASE_PATH/src/kayobe-config/etc/kayobe/environments/ci-aio/controllers.yml" || true
+    fi
+}
```
MoteHue: I think there should also be a check for `KAYOBE_AIO_LVM` here. Otherwise, you could run the script with `KAYOBE_AIO_LVM=true` and still have the LVM config removed. 

```suggestion
remove_lvm_configuration() {
    if ! sudo vgdisplay | grep -q lvm2; then
        echo "Removing LVM configuration..."
        rm -f "$BASE_PATH/src/kayobe-config/etc/kayobe/environments/ci-aio/inventory/group_vars/controllers/lvm.yml" || true
        sed -i -e '/controller_lvm_groups/,+2d' "$BASE_PATH/src/kayobe-config/etc/kayobe/environments/ci-aio/controllers.yml" || true
    elif [[ "$KAYOBE_AIO_LVM" == true ]]; then
        error_exit "You have KAYOBE_AIO_LVM set to true, but LVM is has not been found on this system. Set KAYOBE_AIO_LVM to false to ignore this warning."
cd $BASE_PATH
    fi
}
```

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -1,109 +1,333 @@
...
+    sudo ip l set dummy1 master breth1
+}
+
+# Helper function to activate the Kayobe environment
+activate_kayobe_env() {
+    KAYOBE_VAULT_PASSWORD=$(cat $VAULT_PASSWORD_FILE)
+    export KAYOBE_VAULT_PASSWORD
+    pushd "$BASE_PATH/src/kayobe-config"
+    source kayobe-env --environment ci-aio
+    source $BASE_PATH/venvs/kayobe/bin/activate
```
MoteHue, line 129: Could be good to add kayobe autocompletion here too, like we do in beokay env-vars. Just to test that activating autocompletion works.
```suggestion
    source $BASE_PATH/venvs/kayobe/bin/activate
    source <(kayobe complete)
```

Could be even better to test using beokay in this script, but that's scope creep for this PR.

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -1,109 +1,333 @@
...
+
+# Generate a Kayobe environment script for quick dev access
+generate_kayobe_env_script() {
+
+cat > $BASE_PATH/kayobe-env.sh <<EOF
+
+export KAYOBE_VAULT_PASSWORD=\$(cat $VAULT_PASSWORD_FILE)
+source $BASE_PATH/src/kayobe-config/kayobe-env --environment ci-aio
+source $BASE_PATH/venvs/kayobe/bin/activate
+
```
MoteHue, line 249: Might want autocompletion here too, like in beokay?
```suggestion
source <(kayobe complete)

```

etc/kayobe/environments/ci-aio/automated-setup.sh
```diff
@@ -1,109 +1,333 @@
...
+EOF
+}
+
+# Generate an OpenStack environment script for quick dev access
+generate_openstack_env_script() {
+
+python3 -m venv $BASE_PATH/venvs/openstack
+$BASE_PATH/venvs/openstack/bin/pip install -U pip
+$BASE_PATH/venvs/openstack/bin/pip install -U python-openstackclient
```
MoteHue: It would be good to install python-openstack client with upper constraints.

MoteHue (changes requested):
Looking good! I've tested the script myself, not found any glaring issues. Just have a couple of improvements to suggest

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: I'm personally not a fan of kayobe complete. It doesn't behave properly when you try to reference file paths. I'd prefer to leave it disabled for now. If we edit the script further in the future to use beokay then you'll probably get your wish then anyway.

Like you say, adding beokay would be a bit out of scope for this change. Not opposed to adding it in the future though.

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: See previous discussion

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: I agree with the sentiment, though I think the check is in the wrong place in the example.
I'll push a new version which includes a slightly tweaked check

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: Done, and added a new variable to keep track of all the references to the current openstack release

etc/kayobe/environments/ci-aio/automated-setup.sh
Alex-Welsh: Done

etc/kayobe/environments/ci-aio/automated-setup.sh
MoteHue: Fair enough, let's leave it out for now

MoteHue (approved):

PR #1724, opened 2025-06-18
Support OpenStack Epoxy multinode

seunghun1ee:


seunghun1ee:
Depends on the fix of K-A bug https://bugs.launchpad.net/kolla-ansible/+bug/2105505
Once a new K-A release with the fix is available, this PR will be ready-to-review

etc/kayobe/environments/ci-multinode/kolla/globals.yml
```diff
@@ -15,6 +15,10 @@ kolla_enable_tls_backend: "yes"
 # Enable RabbitMQ TLS
 rabbitmq_enable_tls: "yes"
 
+# Disable database TLS
+database_enable_tls_internal: false
+database_enable_tls_backend: false
```
Alex-Welsh, line 20: What do we need to do to be able to test this?

seunghun1ee: These are already tested with ci-multinode. Commit message explains why we need them to be disabled for now.
Or should I add comment in file too?

Alex-Welsh: Ah, no that's fine. The commit message is enough.

Alex-Welsh (approved):

seunghun1ee:
This change is specific to multinode and well tested on both CI multinode and manually deployed multinode.
Since CI-AIOs are keep crashing, this was forcefully merged

PR #1729, opened 2025-06-23
Add ironic configuration to `baremetal` environment

jackhodgkiss:
Add support for environment mixins that capture the configuration behind `ironic` for both `redfish`, `idrac` and `ipmi`. It is expected the users environments will depend on `baremetal` and will simply add the nodes to the relevant groups.


Alex-Welsh:
I'm a big fan of this approach. What do you need to get this out of draft?

jackhodgkiss:
> I'm a big fan of this approach. What do you need to get this out of draft?

I think there a some further additions that could be included here.

Horizon fixes (scopes and configuration drive default)

Some documentation. I think something for the SKC docs but also something in environment such as small README.md to summerise what the stack does/doesn't do and what you need to do if include it.

oneswig (changes requested):
@jackhodgkiss following a discussion on Friday - if we can rename the environment to be just `baremetal`, I will merge this and we will build the remaining PRs on top of it.

etc/kayobe/environments/baremetal/ironic.yml
```diff
@@ -0,0 +1,141 @@
...
+kolla_ironic_provisioning_network: 'provision-net'
+
+# List of default kernel parameters to append for baremetal PXE boot.
+kolla_ironic_pxe_append_params_default:
+  - nofb
+  - nomodeset
+  - vga=normal
+  - console=tty0
+  - console=ttyS0,115200n8
+  - "ipa-ntp-server={{ provision_wl_net_name | net_ip(inventory_hostname=groups['controllers'][0]) }}"
```
jovial: Should we use a lookup here? Thinking about the configuration dump issue

etc/kayobe/environments/baremetal/ironic.yml
```diff
@@ -0,0 +1,141 @@
...
+
+# Default vendor interface to be used for nodes that do not have
+# vendor_interface field set.
+kolla_ironic_default_vendor_interface:
+
+# Name of the Neutron network to use for cleaning.
+kolla_ironic_cleaning_network: "{{ kolla_ironic_provisioning_network if cleaning_net_name == provision_wl_net_name else 'cleaning-net' }}"
+
+# Name of the Neutron network to use for provisioning.
+kolla_ironic_provisioning_network: 'provision-net'
```
jovial:
```suggestion
kolla_ironic_provisioning_network: '{{ provision_wl_net_name }}'
```

etc/kayobe/environments/baremetal/ironic.yml
```diff
@@ -0,0 +1,141 @@
...
+# Specify the list of vendor interfaces to load during service initialization.
+kolla_ironic_enabled_vendor_interfaces:
+  - no-vendor
+
+# Default vendor interface to be used for nodes that do not have
+# vendor_interface field set.
+kolla_ironic_default_vendor_interface:
+
+# Name of the Neutron network to use for cleaning.
+kolla_ironic_cleaning_network: "{{ kolla_ironic_provisioning_network if cleaning_net_name == provision_wl_net_name else 'cleaning-net' }}"
```
jovial:
```suggestion
kolla_ironic_cleaning_network: "{{ kolla_ironic_provisioning_network if cleaning_net_name == provision_wl_net_name else cleaning_net_name }}"
```

etc/kayobe/environments/baremetal/kolla/config/nova.conf
```diff
@@ -10,3 +10,4 @@
 # - Deletion of ports.
 rpc_response_timeout = 360
 {% endif %}
+force_config_drive = True
```
jovial, line 13: Could add this only to config/nova/nova-compute-ironic.conf, but part of me likes consistency between baremetal and virtual machines

etc/kayobe/environments/baremetal/ansible/add-port-groups.yml
```diff
@@ -0,0 +1,131 @@
...
+        virtualenv: "{{ venv }}"
+        name:
+          - python-openstackclient
+          - python-ironicclient
+        state: latest
+        virtualenv_command: "python3.{{ ansible_facts.python.version.minor }} -m venv"
+        extra_args: "{% if pip_upper_constraints_file %}-c {{ pip_upper_constraints_file }}{% endif %}"
+
+- name: Ensure Ironic port groups exist
+  hosts: baremetal-compute-is-bonded
```
jovial, line 18: Potentially add baremetal-compute-is-bonded declaration to inventory for ease of discovery.

etc/kayobe/environments/baremetal/ansible/add-port-groups.yml
```diff
@@ -0,0 +1,131 @@
...
+          ansible.builtin.set_fact:
+            baremetal_uuid: "{{ (node_show.stdout | from_json).uuid }}"
+
+        - name: Get baremetal port
+          ansible.builtin.command:
+            cmd: "{{ venv }}/bin/openstack baremetal port list --long --node {{ baremetal_uuid }} -f json --sort-column Address"
+          register: port_list
+          changed_when: false
+
+        - name: Set port facts
```
jovial, line 61: Nice idea to select them by physnet!

etc/kayobe/environments/baremetal/inventory/groups
```diff
@@ -0,0 +1,8 @@
+[baremetal-idrac]
```
jovial, line 1: Doc with an example of adding a host into these groups would be great. Maybe focus on redfish?

jovial (changes requested):
All looks like great stuff to add. It would be nice to document:

- How to add the nodes into the baremetal-<driver> inventory groups. An example should be enough.
- How to configure the portgroups. The variables to need to be set and the inventory group that the hosts must be added to.

etc/kayobe/environments/baremetal/ironic.yml
jovial: Ignore this. These are the neutron net names, but why have these been customised?

etc/kayobe/environments/baremetal/ironic.yml
jovial: Ignore this. These are the neutron net names, but why have these been customised?

etc/kayobe/environments/baremetal/ironic.yml
jovial: Are these not the defaults? https://github.com/openstack/kayobe/blob/stable/2025.1/ansible/inventory/group_vars/all/ironic#L94-L98

etc/kayobe/environments/baremetal/inventory/group_vars/controllers/network_routing.yml
```diff
@@ -0,0 +1,18 @@
...
+  - name: admin-api
+    id: 1
+
+# IP routing rule to process all packets from the admin API subnet using the
+# admin-api routing table.
+internal_rules:
+  - from {{ internal_net_name | net_cidr }} table 1
+
+# IP routes for the admin-api routing table.
+internal_routes:
```
jovial, line 16: Wonder if we could include:

  - cidr: "{{ provision_wl_net_name | net_cidr }}"
    gateway: "{{ internal_net_provision_wl_gateway }}"
    table: admin-api
  - cidr: "{{ cleaning_net_name | net_cidr }}"
    gateway: "{{ internal_net_cleaning_gateway }}"
    table: admin-api

conditionally depending on whether internal_net_provision_wl_gateway and internal_net_cleaning_gateway is defined

etc/kayobe/environments/baremetal/inventory/groups
oneswig: Good idea but this might be something we need to document as part of the hardware onboarding workflow, and we could defer it until we have other components that build on this work.

etc/kayobe/environments/baremetal/ironic.yml
jackhodgkiss: Not sure why it was added maybe mistake when merging configs.

I have removed it.

etc/kayobe/environments/baremetal/ironic.yml
jackhodgkiss: Done.

etc/kayobe/environments/baremetal/kolla/config/nova.conf
jackhodgkiss: Happy with either approach. But will leave as is for now.

etc/kayobe/environments/baremetal/inventory/group_vars/controllers/network_routing.yml
oneswig: Let's return to this

oneswig (approved):
Thanks @jackhodgkiss , let's go

etc/kayobe/environments/baremetal/ironic.yml
```diff
@@ -0,0 +1,135 @@
...
+# vendor_interface field set.
+kolla_ironic_default_vendor_interface:
+
+# List of default kernel parameters to append for baremetal PXE boot.
+kolla_ironic_pxe_append_params_default:
+  - nofb
+  - nomodeset
+  - vga=normal
+  - console=tty0
+  - console=ttyS0,115200n8
```
cityofships, line 133: Are we sure about this - would it not prevent using a virtual console on the BMC?

PR #1744, opened 2025-07-07
Move Ansible playbooks into separate directories

Alex-Welsh:
This change subdivides the playbooks under the `etc/kayobe/ansible/` directory into categories based on the content and purpose of the playbook.

The new categories are:
- pulp
- ceph
- secret-store
- fixes
- deployment
- maintenance
- tools

The changes have been applied using various scripts, a few of which I have refined and added to this PR, namely:
 - `tools/get-new-playbook-path.sh` - returns the new category of a given playbook
 - `tools/magic-symlink-fix.sh` - Fixes all the symlinks it can find, using the previous script


tools/get-new-playbook-path.sh
```diff
@@ -0,0 +1,129 @@
...
+    "pulp-container-sync.yml"
+    "pulp-host-image-download.yml"
+    "pulp-repo-promote-production.yml"
+    "pulp-amphora-image-download.yml"
+    "pulp-repo-publish.yml"
+    "pulp-artifact-promote.yml"
+    "pulp-repo-sync.yml"
+    "pulp-artifact-upload.yml"
+    "pulp-sync-publish-promote.yml"
+    "pulp-auth-proxy.yml"
```
priteau: Can you sort alphabetically? (same for other lists)

Alex-Welsh: Done

technowhizz:
@Alex-Welsh is there a reason certain playbooks are missing from `tools/get-new-playbook-path.sh` ?

Alex-Welsh:
> @Alex-Welsh is there a reason certain playbooks are missing from `tools/get-new-playbook-path.sh` ?

Probably because they were added after I wrote the script

technowhizz:
Makes sense. Would be good to get them into the script before this gets merged though

seunghun1ee:
What do you think of dropping prefix of playbooks when the prefix and the category name is the same? Like secret-store, pulp, ceph.

Alex-Welsh:
> What do you think of dropping prefix of playbooks when the prefix and the category name is the same? Like secret-store, pulp, ceph.

Discussed in-person
Decided to keep the long name since it's useful to know exactly what the playbook does & not duplicate names e.g. `deploy.yml` could exist in multiple directories and do different things.

priteau (approved):

PR #1769, opened 2025-07-17
kolla_base_arch: derive from local uname

bbezak:
Some playbooks skip fact gathering. Kayobe's default references ansible_facts.architecture for kolla_base_arch, which may be undefined. Derive from `uname -m` so templating works without gathered facts.


g0rgamesh:
Works ok, approved

g0rgamesh (approved):

Alex-Welsh (approved):

PR #1772, opened 2025-07-17
Bump wazuh-ansible

MoteHue:


Alex-Welsh:
I think the reason we call the branch `stackhpc-v4.10.0` is because it's the stackhpc version of `v4.10.0`, so following that pattern, you should either call the new branch something like `stackhpc-v4.10.0.1` or just merge the changes straight into stackhpc-v4.10.0

priteau:
> I think the reason we call the branch `stackhpc-v4.10.0` is because it's the stackhpc version of `v4.10.0`, so following that pattern, you should either call the new branch something like `stackhpc-v4.10.0.1` or just merge the changes straight into stackhpc-v4.10.0

Agreed, for such trivial bug fixes there is no point in creating a new branch.

priteau:
The fix was merged in branch `stackhpc-v4.10.0`.

PR #1779, opened 2025-07-18
[Kolla] Use docker-ce repos from ark 

grzegorzkoper:


Alex-Welsh (changes requested):
So we can remove the hacky edits for apt kolla.yml?

also needs a reno

grzegorzkoper:
> So we can remove the hacky edits for apt kolla.yml?
> 
> also needs a reno

Looks like we can, better to test it first :D Trade off is we need to maintain repos.yaml, preferably point what repos we can to Ark ?

I've started looking at resurrecting https://review.opendev.org/c/openstack/kolla/+/909879, which would let us remove the hacky edits for rpm too .

priteau:
@grzegorzkoper Would you be able to refresh this pull request?

grzegorzkoper:
@priteau Refreshed, rebased onto stackhpc/2025.1 (no conflicts) and added a reno.

@Alex-Welsh Reno added.

On the apt hacks question: I think the proper fix is to move the base Ubuntu + UCA repos out of /etc/apt/sources.list blob in kolla_build_blocks and into etc/kayobe/kolla/repos.yaml. Then we can drop the rm -f /etc/apt/sources.list* shell. That's a fair bit larger than this change though, so I'd rather do it as a follow-up to keep this PR small and mergeable.

On the Rocky side we can also normalise to the dict form (url: + gpg_key:) but https://review.opendev.org/c/openstack/kolla/+/909879 will not be backported to 2025.1 so this would need to wait for 2025.2 (or 2026.1) to catch-up.

Happy to do a follow-up if that sounds reasonable.

priteau:
@grzegorzkoper Sorry to ask you to update again, but this should target 2026.1 now.

PR #1782, opened 2025-07-21
docs: update certificate rotation warning

jackhodgkiss:
The original section stated that if the overcloud certificate expired it would not have any impact the cloud however this is not true and will cause issues with Barbican if integrated.

### DIFF
--- a/doc/source/configuration/openbao.rst
+++ b/doc/source/configuration/openbao.rst
@@ -110,10 +110,9 @@ Setup OpenBao HA on the overcloud hosts
 Rotating OpenBao certificate on the overcloud hosts
 ---------------------------------------------------
 
-The certificate for the overcloud OpenBao has an expiry time of one year. While
-the cloud won't break if this expires, it will need rotating before new
-certificates can be generated for internal PKI. If a OpenBao becomes sealed, it
-cannot be unsealed with an expired certificate.
+The certificate for the overcloud OpenBao has an expiration of one or two years after the certificate was generated.
+The expiration date of a certificate can be determined with ``openssl x509 -enddate -noout -in overcloud.crt``
+This will be problematic if anything needs to interact with the OpenBao API such as issuing new certificates or Barbican integration.
 
 1. Delete the old certificate:
 
--- a/doc/source/configuration/vault.rst
+++ b/doc/source/configuration/vault.rst
@@ -117,10 +117,9 @@ Setup Vault HA on the overcloud hosts
 Rotating Vault certificate on the overcloud hosts
 -------------------------------------------------
 
-The certificate for the overcloud vaults has an expiry time of one year. While
-the cloud won't break if this expires, it will need rotating before new
-certificates can be generated for internal PKI. If a vault becomes sealed, it
-cannot be unsealed with an expired certificate.
+The certificate for the overcloud Vault has an expiration of one or two years after the certificate was generated.
+The expiration date of a certificate can be determined with ``openssl x509 -enddate -noout -in overcloud.crt``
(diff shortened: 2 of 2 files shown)


doc/source/configuration/openbao.rst
```diff
@@ -110,10 +110,8 @@ Setup OpenBao HA on the overcloud hosts
 Rotating OpenBao certificate on the overcloud hosts
 ---------------------------------------------------
 
-The certificate for the overcloud OpenBao has an expiry time of one year. While
-the cloud won't break if this expires, it will need rotating before new
-certificates can be generated for internal PKI. If a OpenBao becomes sealed, it
-cannot be unsealed with an expired certificate.
+The certificate for the overcloud OpenBao has an expiry time of one or two years.
+This will be problematic if anything needs to interact with the OpenBao API such as issuing new certificates or Barbican integration.
```
Alex-Welsh: If the default is one year, we should just say one year here.
I do think we should bump it to 2 years in Epoxy though
```suggestion
The certificate for the overcloud OpenBao has an expiry time of one year by default.
Certificate expiry will be problematic if anything needs to interact with the OpenBao API such as issuing new certificates or Barbican integration.
```

doc/source/configuration/vault.rst
```diff
@@ -117,10 +117,8 @@ Setup Vault HA on the overcloud hosts
 Rotating Vault certificate on the overcloud hosts
 -------------------------------------------------
 
-The certificate for the overcloud vaults has an expiry time of one year. While
-the cloud won't break if this expires, it will need rotating before new
-certificates can be generated for internal PKI. If a vault becomes sealed, it
-cannot be unsealed with an expired certificate.
+The certificate for the overcloud Vault has an expiry time of one or two years.
+This will be problematic if anything needs to interact with the Vault API such as issuing new certificates or Barbican integration.
```
Alex-Welsh: See previous comment

Alex-Welsh (changes requested):

doc/source/configuration/openbao.rst
jackhodgkiss: Once this merges the default will be 2 years. https://github.com/stackhpc/stackhpc-kayobe-config/pull/1775

Alex-Welsh: Well, that complicates things.
It would be nice to have slightly more clarity than just "one or two years". I think the best way to do that is just merge https://github.com/stackhpc/stackhpc-kayobe-config/pull/1775 first, then just say something like "certificates created since xxx will expire after 2 years, before it was 1 year"

jackhodgkiss: The operator should know how long a certificate is valid for. Simply merge #1775 won't changing anything unless operators reconfigure vault/openbao and then generate new certificates.

Alex-Welsh: > The operator should know how long a certificate is valid for.

These docs should be where the operator looks when they _don't_ know how long a certificate is valid for. We have plenty of clients and junior engineers who might not know how to check that sort of thing.

> Simply merge https://github.com/stackhpc/stackhpc-kayobe-config/pull/1775 won't changing anything unless operators reconfigure vault/openbao and then generate new certificates.

That's a good point. How about instead we just show people how to check:
```suggestion
The certificate for the overcloud OpenBao has an expiry time of one or two years depending on when was created.
This will be problematic if anything needs to interact with the OpenBao API such as issuing new certificates or Barbican integration.

Use the following command to check the expiry date:

.. code-block::

    openssl foo...
```

Alex-Welsh: Great minds :brain:

Alex-Welsh (approved):

PR #1795, opened 2025-07-24
Caracal to Epoxy Upgrade support

seunghun1ee:
Depends-on: https://github.com/stackhpc/stackhpc-kayobe-config/pull/1796


Alex-Welsh:
> Depends-on: #1796

Dependency has merged

releasenotes/notes/caracal-to-epoxy-upgrade-support-9f8896fa1fd230bc.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Support 2024.1 to 2025.1 upgrade in CI.
```
Alex-Welsh: nit: release notes are a list, so you can just add more elements to the same file like this:
```yaml
features:
  - |
    Thing 1.
  - |
    Thing 2.
```

Alex-Welsh: Also, release notes are only designed for customers and end-users, so you don't need to add them for CI-only changes

Alex-Welsh (approved):
LGTM if CI passes

PR #1800, opened 2025-07-28
Add first custom Ansible filter

Alex-Welsh:
Adding a custom ansible filter.

This is mainly to serve as an example that we can extend in the future with more custom filters so we can avoid some of the hairy jinja templating we've needed in the past. (See `etc/kayobe/inventory/group_vars/all/firewall` for example)

### DIFF
--- a/etc/kayobe/ansible.cfg
+++ b/etc/kayobe/ansible.cfg
@@ -10,6 +10,8 @@ inject_facts_as_vars = False
 callbacks_enabled = ansible.posix.profile_tasks
 # Silence warning about invalid characters found in group names
 force_valid_group_names = ignore
+# Default value plus custom filter plugins path
+filter_plugins = $ANSIBLE_HOME/plugins/filter:/usr/share/ansible/plugins/filter:$KAYOBE_CONFIG_PATH/ansible/filter_plugins/
 
 [inventory]
 # Fail when any inventory source cannot be parsed.
--- a/etc/kayobe/ansible/filter_plugins/filters.py
+++ b/etc/kayobe/ansible/filter_plugins/filters.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python3
+
+from ansible.errors import AnsibleFilterError
+from ansible.plugins.filter.core import to_bool
+
+class FilterModule(object):
+    def filters(self):
+        return {
+            'select_enabled': self.select_enabled
+        }
+
+    def select_enabled(self, items: list[dict], key: str = 'items') -> list:
+        """Filters a list of dictionaries to select enabled items.
+
+        Parameters:
+            items:
+                List of dictionaries - Each item contains an 'enabled' key, and
+                a key matching the "key" parameter.
+            key:
+                String - The key in each dictionary to extract when 'enabled'
+                is true.
+        
+        Returns:
+            A list of values from the specified key in each dictionary where
+            'enabled' is true. List values are flattened.
+        """
+        result = []
+        for item in items:
+            try:
+                if to_bool(item["enabled"]):
+                    result += item[key] if isinstance(item[key], list) else [item[key]]
+            except KeyError as e:
+                raise AnsibleFilterError("Key %s not found in item: %s" % (e, item))
+        return result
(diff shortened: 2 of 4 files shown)


Alex-Welsh:
This is directly compatible with https://github.com/stackhpc/stackhpc-kayobe-config/pull/1744

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,23 @@
...
+class FilterModule(object):
+    def filters(self):
+        return {
+            'select_enabled': self.select_enabled
+        }
+
+    # Takes a list of dict
+    # Returns a list of values of the key 'key' where the key 'enabled' is True
+    # If key is a list, flatten into single result list
+    def select_enabled(self, items, key='items'):
```
sd109: Would it be better to use a Python [type signature](https://docs.python.org/3/library/typing.html) or even a [docstring](https://peps.python.org/pep-0257/) here instead of a comment?

etc/kayobe/ansible.cfg
```diff
@@ -10,6 +10,8 @@ inject_facts_as_vars = False
 callbacks_enabled = ansible.posix.profile_tasks
 # Silence warning about invalid characters found in group names
 force_valid_group_names = ignore
+# Default value plus custom filter plugins path
+filter_plugins = $ANSIBLE_HOME/plugins/filter:/usr/share/ansible/plugins/filter:$KAYOBE_CONFIG_PATH/ansible/filter_plugins/
```
jovial, line 14: Would be cool if kayobe searched this path by default. Does it not?

Alex-Welsh: It does under some circumstances. I think the difference is that it does for kayobe actions but not for Kolla Ansible

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
+
+from ansible.errors import AnsibleFilterError
+from ansible.plugins.filter.core import to_bool
+
+class FilterModule(object):
+    def filters(self):
+        return {
+            'select_enabled': self.select_enabled
```
jovial, line 9: Can't you do something similar with the builtin filters:

` listofdicts| selectattr(''enabled")`

?

Alex-Welsh: Yes, I think this would do the same thing: `"{{ items | selectattr('enabled') | map(attribute='key') | flatten }}"`
Though the line it's used on is already very long so any improvement is good.

That being said, the point of this isn't really to include this filter in particular. It's mainly just to set a precedent we can follow for the future for more advanced filters

etc/kayobe/ansible/filter_plugins/filters.py
Alex-Welsh: Sounds good, added both

etc/kayobe/ansible/filter_plugins/filters.py
jovial: Fair, I like the idea of adding unit tests for more complicated bits of logic :+1:

seunghun1ee (approved):

dougszumski (commented):
Maybe I've missed something, but shouldn't the filter live in kayobe? There is already a test framework there too.

dougszumski (changes requested):
Any filters designed for Kayobe config would best live in Kayobe, so they can be used there, and in OpenStack Kayobe config etc.

Alex-Welsh:
> Any filters designed for Kayobe config would best live in Kayobe, so they can be used there, and in OpenStack Kayobe config etc.

I agree that this filter might also be useful in Kayobe upstream, though I don't think that would apply to _all_ custom filters. The idea that started this in the first place was that @GavinHeff is writing a custom fact-gathering playbook and there's some very specific templating. A custom filter would be very useful for formating that data, but would be pretty useless upstream.

The main reason I'm adding this is just so that we have a precedent for future filters that we would only care about downstream. The actual content of the filter isn't really the point of the change. Personally, I'd like to keep this downstream for now, and move it upstream if we have a need for it in the future.

PR #1808, opened 2025-07-29
Improve Ironic Horizon out of the box experience

jovial:
Policy override was causing lots of error messages in the Ironic view of Horizon.

I've asked upstream if we can get this enabled by default:

https://bugs.launchpad.net/horizon/+bug/2102214

But are yet to see any traction.


etc/kayobe/kolla/config/ironic/policy.yaml
```diff
@@ -4,4 +4,4 @@
 # GET  /nodes/detail
 # Intended scope(s): system, project
 # Overridden: added role:admin
-"baremetal:node:list_all": "role:admin or (role:reader and system_scope:all) or (role:service and system_scope:all) or rule:service_role"
+"baremetal:node:list_all": "role:baremetal_node_list_all or (role:reader and system_scope:all) or (role:service and system_scope:all) or rule:service_role"
```
bbezak, line 7: does it mean that regular admin users now won't be able to list all baremetals from cli? - I think that was idea behind this policy override. also baremetal_node_list_all don't exist by default?

jovial: > baremetal_node_list_all don't exist by default?

Correct, you'd have to create that.

> does it mean that regular admin users now won't be able to list all baremetals from cli?

Correct, unless they own all nodes (via setting the owner property on baremetal nodes). You can add the new role if you need that behaviour (but it will break your horizon).

MoteHue: Will this break Horizon, even with `SYSTEM_SCOPE_SERVICES = ['ironic']` set?

jovial: Yes, as I think system scope appears in the project drop down. So if you have a project selected (rather than system scope) and you navigate to the baremetal provisioning section, you will get a bunch of errors. You would be able to select system scope and browse to the same page though.

releasenotes/notes/fix-ironic-horizon-integration-c844d19bd36ed014.yaml
```diff
@@ -0,0 +1,11 @@
+---
+upgrade:
+  - |
+    Users with the admin role can can no longer list all baremetal nodes by
```
seunghun1ee: nit: 'can' was written twice

etc/kayobe/kolla/config/ironic/policy.yaml
JohnGarbutt: @bbezak what I think we should do is make sure the nodes are all added into a single project, e.g. admin project, so the regular project scoped admin user can still list them all, and do all the usual actions to the nodes that worked in yoga, etc.

In my head, this extra role here is only needed to help in the Slurm case, where some other member in some other project, needs to list all baremetal nodes, to convert the baremetal node name into a uuid, to pass into the scheduler hint.

releasenotes/notes/fix-ironic-horizon-integration-c844d19bd36ed014.yaml
```diff
@@ -0,0 +1,11 @@
...
+    default. This broke horizon as it tries to collect data for any nodes
+    that it can see. Please add the baremetal_node_list_all role
+    to any users to want this capability; it is not recommened since
+    Horizon will remain broken for them.
```
JohnGarbutt: I don't think that will break Horizon, right, as the member + baremetal_node_list_all user will not have the ironic tab appear in horizon, anyways.

The only remaining broken combination would be having the admin role, in a project other than the project that owns all the ironic nodes. I guess? Which I suppose you fix by adding the system scope thing?

etc/kayobe/kolla/config/ironic/policy.yaml
jovial: I think John is right. We don't break horizion for ordinary users (non-admin). It is is breaking change for users with the admin role though - so perhaps best to merge in 2026.1?

jovial:
I think we should merge this in 2026.1

PR #1819, opened 2025-08-05
Epoxy: Update Magnum CAPI Helm driver

cityofships:


Alex-Welsh:
Have we deployed this to SMS yet?

cityofships:
> Have we deployed this to SMS yet?

Not these containers - but the updated driver had lived there for a long time.

Alex-Welsh (approved):

PR #1823, opened 2025-08-11
Upstream magnum

seunghun1ee:
Our downstream Magnum [1] was for keeping Heat driver support.
But we dropped them, so there's no need to build magnum from downstream.

[1] https://github.com/stackhpc/magnum/pull/202


Alex-Welsh:
Need to be careful this doesn't conflict with https://github.com/stackhpc/stackhpc-kayobe-config/pull/1819

seunghun1ee:
I'll rebuild again to include that change

Alex-Welsh:
Now that Epoxy is stable, we should also include release notes for container image rebuilds

Alex-Welsh (approved):
If only we could test Magnum in CI 😢

PR #1831, opened 2025-08-12
docs: Add instructions for Ceph shutdown

Alex-Welsh:
Client has a full shutdown coming up, which they are handling themselves. Most of the guide is fine but the Ceph section was quite sparse.

Renderd outupt:
image

### DIFF
--- a/doc/source/operations/control-plane-operation.rst
+++ b/doc/source/operations/control-plane-operation.rst
@@ -174,8 +174,30 @@ is advisable to migrate all of the instances to another machine. See
 Ceph
 ----
 
-The following guide provides a good overview:
-https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/8/html/director_installation_and_usage/sect-rebooting-ceph
+#. Check that the cluster is healthy (i.e. ``ceph -s``). Where possible, solve
+   or isolate any issues before the shutdown e.g. by marking unhealthy OSDs as
+   'out' in the cluster.
+
+#. Stop all clients. This includes
+
+   * **All** OpenStack VMs (if their storage is RBD-backed).
+
+   * CephFS mounts.
+
+   * Ceph-backed OpenStack services such as Glance, Cinder, Manila, and RGW/S3/Swift.
+
+#. Set the ``noout`` flag, so that the cluster does not attempt to redistribute
+   data when OSDs go down. Use the following command on a MON node:
+
+   .. code-block:: console
+
+      sudo cephadm shell -- ceph osd set noout
+
+#. Shut down all the nodes, with those holding MON services last.
+
+Note that if it is not desired for Ceph services to automatically start later
+with the operating system, extra steps need to be taken and are not described
+here.
 
 Shutting down the seed VM
 -------------------------
@@ -200,6 +222,24 @@ following order:
 * Shut down Ceph nodes (if applicable)
 * Shut down seed VM
 * Shut down Ansible control host
+
+Full startup
+-------------
+
+If the entire control plane is powered down, it is best to bring the nodes up
+in the reverse order of shutdown:
+
+* Power on Ansible control host
+* Power on seed VM (and other service VMs)
+* Power on Ceph nodes (if applicable)
+   * Where possible, start the nodes running MON services first.
+   * Make sure that all OSD services are back up and running. At this point
+     it is safe to unset  the ``noout`` cluster flag.
+* Power on controllers
+* Power on network nodes (if separate from controllers)
+* Power on monitoring node (if separate from controllers)
+* Power on compute nodes
+* Power on virtual machines
 
 Rebooting a node
 ----------------


doc/source/operations/control-plane-operation.rst
```diff
@@ -174,8 +174,30 @@ is advisable to migrate all of the instances to another machine. See
...
 
-The following guide provides a good overview:
-https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/8/html/director_installation_and_usage/sect-rebooting-ceph
+#. Check that the cluster is healthy (i.e. ``ceph -s``). Where possible, solve
+   or isolate any issues before the shutdown e.g. by marking unhealthy OSDs as
+   'out' in the cluster.
+
+#. Stop all clients. This includes
+
+   * **All** OpenStack VMs (if their storage is RBD based).
```
priteau: RBD based -> RBD-backed

doc/source/operations/control-plane-operation.rst
```diff
@@ -174,8 +174,30 @@ is advisable to migrate all of the instances to another machine. See
...
+
+   * CephFS mounts.
+
+   * Ceph-backed openstack services such as Glance, Cinder, and RGW/S3/Swift.
```
priteau: OpenStack

doc/source/operations/control-plane-operation.rst
```diff
@@ -201,6 +223,23 @@ following order:
...
 
+Full startup
+-------------
+
+If the entire control plane is powered down, it is best to bring the nodes up
+in the reverse order of shutdown:
+
+* Power on Ansible control host
+* Power on seed VM (and other service VMs)
+* Power on Ceph nodes (if applicable)
```
priteau, line 234: ideally those holding MON services first?

priteau (dismissed):

doc/source/operations/control-plane-operation.rst
Alex-Welsh: Done

doc/source/operations/control-plane-operation.rst
Alex-Welsh: Done

doc/source/operations/control-plane-operation.rst
Alex-Welsh: Done

doc/source/operations/control-plane-operation.rst
```diff
@@ -174,8 +174,30 @@ is advisable to migrate all of the instances to another machine. See
...
+   or isolate any issues before the shutdown e.g. by marking unhealthy OSDs as
+   'out' in the cluster.
+
+#. Stop all clients. This includes
+
+   * **All** OpenStack VMs (if their storage is RBD-backed).
+
+   * CephFS mounts.
+
+   * Ceph-backed OpenStack services such as Glance, Cinder, and RGW/S3/Swift.
```
MoteHue:
```suggestion
   * Ceph-backed OpenStack services such as Glance, Cinder, Manila, and RGW/S3/Swift.
```

MoteHue (dismissed):

doc/source/operations/control-plane-operation.rst
```diff
@@ -201,6 +223,24 @@ following order:
...
+-------------
+
+If the entire control plane is powered down, it is best to bring the nodes up
+in the reverse order of shutdown:
+
+* Power on Ansible control host
+* Power on seed VM (and other service VMs)
+* Power on Ceph nodes (if applicable)
+   * Where possible, start the nodes running MON services first.
+   * Be sure to unset the ``noout`` flag, so that the cluster can
```
cityofships: noout is not relevant to rebalancing at all

Alex-Welsh: What does it actually do? The docs Ceph docs say:

> when you stop OSDs for maintenance, you might want to prevent CRUSH from automatically rebalancing the cluster. To avert this rebalancing behavior, set the cluster to noout

But also there's a warning after that which seems to say you don't need to do it since Luminous, so I'm confused

https://docs.ceph.com/en/reef/rados/troubleshooting/troubleshooting-osd/#stopping-without-rebalancing

MoteHue: `noout` is telling Ceph not to treat the osds as being out of the cluster. This I suppose indirectly stops the rebalancing that would otherwise occur when an osd is out. 

There's also a `norebalance` flag that explicitly stops rebalancing, I don't know if we'd _need_ to set both, but I usually do so. 

@cityofships what would you suggest?

cityofships:
```suggestion
   * Make sure that all OSD services are back up and running. At this point
     it is safe to unset  the ``noout`` cluster flag.
```

cityofships: `norebalance` wouldn't hurt but `noout` is a must. Without `norebalance`, a couple of placement groups might end up being remapped following service stop / start and that's no issue. A small _rebalancing_ traffic then initiates and finishes naturally. With `noout` you're indicating cluster not to start a _recovery_ traffic which would be huge in case of multiple servers powered off and unnecessary.

Alex-Welsh: Many thanks @cityofships. I've applied your suggestion

cityofships (approved):

PR #1845, opened 2025-08-20
feat: add `security` stack

jackhodgkiss:
Add a new `stack` for managing the security baseline expected of new and existing deployments.

[skip ci]

### DIFF
--- a/doc/source/configuration/stackhpc-mixin-environments.rst
+++ b/doc/source/configuration/stackhpc-mixin-environments.rst
@@ -33,3 +33,10 @@ baremetal-policy
 ----------------
 
 .. include:: ../../../etc/kayobe/environments/baremetal-policy/README.rst
+
+.. _mixin-security:
+
+security
+--------
+
+.. include:: ../../../etc/kayobe/environments/security/README.rst
--- a/etc/kayobe/environments/security/README.rst
+++ b/etc/kayobe/environments/security/README.rst
@@ -1,0 +1,96 @@
+Security stack
+~~~~~~~~~~~~~~
+
+Mixin environment that applies a security baseline to new and existing
+deployments. It bundles several security-related configuration modules into a
+single, opinionated environment:
+
+* **Firewall** — firewalld is enabled on all host types (controllers, compute,
+  storage, monitoring, infrastructure VMs, seed, and seed hypervisor) using the
+  standardised StackHPC firewalld zones and rules. Kolla Ansible is configured
+  to open ports in firewalld for services on the public API network. See
+  :ref:`firewall` for details.
+
+* **CIS benchmark hardening** — the
+  ``stackhpc_enable_cis_benchmark_hardening_hook`` flag is set to ``true``,
+  which means the CIS hardening playbooks run automatically as part of
+  ``kayobe * host configure``. See :doc:`security-hardening` for details.
+
+* **Walled garden** — a Squid caching proxy is enabled on the seed, and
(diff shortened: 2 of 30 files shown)


etc/kayobe/environments/security/compute.yml
```diff
@@ -0,0 +1,23 @@
...
+compute_firewalld_enabled: true
+
+# A list of zones to create. Each item is a dict containing a 'zone' item.
+compute_firewalld_zones: "{{ stackhpc_firewalld_zones }}"
+
+# A firewalld zone to set as the default. Default is unset, in which case
+# the default zone will not be changed.
+# Predefined zones are listed here:
+# https://firewalld.org/documentation/zone/predefined-zones.html
+compute_firewalld_default_zone: trusted
```
Alex-Welsh, line 15: We should ideally default to `drop` here, though it does make it trickier to apply

jackhodgkiss: I had just taken everything from the docs.

Need to figure out how to manage networks and their zones within this setup.

Alex-Welsh: We should probably update those docs. I don't think that the firewall will really do anything if the default zone is trusted. Basically that just means if there's no specific rule, it just passes, which is pretty insecure.

The easiest way to manage this right now is to just set `*_firewalld_default_zone: "{{ stackhpc_firewalld_default_zone }}"`

Then I don't mind whether we set `stackhpc_firewalld_default_zone` to trusted or drop, as long as we include a decent comment next to it explaining what it does

Alex-Welsh: Actually, I might be wrong. I need to dig into this more

Alex-Welsh:
@jackhodgkiss I've removed the `[skip-ci]` in the commit message. I'll break because the required checks never run

Alex-Welsh (changes requested):
Code changes look good to me, just needs a README, docs, and a reno so people can start using it

etc/kayobe/environments/security/README.rst
```diff
@@ -0,0 +1,96 @@
...
+   kayobe seed host configure -t network,firewall
+   kayobe infra vm host configure -t network,firewall
+   kayobe overcloud host configure -t network,firewall
+
+.. note::
+
+   Applying the firewall for the first time carries a risk of locking yourself
+   out of hosts. Read the safety guidance in :ref:`firewall` — in particular
+   the sections on using the ``firewalld-watchdog.yml`` playbook and applying
+   controller changes one at a time — before proceeding.
```
Alex-Welsh, line 85: You may also want to mention here that deploying the firewall config can cause issues on a running cloud, because kayobe sets up the firewall rules, but kolla ansible won't configure it so services can run until a service deploy is run. I think @wtripp180901 hit this recently, and might be able to share some more.

doc/source/configuration/stackhpc-mixin-environments.rst
```diff
@@ -33,3 +33,10 @@ baremetal-policy
 ----------------
 
 .. include:: ../../../etc/kayobe/environments/baremetal-policy/README.rst
+
+.. _mixin-security:
```
Alex-Welsh, line 37: Unused ref?

etc/kayobe/environments/security/compute.yml
Alex-Welsh: I'm un-resolving this thread, because I'm still not convinced that `trusted` is a good approach here. Let's discuss this later this week

PR #1848, opened 2025-08-22
Fix Grafana project metrics dashboard [2025.1]

Alex-Welsh:
See also https://github.com/stackhpc/stackhpc-kayobe-config/pull/1843


MoteHue (approved):
Works in my AIO< cheers @Alex-Welsh

PR #1869, opened 2025-09-08
Gather cloud factsv2

GavinHeff:
Adds playbook to gather more advanced facts about a multinode cloud

### DIFF
--- a/etc/kayobe/ansible/filter_plugins/filters.py
+++ b/etc/kayobe/ansible/filter_plugins/filters.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python3
+
+from ansible.errors import AnsibleFilterError
+
+class FilterModule(object):
+    def filters(self):
+        return {
+            'group_hostvars_by_var':
+            self.group_hostvars_by_var,
+            'get_hostvars_by_host':
+            self.get_hostvars_by_host
+        }
+
+    def group_hostvars_by_var(self, hostvars, var, subkey=None):
+        """
+        Returns a dictionary where the keys are values for the
+        specified var in hostvars, and the values are the hosts
+        that match that value.
+
+        For example, a grouping of hosts by OS release might look like:
+        distribution_release:
+          noble:
+            - node1
+            - node2
+          jammy:
+            - node3
+            - node4
+            - node5
+
+        Some Ansible commands, such as ansible.builtin.command, return a
+        dict rather than a single value. So 'subkey' is used for these cases
+        to access the desired value.
+        """
+        result = {}
+
+        for host in hostvars.keys():
+            try:
+                indiv_host_var = hostvars[host][var]
+                if subkey is not None:
+                    indiv_host_var = indiv_host_var[subkey]
+                result.setdefault(indiv_host_var, []).append(host)
+            except KeyError as e:
+                raise AnsibleFilterError(f"Variable {var} not found for host {host} in hostvars: {e}")
+
+        return result
+
+    def get_hostvars_by_host(self, hostvars, var, subkey=None):
+        """
+        Returns a dictionary where the keys are hosts and the values
+        are the values for the specified var in hostvars.
+
+        For example, the deployed containers by host might look like:
+        deployed_containers:
+            node1:
+              - grafana
+              - glance
+              - nova
+              - prometheus
+            node2:
+              - designate
+              - neutron
+              - nova
+        
+        Some Ansible commands, such as ansible.builtin.command, return a
+        dict rather than a single value. So 'subkey' is used for these cases
+        to access the desired value.
+        """
+        result = {}
+        for host in hostvars.keys():
+            try:
+                indiv_host_var = hostvars[host][var]
+                for key in indiv_host_var.keys():
+                    # Check if task to assign value was skipped
+                    if key == "skipped":
+                        result[host] = "No data"
+                        continue
+                if subkey is not None:
+                    indiv_host_var = indiv_host_var[subkey]
+                if indiv_host_var:
+                    result[host] = indiv_host_var
+                else:
+                    result[host] = []
+
+            except KeyError as e:
+                raise AnsibleFilterError(f"Variable {var} not found for host {host} in hostvars: {e}")
+
+        return result
--- a/etc/kayobe/ansible/tools/get-cloud-facts.yml
+++ b/etc/kayobe/ansible/tools/get-cloud-facts.yml
@@ -1,13 +1,68 @@
 ---
-- name: Gather Cloud Facts
+- name: Set facts for control host
+  hosts: localhost
+  gather_facts: true
+  tasks:
+    - name: Check if ngs enabled
+      ansible.builtin.set_fact:
+        ngs_enabled: >-
+          {{ 'genericswitch' in kolla_neutron_ml2_mechanism_drivers | default(false) }}
+      when: kolla_neutron_ml2_mechanism_drivers
+
+- name: Set facts for seed node
+  hosts: seed
+  gather_facts: true
+  tasks:
+    - name: Is there a dedicated seed node?
+      ansible.builtin.set_fact:
+        seed_node_is_vm: "{{ ansible_facts.virtualization_role == 'guest' }}"
+
+- name: Set facts for all hosts
+  hosts: all
+  gather_facts: true
+  tasks:
+    - name: Pinging stackhpc.com to check internet connectivity
+      ansible.builtin.command: "ping stackhpc.com -c 3"
+      register: successful_ping
+      timeout: 5
+      ignore_errors: true
+
+    - name: Get kernel version
+      ansible.builtin.command: "uname -r"
+      register: kernel_version
+
+    - name: Check if Docker is installed
+      ansible.builtin.command: "which docker"
+      register: docker_check
+      failed_when: docker_check.rc not in [0]
+      ignore_errors: true
+
+    - name: Get running contianers
+      ansible.builtin.command: "docker ps --format {% raw %}'{{.Names}}'{% endraw %}"
+      become: true
+      register: containers_list
+      when: not docker_check.failed
+
+    - name: Set facts
+      ansible.builtin.set_fact:
+        internet_connectivity: "{{ not successful_ping.failed }}"
+        distribution: "{{ ansible_facts.distribution }}"
+        distribution_release: "{{ ansible_facts.distribution_release }}"
+        hugepages_enabled: "{{ 'hugepages' in ansible_facts.cmdline }}"
+
+- name: Write cloud facts file
   hosts: localhost
   gather_facts: true
   tasks:
     - name: Write facts to file
       vars:
         cloud_facts:
-          ansible_control_host_distribution: "{{ ansible_facts.distribution }}"
-          ansible_control_host_distribution_release: "{{ ansible_facts.distribution_release }}"
+          internet_connectivity: "{{ hostvars | group_hostvars_by_var('internet_connectivity') }}"
+          seed_node_is_vm: "{{ hostvars[groups['seed'][0]]['seed_node_is_vm'] | default('N/A') if groups ['seed'] else 'N/A' }}"
+          distributions: "{{ hostvars | group_hostvars_by_var('distribution') }}"
+          dist_releases: "{{ hostvars | group_hostvars_by_var('distribution_release') }}"
+          kernel_versions: "{{ hostvars | group_hostvars_by_var('kernel_version','stdout') }}"
+          deployed_containers: "{{ hostvars | get_hostvars_by_host('containers_list', 'stdout_lines') | default({}) }}"
           openstack_release: "{{ openstack_release }}"
           openstack_release_name: "{{ openstack_release_codename }}"
           ansible_control_host_is_vm: "{{ ansible_facts.virtualization_role == 'guest' }}"
@@ -22,41 +77,10 @@
           ceph_release: "{{ cephadm_ceph_release }}"
           storage_hyperconverged: "{{ groups['controllers'] | intersect(groups['osds']) | length > 0 | bool }}"
           wazuh_enabled: "{{ groups['wazuh-agent'] | length > 0 | bool }}"
+          ngs_enabled: "{{ ngs_enabled | default(false) }}"
           kayobe_managed_switches: "{{ groups['switches'] | length > 0 | bool }}"
           proxy_configured: "{{ http_proxy | bool or https_proxy | bool }}"
           bifrost_version: "{{ kolla_bifrost_source_version }}"
-          barbican_enabled: "{{ kolla_enable_barbican }}"
-          nova_enabled: "{{ kolla_enable_nova }}"
-          neutron_enabled: "{{ kolla_enable_neutron }}"
-          ovs_enabled: "{{ kolla_enable_openvswitch }}"
-          ovn_enabled: "{{ kolla_enable_ovn }}"
-          glance_enabled: "{{ kolla_enable_glance }}"
-          cinder_enabled: "{{ kolla_enable_cinder }}"
-          keystone_enabled: "{{ kolla_enable_keystone }}"
-          horizon_enabled: "{{ kolla_enable_horizon }}"
-          fluentd_enabled: "{{ kolla_enable_fluentd }}"
-          rabbitmq_enabled: "{{ kolla_enable_rabbitmq }}"
-          mariadb_enabled: "{{ kolla_enable_mariadb }}"
-          mariabackup_enabled: "{{ kolla_enable_mariabackup }}"
-          memcached_enabled: "{{ kolla_enable_memcached }}"
-          haproxy_enabled: "{{ kolla_enable_haproxy }}"
-          keepalived_enabled: "{{ kolla_enable_keepalived }}"
-          octavia_enabled: "{{ kolla_enable_octavia }}"
-          designate_enabled: "{{ kolla_enable_designate }}"
-          manila_enabled: "{{ kolla_enable_manila }}"
-          magnum_enabled: "{{ kolla_enable_magnum }}"
-          heat_enabled: "{{ kolla_enable_heat }}"
-          ironic_enabled: "{{ kolla_enable_ironic }}"
-          skyline_enabled: "{{ kolla_enable_skyline }}"
-          blazar_enabled: "{{ kolla_enable_blazar }}"
-          pulp_enabled: "{{ seed_pulp_container_enabled }}"
-          opensearch_enabled: "{{ kolla_enable_opensearch }}"
-          opensearch_dashboards_enabled: "{{ kolla_enable_opensearch_dashboards }}"
-          influxdb_enabled: "{{ kolla_enable_influxdb }}"
-          grafana_enabled: "{{ kolla_enable_grafana }}"
-          prometheus_enabled: "{{ kolla_enable_prometheus }}"
-          cloudkitty_enabled: "{{ kolla_enable_cloudkitty }}"
-          telegraf_enabled: "{{ kolla_enable_telegraf }}"
           internal_tls_enabled: "{{ kolla_enable_tls_internal }}"
           external_tls_enabled: "{{ kolla_enable_tls_external }}"
           firewalld_enabled_all: >-
@@ -82,6 +106,9 @@
           stackhpc_package_repos_enabled: "{{ stackhpc_repos_enabled }}"
           pulp_tls_enabled: "{{ pulp_enable_tls }}"
           kolla_image_tags: "{{ kolla_image_tags }}"
+          gpu_passthrough_enabled: "{{ gpu_group_map | length > 0 | bool }}"
+          vgpu_enabled: "{{ groups['vgpu'] | length > 0 | bool }}"
+          hugepages_enabled: "{{ hostvars | group_hostvars_by_var('hugepages_enabled') }}"
       ansible.builtin.copy:
         content: "{{ cloud_facts | to_nice_json(sort_keys=false) }}"
         dest: ~/cloud-facts.json
--- a/etc/kayobe/ansible/tools/get-cloud-facts.yml
+++ b/etc/kayobe/ansible/tools/get-cloud-facts.yml
@@ -58,10 +58,10 @@
       vars:
         cloud_facts:
           internet_connectivity: "{{ hostvars | group_hostvars_by_var('internet_connectivity') }}"
-          seed_node_is_vm: "{{ hostvars[groups['seed'][0]]['seed_node_is_vm'] | default('N/A') if groups ['seed'] else 'N/A' }}"
+          seed_node_is_vm: "{{ hostvars[groups['seed'][0]]['seed_node_is_vm'] | default('N/A') if groups['seed'] else 'N/A' }}"
           distributions: "{{ hostvars | group_hostvars_by_var('distribution') }}"
           dist_releases: "{{ hostvars | group_hostvars_by_var('distribution_release') }}"
-          kernel_versions: "{{ hostvars | group_hostvars_by_var('kernel_version','stdout') }}"
+          kernel_versions: "{{ hostvars | group_hostvars_by_var('kernel_version', 'stdout') }}"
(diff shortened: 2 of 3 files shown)


etc/kayobe/ansible/get-cloud-facts-v2.yml
```diff
@@ -0,0 +1,114 @@
...
+  tasks:
+    - name: Write facts to file
+      vars:
+        cloud_facts:
+          internet_connectivity: "{{ internet_connectivity }}"
+          seed_node_is_vm: "{{ hostvars[groups['seed'][0]]['seed_node_is_vm'] }}"
+          distributions: "{{ hostvars | group_hostvars_by_var('distribution') }}"
+          dist_releases: "{{ hostvars | group_hostvars_by_var('distribution_release') }}"
+          kernel_versions: "{{ hostvars | group_hostvars_by_var('kernel_version','stdout') }}"
+          deployed_containers: "{{ hostvars | group_hostvars_by_host('containers_list', 'stdout') }}"
```
Alex-Welsh: group_hostvars_by_host and group_hostvars_by_var are the new filters right? You appear to be missing it in this PR

etc/kayobe/stackhpc.yml
```diff
@@ -165,11 +165,11 @@ stackhpc_repo_elrepo_9_version: "{{ stackhpc_repo_distribution }}"
 
 # Kolla source repository.
 stackhpc_kolla_source_url: "https://github.com/stackhpc/kolla"
-stackhpc_kolla_source_version: stackhpc/18.7.0.2
+stackhpc_kolla_source_version: stackhpc/18.7.0.4
```
Alex-Welsh: These seem to be an accidental change.

You should be able to fix them by rebasing on stackhpc/2025.1

etc/kayobe/ansible/get-cloud-facts-v2.yml
GavinHeff: Looks like I might not have added those files to the commit and have since torn down the multinode I was using to work on this :cry:

GavinHeff: Re-added the Python filters

tools/scan-images.sh
Alex-Welsh, line 1: It appears to have picked up some stray changes from another PR. You might need another rebase

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
+#!usr/bin/env python3
+
+from ansible.errors import AnsibleFilterError
+# from ansible.plugins.filter.core imprt FilterModule
```
Alex-Welsh: Is this required?

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -22,41 +77,11 @@
           ceph_release: "{{ cephadm_ceph_release }}"
           storage_hyperconverged: "{{ groups['controllers'] | intersect(groups['osds']) | length > 0 | bool }}"
           wazuh_enabled: "{{ groups['wazuh-agent'] | length > 0 | bool }}"
+          # ngs_enabled_: "{{ 'genericswitch' in (kolla_neutron_ml2_mechanism_drivers | default([])) | bool }}"
+          # # ABOVE LEADS TO ERROR: argument of type ''NoneType'' is not iterable. argument of type ''NoneType'' is not iterable'
           kayobe_managed_switches: "{{ groups['switches'] | length > 0 | bool }}"
```
Alex-Welsh: I'd guess it's only defined if it's used, so could you add a check to first ensure `kolla_neutron_ml2_mechanism_drivers` is defined, and if it is, check the values it contains?

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -1,13 +1,68 @@
...
+  tasks:
+    - name: Check internet connectivity
+      ansible.builtin.command: "ping stackhpc.com -c 3"
+      register: internet_connectivity
+      timeout: 5
+      ignore_errors: true
+
+    - name: Set internet connectivity fact
+      ansible.builtin.set_fact:
+        internet_connectivity: "{{ not internet_connectivity.failed }}"
```
Alex-Welsh: Two things to note here:

1. it's probably useful to know this fact for all the hosts, not just localhost, so I'd move it to the "All" section later.
2. It's not _just_ checking internet connectivity, it's also resolving the DNS entry for stackhpc.com, and it's ICMP (ping) in particular. Could it be rephrased, so it's more specific about what it's doing? i.e. "pinging stackhpc.com"

Alex-Welsh: If you want to extend it further, you could also add a `curl` command in, to check other protocols, ping an IP address (to sidestep DNS), or ping something that isn't `stackhpc.com`

etc/kayobe/ansible/tools/get-cloud-facts.yml
```diff
@@ -1,13 +1,68 @@
...
+    - name: Get OS distribution release
+      ansible.builtin.set_fact:
+        distribution_release: "{{ ansible_facts.distribution_release }}"
+
+    - name: Get kernel version
+      ansible.builtin.command: "uname -r"
+      register: kernel_version
+
+    - name: Get running contianers
+      ansible.builtin.command: "docker ps --format {% raw %}'{{.Names}}'{% endraw %}"
```
Alex-Welsh, line 41: This could fail if it runs on a host that doesn't have docker installed, so needs to account for that

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -82,6 +107,9 @@
           stackhpc_package_repos_enabled: "{{ stackhpc_repos_enabled }}"
           pulp_tls_enabled: "{{ pulp_enable_tls }}"
           kolla_image_tags: "{{ kolla_image_tags }}"
+          gpu_passthrough_enabled: "{{ gpu_group_map | length > 0 | bool }}"
+          vGPU_enabled: "{{ groups['vgpu'] | length > 0 | bool }}"
+          hugepages_enabled: "{{ hostvars | group_hostvars_by_host('hugepages_enabled', hosts=['compute']) }}"
       ansible.builtin.copy:
         content: "{{ cloud_facts | to_nice_json(sort_keys=false) }}"
-        dest: ~/cloud-facts.json
+        dest: ~/cloud-facts_2.json
```
Alex-Welsh: Lets just keep this as cloud-facts.json

Alex-Welsh (changes requested):

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+class FilterModule(object):
+    def filters(self):
+        return {
+            'group_hostvars_by_var':
+            self.group_hostvars_by_var,
+            'group_hostvars_by_host':
+            self.group_hostvars_by_host
+        }
+
+    def group_hostvars_by_var(self, hostvars, var, stdout=None):
```
Alex-Welsh: Could you add [docstrings](https://peps.python.org/pep-0257/#multi-line-docstrings) for these filters?

etc/kayobe/ansible/filter_plugins/filters.py
GavinHeff: FilterModule isn't but AnsibleFilterError allows you to throw an error in the ansible output if it hits an issue with the filter

etc/kayobe/ansible/get-cloud-facts.yml
GavinHeff: Made first change, might add the curl step later

etc/kayobe/ansible/tools/get-cloud-facts.yml
GavinHeff: How would I go about that?

Alex-Welsh: I'd suggest another task before this one, which just runs something like "docker --version", register the output and set failed_when: false, then add a condition to this task so it only runs if the last one succeeded

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        Returns a dictionary where the keys are values for the
+        specified var in hostvars, and the values are the hosts
+        that match that value.
+        """
```
Alex-Welsh, line 33: Would benefit a lot from an example e.g:
```suggestion
        """
        Returns a dictionary where the keys are values for the
        specified var in hostvars, and the values are a list of hosts
        that match that value.
        For example, a grouping of hosts by OS release might look like:
        example_var:
          noble:
            - node1
            - node2
          jammy:
            - node3
            - node4
            - node5
        
        """
```

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+        result = {}
+
+        for host in hostvars.keys():
+            try:
+                indiv_host_all = hostvars[host]
+                indiv_host_var = indiv_host_all[var]
```
Alex-Welsh: Can be done in one jump
```suggestion
                indiv_host_var = hostvars[host][var]
```

etc/kayobe/ansible/filter_plugins/filters.py
Alex-Welsh: Would also help to include a description of "stdout"

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+                if stdout is not None:
+                    indiv_host_var = indiv_host_var[stdout]
```
Alex-Welsh: I think the name of the `stdout` variable is a bit deceptive, since it can be used in a much more generic way than just for stdout. I'd change the name to something like `key` or `sub_key` etc.

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+                result.setdefault(indiv_host_var, []).append(host)
+            except KeyError as e:
+                raise AnsibleFilterError(f"Variable {var} not found for host {host} in hostvars: {e}")
+
+        return result
+
+    def group_hostvars_by_host(self, hostvars, var, stdout=None, hostgroup=None):
+        result = {}
```
Alex-Welsh: Move below docstring

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        Returns a dictionary where the keys are hosts and the values
+        are the values for the specified var in hostvars. If hosts is
```
Alex-Welsh:
```suggestion
        are the values for the specified var in hostvars. If hostgroup is
```

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+                indiv_host_var = indiv_host_all[var]
+                if stdout is not None:
+                    indiv_host_var = indiv_host_var[stdout]
+                result.setdefault(indiv_host_var, []).append(host)
+            except KeyError as e:
+                raise AnsibleFilterError(f"Variable {var} not found for host {host} in hostvars: {e}")
+
+        return result
+
+    def group_hostvars_by_host(self, hostvars, var, stdout=None, hostgroup=None):
```
Alex-Welsh: hostgroup is defined but not used

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+        Returns a dictionary where the keys are hosts and the values
+        are the values for the specified var in hostvars. If hosts is
+        specified, only those hosts are included.
+        """
+        for host in hostvars.keys():
+            try:
+                indiv_host_vars = hostvars[host][var]
+                for key in indiv_host_vars.keys():
+                    if key == "skipped":
+                        return {}
```
Alex-Welsh: I think `return` here is just going to instantly exit the function completely and return an empty dictionary. I think you want `continue`. May have to refactor the loop a bit to get that to work properly though.

Alex-Welsh: Just to be clear, what's the intention with this bit of the loop?

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+                if stdout is not None:
```
Alex-Welsh: Same as before, it'd be best to rename `stdout` to something new

etc/kayobe/ansible/filter_plugins/filters.py
```diff
@@ -0,0 +1,57 @@
...
+    def group_hostvars_by_host(self, hostvars, var, stdout=None, hostgroup=None):
+        result = {}
+        """
+        Returns a dictionary where the keys are hosts and the values
+        are the values for the specified var in hostvars. If hosts is
+        specified, only those hosts are included.
+        """
+        for host in hostvars.keys():
+            try:
+                indiv_host_vars = hostvars[host][var]
```
Alex-Welsh: You're selecting a single variable here, so probably best to use the singular
```suggestion
                indiv_host_var = hostvars[host][var]
```

etc/kayobe/ansible/filter_plugins/filters.py
Alex-Welsh: I'm not sure it's really doing any "grouping" here, what do you think of renaming the function to something like `get_hostvars_by_host`?

Alex-Welsh (changes requested):

etc/kayobe/ansible/filter_plugins/filters.py
GavinHeff: Think trying to specify particular hosts wasn't working so I got rid of it but forgot the argument

etc/kayobe/ansible/filter_plugins/filters.py
GavinHeff: Its a way to handle hosts not having Docker installed (or any other service it might be checking). If the docker ps command is skipped then Ansible saves "skipped" to containers_list. This part of the loop catches that and returns an empty dict as the value for that host

Alex-Welsh: Ah, okay have you tested it? I'm concerned that `return` is going to exit the entire function and just ignore any future iterations of the loop

Alex-Welsh: Here's a quick example of what I mean:

```python

def testfunc():
    for i in range(10):
        if not i == 3:
            print(i)
        else:
            return i

    return "complete"

output = testfunc()

print("Final output")
print(output)
```
```
alex@alex-framework:/tmp$ python3 testo.py 
0
1
2
Final output
3
```
The loop never gets past 3 because it returns too early

Alex-Welsh: Whereas `continue` will just skip the element:
```
alex@alex-framework:/tmp$ python3 testo.py 
0
1
2
4
5
6
7
8
9
Final output
complete
```

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -0,0 +1,124 @@
...
+
+- name: Get facts for seed node
+  hosts: seed
+  gather_facts: true
+  tasks:
+    - name: Is there a dedicated seed node?
+      ansible.builtin.set_fact:
+        seed_node_is_vm: "{{ ansible_facts.virtualization_role == 'guest' }}"
+
+- name: Get facts
```
Alex-Welsh: Just to be extra clear

```suggestion
- name: Set facts for all hosts
```

etc/kayobe/ansible/tools/get-cloud-facts.yml
```diff
@@ -0,0 +1,124 @@
...
+  gather_facts: true
+  tasks:
+    - name: Is there a dedicated seed node?
+      ansible.builtin.set_fact:
+        seed_node_is_vm: "{{ ansible_facts.virtualization_role == 'guest' }}"
+
+- name: Get facts
+  hosts: all
+  gather_facts: true
+  tasks:
```
Alex-Welsh, line 23: Needs a pass on spacing between these tasks. Sometimes there's no gap, sometimes there's two lines.
Should be one empty line between each task (no gap between the first task and the `tasks:` line)

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -0,0 +1,124 @@
...
+      ansible.builtin.set_fact:
+        internet_connectivity: "{{ not successful_ping.failed }}"
+    - name: Get OS distribution
+      ansible.builtin.set_fact:
+        distribution: "{{ ansible_facts.distribution }}"
+
+
+    - name: Get OS distribution release
+      ansible.builtin.set_fact:
+        distribution_release: "{{ ansible_facts.distribution_release }}"
```
Alex-Welsh: I went looking for ways to deduplicate this and it turns out you can specify multiple facts in one task, see [docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/set_fact_module.html)

So you can do:
```suggestion
    - name: Set facts
      ansible.builtin.set_fact:
        internet_connectivity: "{{ not successful_ping.failed }}"
        distribution: "{{ ansible_facts.distribution }}"
        distribution_release: "{{ ansible_facts.distribution_release }}"
        hugepages_enabled: "{{ 'hugepages' in ansible_facts.cmdline }}"
```
(note I've also added `hugepages_enabled` to my example)

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -0,0 +1,124 @@
...
+      ansible.builtin.command: "docker ps --format {% raw %}'{{.Names}}'{% endraw %}"
+      become: true
+      register: containers_list
+      when: not docker_check.failed
+
+    - name: Check if hugepages enabled
+      ansible.builtin.set_fact:
+        hugepages_enabled: "{{ 'hugepages' in ansible_facts.cmdline }}"
+
+- name: Gather Cloud Facts
```
Alex-Welsh: The name of this play is now a bit similar to the previous "gather facts" plays. I'd rename it to be a bit more clear 
```suggestion
- name: Write cloud facts file
```

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -0,0 +1,124 @@
...
+              monitoring_firewalld_enabled or
+              infra_vm_firewalld_enabled or
+              seed_firewalld_enabled or
+              seed_hypervisor_firewalld_enabled
+            }}
+          stackhpc_package_repos_enabled: "{{ stackhpc_repos_enabled }}"
+          pulp_tls_enabled: "{{ pulp_enable_tls }}"
+          kolla_image_tags: "{{ kolla_image_tags }}"
+          gpu_passthrough_enabled: "{{ gpu_group_map | length > 0 | bool }}"
+          vGPU_enabled: "{{ groups['vgpu'] | length > 0 | bool }}"
```
Alex-Welsh: Just by convention, lets keep all vars lowercase
```suggestion
          vgpu_enabled: "{{ groups['vgpu'] | length > 0 | bool }}"
```

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -0,0 +1,124 @@
+---
+- name: Get facts for control host
```
Alex-Welsh: nit: `Set facts` rather than `Get facts`
```suggestion
- name: Set facts for control host
```
Applies to subsequent plays as well

etc/kayobe/ansible/get-cloud-facts.yml
```diff
@@ -0,0 +1,124 @@
...
+    - name: Get OS distribution release
+      ansible.builtin.set_fact:
+        distribution_release: "{{ ansible_facts.distribution_release }}"
+
+
+    - name: Get kernel version
+      ansible.builtin.command: "uname -r"
+      register: kernel_version
+
+    - name: Check if docker is installed
```
Alex-Welsh: nit: capitalisation
```suggestion
    - name: Check if Docker is installed
```

Alex-Welsh (changes requested):
Looking good, just a few minor tweaks

Alex-Welsh:
You'll also need to rebase this on stackhpc/2025.1. In a recent change, all the ansible playbooks were moved to separate sub-directories, so there's a conflict where the playbook moved

etc/kayobe/ansible/tools/get-cloud-facts.yml
GavinHeff: Done

etc/kayobe/ansible/get-cloud-facts.yml
GavinHeff: This looks a lot tidier thanks

etc/kayobe/ansible/tools/get-cloud-facts.yml
```diff
@@ -22,41 +77,10 @@
...
-          skyline_enabled: "{{ kolla_enable_skyline }}"
-          blazar_enabled: "{{ kolla_enable_blazar }}"
-          pulp_enabled: "{{ seed_pulp_container_enabled }}"
-          opensearch_enabled: "{{ kolla_enable_opensearch }}"
-          opensearch_dashboards_enabled: "{{ kolla_enable_opensearch_dashboards }}"
-          influxdb_enabled: "{{ kolla_enable_influxdb }}"
-          grafana_enabled: "{{ kolla_enable_grafana }}"
-          prometheus_enabled: "{{ kolla_enable_prometheus }}"
-          cloudkitty_enabled: "{{ kolla_enable_cloudkitty }}"
-          telegraf_enabled: "{{ kolla_enable_telegraf }}"
```
Alex-Welsh, line 59: Was removing all of these intentional?

etc/kayobe/ansible/tools/get-cloud-facts.yml
```diff
@@ -1,13 +1,68 @@
...
       vars:
         cloud_facts:
-          ansible_control_host_distribution: "{{ ansible_facts.distribution }}"
-          ansible_control_host_distribution_release: "{{ ansible_facts.distribution_release }}"
+          internet_connectivity: "{{ hostvars | group_hostvars_by_var('internet_connectivity') }}"
+          seed_node_is_vm: "{{ hostvars[groups['seed'][0]]['seed_node_is_vm'] | default('N/A') if groups ['seed'] else 'N/A' }}"
+          distributions: "{{ hostvars | group_hostvars_by_var('distribution') }}"
+          dist_releases: "{{ hostvars | group_hostvars_by_var('distribution_release') }}"
+          kernel_versions: "{{ hostvars | group_hostvars_by_var('kernel_version','stdout') }}"
+          deployed_containers: "{{ hostvars | get_hostvars_by_host('containers_list', 'stdout_lines') | default({}) }}"
```
Alex-Welsh: Couple of minor spacing issues here (see linter warnings)
```suggestion
          seed_node_is_vm: "{{ hostvars[groups['seed'][0]]['seed_node_is_vm'] | default('N/A') if groups['seed'] else 'N/A' }}"
          distributions: "{{ hostvars | group_hostvars_by_var('distribution') }}"
          dist_releases: "{{ hostvars | group_hostvars_by_var('distribution_release') }}"
          kernel_versions: "{{ hostvars | group_hostvars_by_var('kernel_version', 'stdout') }}"
          deployed_containers: "{{ hostvars | get_hostvars_by_host('containers_list', 'stdout_lines') | default({}) }}"
```

Alex-Welsh (changes requested):
Nearly there

etc/kayobe/ansible/tools/get-cloud-facts.yml
GavinHeff: I think so as it should be captured by the Get running containers play

Alex-Welsh (approved):

Alex-Welsh:
recheck

PR #1497, opened 2025-02-06
Add DiB variables to allow platform architecture choice

MaxBed4d:

PR #1874, opened 2025-09-09
Periodic host image build

MaxBed4d:
This relies on the merging of [PR#1497](https://github.com/stackhpc/stackhpc-kayobe-config/pull/1497) before this one can be merged!


Alex-Welsh:
@MaxBed4d I don't really see the benefit in running these builds periodically. Why do you want to add this?

MaxBed4d:
> @MaxBed4d I don't really see the benefit in running these builds periodically. Why do you want to add this?

This was a request from @bbezak, who may be able to provide you with more info.

Originally this was supposed to be included as part of the release train workflow, however, due to the complexity of reusing a workflow from a different repository with different sets of secrets, it made more sense to mirror the other periodic tasks in this repo.

Alex-Welsh:
Dependent PR closed, effort required to get this up to date is not worth the benefit

PR #1886, opened 2025-09-15
Fixes for installation of openbao for automation

m-bull:
Some variables needed updating in the move to openbao from Vault.

### DIFF
--- a/etc/kayobe/ansible/deploy-openbao-kayobe-automation.yml
+++ b/etc/kayobe/ansible/deploy-openbao-kayobe-automation.yml
@@ -36,15 +36,16 @@
       ansible.builtin.import_role:
         name: stackhpc.hashicorp.openbao
       vars:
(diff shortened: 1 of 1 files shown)


etc/kayobe/ansible/deploy-openbao-kayobe-automation.yml
```diff
@@ -36,23 +36,24 @@
       ansible.builtin.import_role:
         name: stackhpc.hashicorp.openbao
       vars:
-        openbao_config_dir: "/opt/kayobe/vault"
+        openbao_api_addr: "http://127.0.0.1:8200"
```
holliefae, line 39: Should we be setting this to localhost? I had to set this to the admin network for it to be reachable for the runner to get secrets.

Following on from that, I needed to include something like `openbao_bind_addr: "{{ admin_oc_net_ips[inventory_hostname] }}"` in the OpenBao role vars to bind it to that interface too.

PR #1897, opened 2025-09-17
feat: add support for `A30` and `H100L`

jackhodgkiss:


etc/kayobe/stackhpc-compute.yml
```diff
@@ -101,3 +101,15 @@ stackhpc_gpu_data:
...
     product_id: "26b9"
     device_type: "type-PF"
+  # Nvidia A30
+  a30:
+    resource_name: "{{ a30_resource_name | default('a30') }}"
+    vendor_id: "10de"
+    product_id: "20b7"
+    device_type: "type-PF"
+  # Nvidia H100L
+  h100l:
```
Alex-Welsh, line 111: Should this be `h100_nvl` rather than `h100l`? Googling "nvidia h100l" doesn't really come up with anything so maybe `h100l` is a customer-specific name

jackhodgkiss: It is what it says in Linux and on the BMC.

```
sudo lspci -vvvknn | grep -i NVIDIA -A 10
c2:00.0 3D controller [0302]: NVIDIA Corporation GH100 [H100L 94GB] [10de:2321] (rev a1)
```

Alex-Welsh (approved):

etc/kayobe/stackhpc-compute.yml
Alex-Welsh: I'd prefer the _nvl suffix just because it's the more common name, but that's just personal preference

priteau (approved):

PR #1903, opened 2025-09-19
Add container image scanning script

Alex-Welsh:
Added `tools/wazuh-scan-images.sh` to scan all the container images on a host. The script will be used in the future to scan images on a schedule using Wazuh.

The script first checks if an SBOM exists for the image, if not it'll generate one and scan that rather than the image directly. That makes it much quicker to re-scan images

See also https://wazuh.com/blog/container-image-security-with-wazuh-and-trivy/

### DIFF
--- a/tools/wazuh-scan-images.sh
+++ b/tools/wazuh-scan-images.sh
@@ -7,24 +7,30 @@ SBOM_DIR="/opt/kayobe/stackhpc/sboms"
 mkdir -p "$SBOM_DIR"
 
 # Ensure the custom output template exists
-cat <<EOL > "$SBOM_DIR/trivy-custom.tmpl"
-"Package","Version Installed","Vulnerability ID","Severity","Title"
-{{- range \$ri, \$r := . }}
-{{- range \$vi, \$v := .Vulnerabilities }}
+if [[ ! -f "$SBOM_DIR/trivy-custom.tmpl" ]]; then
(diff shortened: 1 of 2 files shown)


Alex-Welsh:
Needs a few tweaks as discussed in the monitoring meeting last week

tools/wazuh-scan-images.sh
```diff
@@ -0,0 +1,50 @@
...
+        echo "Generating SBOM for $image"
+        if ! trivy image --quiet --format spdx-json --output "$sbom_file" "$image_id"; then
+            echo "Failed to generate SBOM for $image. Skipping."
+            continue
+        fi
+    fi
+    
+    echo "Scanning SBOM: $sbom_file"
+    # Scan SBOM and prepend image info to each output line
+    trivy sbom \
```
JohnGarbutt, line 41: Would it make sense to use cosign to add in-toto attestations during our build creation process, so we can store the signed sbom in the OCI repository in the standard way?
https://trivy.dev/docs/latest/supply-chain/attestation/sbom/#sbom-attestation

It suggests in the docs you can then verify that with trivy to understand if there are CVEs:
https://trivy.dev/docs/latest/supply-chain/attestation/sbom/#scanning

Unsure if that works with Wazah scanning though.

Alex-Welsh: It's been in the infra backlog for a little while. The plan would be to integrate it at build time. There might be a way to pull that straight to Wazuh, but I've not investigated that

PR #1905, opened 2025-09-22
Fix top_level_group_map comment

m-bull:
The Kayobe monitoring group, not the contollers group, is mapped to the Kolla-Ansible monitoring group by default.

### DIFF
--- a/etc/kayobe/kolla.yml
+++ b/etc/kayobe/kolla.yml
@@ -504,22 +504,8 @@ kolla_build_args: {}
 #         kolla-ansible group.
 # NOTE(Alex-Welsh): If you want to extend the map rather than replace it, you
 # must include the Kayobe defaults in the mapping.
-# Standard Kayobe defaults:
-#  compute:
-#    groups:
-#      - "compute"
-#  control:
-#    groups:
-#      - "controllers"
-#  monitoring:
-#    groups:
-#      - "controllers"
-#  network:
-#    groups:
-#      - "controllers"
-#  storage:
-#    groups:
-#      - "controllers"
+# The standard Kayobe defaults can be found here:
+# https://github.com/stackhpc/kayobe/blob/stackhpc/<release>/ansible/inventory/group_vars/all/kolla
 #kolla_overcloud_inventory_top_level_group_map:
 
 # List of names of top level kolla-ansible groups. Any of these groups which


MoteHue:
Should we change the comment to just link to the defaults instead? 

The network group is no longer the controllers too, and there's now a compute-vgpu group we don't mention at all

m-bull:
I guess we'd have to remember to update the link each release, but a `grep -r <previous-release>` should catch it - we have to do that in a few places, so seems like a reasonable tradeoff to me.

Alex-Welsh:
> I guess we'd have to remember to update the link each release, but a `grep -r <previous-release>` should catch it - we have to do that in a few places, so seems like a reasonable tradeoff to me.

We've got `NOTE(upgrade):` which we use for stuff like that, as a thing to search for and update each release

Alex-Welsh (dismissed):
We can tweak this in a follow up PR if needed. Lets just fix it as-is for now

MoteHue:
> We can tweak this in a follow up PR if needed. Lets just fix it as-is for now

We should still update the other groups to be correct. i.e. change `network`, `storage`,  and add `compute-vgpu`.
https://github.com/openstack/kayobe/blob/33f39fa6cace1f7b408163f7d5faa901a6491be4/ansible/inventory/group_vars/all/kolla#L360-L378
```
kolla_overcloud_inventory_top_level_group_map:
  control:
    groups:
      - controllers
  network:
    groups:
      - network
  compute:
    groups:
      - compute
  compute-vgpu:
    groups:
      - compute-vgpu
  monitoring:
    groups:
      - monitoring
  storage:
    groups:
      "{{ kolla_overcloud_inventory_storage_groups }}"

Alex-Welsh:
> > We can tweak this in a follow up PR if needed. Lets just fix it as-is for now
> 
> We should still update the other groups to be correct. i.e. change `network`, `storage`, and add `compute-vgpu`. https://github.com/openstack/kayobe/blob/33f39fa6cace1f7b408163f7d5faa901a6491be4/ansible/inventory/group_vars/all/kolla#L360-L378
> 
> ```
> kolla_overcloud_inventory_top_level_group_map:
>   control:
>     groups:
>       - controllers
>   network:
>     groups:
>       - network
>   compute:
>     groups:
>       - compute
>   compute-vgpu:
>     groups:
>       - compute-vgpu
>   monitoring:
>     groups:
>       - monitoring
>   storage:
>     groups:
>       "{{ kolla_overcloud_inventory_storage_groups }}"
> ```

Ah, of course :facepalm:. My fault, trying to rush through PRs again

Alex-Welsh:
Updated to link to the right section. I've made the link generic enough that it shouldn't need to be changed between releases

MoteHue (approved):

PR #1909, opened 2025-09-24
Add missing burn-in element

darmach:


mnasiadka (changes requested):
Build new images and add their tags please

releasenotes/notes/ipa-burn-in-52df17494de29627.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Adds burn-in element.
```
mnasiadka:
```suggestion
    Adds burn-in element to Ironic Python Agent to support Ironic burn-in capability.
```

Alex-Welsh:
@darmach does this enable the burn-in test by default or is another step required?
If it's enabled by default, how long does it take and how disruptive would it be?
And if it's not enabled by default, how do you use it?

mnasiadka:
> @darmach does this enable the burn-in test by default or is another step required? If it's enabled by default, how long does it take and how disruptive would it be? And if it's not enabled by default, how do you use it?

You need to run cleaning with burn in step - https://docs.openstack.org/ironic/latest/admin/hardware-burn-in.html

mnasiadka (dismissed):

mnasiadka (approved):

PR #1913, opened 2025-09-26
Updating policy for neutron and nova to include role: baremetaluser

claudia-lola:
Updating policy to allow role: baremetaluser to map baremetal instances to specific ironic nodes (this is desirable so that the instance name = ironic node name, making debugging less confusing)

### DIFF
--- a/etc/kayobe/environments/baremetal-policy/README.rst
+++ b/etc/kayobe/environments/baremetal-policy/README.rst
@@ -1,0 +1,13 @@
+Policy for a baremetaluser role
+===============================
+
+When deploying Slurm on baremetal nodes, it is typical to select a specific
+baremetal node, and give it the expected hostname. We allow this via a tweak to
+Nova policy.
+
+Similarly, it is common that the IP address has to match the expected one for
+the given node. We tweak neutron policy to allow fixed IPs, even when we do
+not own the network.
+
+We should never use the admin role to do these operations, as it has far too
+much privilege.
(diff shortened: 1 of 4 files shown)


Alex-Welsh:
@claudia-lola could you add a release note?

bbezak:
do we want that to be default for to all our deployments? IIRC we agreed to implement a mixin type of environments to implement such features

JohnGarbutt:
> do we want that to be default for to all our deployments? IIRC we agreed to implement a mixin type of environments to implement such features

Its a good question... I am not sure.

In this specific case, it only allows extra access to the APIs to people with additional roles, and I like the idea of applying this everywhere. But I think this should be at upgrade, not mid cycle.

It might be too late to get this into Epoxy before we have people running Epoxy? Unsure. We should discuss at the next infra meeting?

mnasiadka (changes requested):
So first of all - it can't be the default.
Second of all - we don't have support for merging policy.yaml files together - so I would say before that lands - we need to work on the kolla-ansible side to add support for either policy.d (which would be ideal) - or merging policy.yaml files together and validating them?

JohnGarbutt:
> So first of all - it can't be the default.
> Second of all - we don't have support for merging policy.yaml files together - so I would say before that lands - we need to work on the kolla-ansible side to add support for either policy.d (which would be ideal) - or merging policy.yaml files together and validating them?

But the alternative is users bricking the cloud using the admin role?

I didn't know the policy didn't merge. Given people have overrides that is a very good reason for this to not be the default. Supporting policy.d makes very good sense as something for the backlog.

Is a mix-in for now a better middle ground for now?

mnasiadka (dismissed):

priteau (approved):

PR #1939, opened 2025-10-14
CI: Add better/longer retries to AIO workflow TF

Alex-Welsh:
Previously, the terraform apply would just run 5 times. If it failed, it would wait 60 seconds.
All AIOs are triggered at the same time, so might conflict.

Now, it runs 6 times. After 3 attempts, wait for 2 hours (the cloud is probably at capacity, waits for other jobs to finish)
Normal waits are randomised from 1 to 3 minutes

### DIFF
--- a/.github/workflows/stackhpc-all-in-one.yml
+++ b/.github/workflows/stackhpc-all-in-one.yml
@@ -190,7 +190,10 @@ jobs:
       - name: Terraform Apply
         id: tf_apply
         run: |
-          for attempt in $(seq 5); do
+          # Try up to 6 times to create the infrastructure, destroying and retrying if it fails.
+          # If it fails 3 times, wait 2 hours before trying again.
+          # The cloud is likely just at capacity, so wait until other jobs finish.
+          for attempt in $(seq 6); do
               if terraform apply -auto-approve; then
                   echo "Created infrastructure on attempt $attempt"
                   exit 0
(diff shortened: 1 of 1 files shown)


Alex-Welsh:
The one time I want to test a CI failure, every AIO works perfectly first time :upside_down_face:

.github/workflows/stackhpc-all-in-one.yml
```diff
@@ -190,16 +190,23 @@ jobs:
...
               sleep 10
               terraform destroy -auto-approve
-              sleep 60
-          done
+              if [ "$attempt" -eq 3 ]; then
+                  echo "Sleeping for 2 hours after 3 failed attempts..."
+                  sleep 7200
+              else
+                  sleep $(shuf -i 60-180 -n 1)
+              fi          done
```
priteau: Please move done to its own line with the right indentation.

priteau (changes requested):

.github/workflows/stackhpc-all-in-one.yml
Alex-Welsh: Done

priteau (approved):
The jump from 60-180 seconds to 7200 is huge (an exponential backoff is more conventional), but let's see how it works. We can keep tuning it.

PR #1941, opened 2025-10-16
Add ARM64 overcloud host image builds

cityofships:
An alternative to #1497. 
-  Don't give a choice to operators for picking a platform - images targeting both platforms are always built and uploaded, so they stay close to each other.
- Make distro and platform builds parallel so the workflow finishes faster.
- Does not use cross-arch at all. Leafcloud builds will not be supported, since there's no arm64 hypervisors there.

Depends on: https://github.com/stackhpc/stackhpc-release-train/pull/441

### DIFF
--- a/etc/kayobe/stackhpc.yml
+++ b/etc/kayobe/stackhpc.yml
@@ -247,7 +247,7 @@ stackhpc_enable_kolla_ansible_check: true
 download_amphora_from_ark: true
 
 # Octavia Amphora image version
-stackhpc_amphora_image_version: "2025.1-20250619T113933"
+stackhpc_amphora_image_version: "2025.1-20260319T131237"
 
 ################################################################################
 # Certificate Authority
--- a/releasenotes/notes/bump-amphora-image-noble-ea7a163beff43242.yaml
+++ b/releasenotes/notes/bump-amphora-image-noble-ea7a163beff43242.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
+    The Octavia Amphora image was rebuilt and now uses Ubuntu 24.04 LTS instead
+    of Ubuntu 22.04 LTS.


cityofships:
https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/18563982840

.github/workflows/overcloud-host-image-build.yml
```diff
@@ -288,15 +311,211 @@ jobs:
...
+          scp -r ubuntu@$(jq -r .access_ip_v4.value src/kayobe-config/etc/kayobe/environments/ci-builder/tf-outputs.yml):/opt/kayobe/images/*/*.std* ./logs/
+          scp -r ubuntu@$(jq -r .access_ip_v4.value src/kayobe-config/etc/kayobe/environments/ci-builder/tf-outputs.yml):/tmp/updated_images.txt ./logs/ || true
+        if: always()
+
+      - name: Fail if any overcloud host image builds failed
+        run: |
+          echo "Builds failed. See workflow artifacts for details." &&
+          exit 1
+        if: steps.build_rocky_9.outcome == 'failure' ||
+            steps.build_ubuntu_noble.outcome == 'failure'
```
Alex-Welsh: build_ubuntu_noble will now be separate

```suggestion
        if: steps.build_rocky_9.outcome == 'failure'
```

.github/workflows/overcloud-host-image-build.yml
```diff
@@ -396,42 +615,530 @@ jobs:
...
+          scp -r ubuntu@$(jq -r .access_ip_v4.value src/kayobe-config/etc/kayobe/environments/ci-builder/tf-outputs.yml):/opt/kayobe/images/*/*.std* ./logs/
+          scp -r ubuntu@$(jq -r .access_ip_v4.value src/kayobe-config/etc/kayobe/environments/ci-builder/tf-outputs.yml):/tmp/updated_images.txt ./logs/ || true
+        if: always()
+
+      - name: Fail if any overcloud host image builds failed
+        run: |
+          echo "Builds failed. See workflow artifacts for details." &&
+          exit 1
+        if: steps.build_rocky_9_arm64.outcome == 'failure' ||
+            steps.build_ubuntu_noble_arm64.outcome == 'failure'
```
Alex-Welsh:
```suggestion
        if: steps.build_rocky_9_arm64.outcome == 'failure' 
```

.github/workflows/overcloud-host-image-build.yml
```diff
@@ -396,42 +615,530 @@ jobs:
...
```
Alex-Welsh:
```suggestion
        if: steps.build_ubuntu_noble_arm64.outcome == 'failure'
```

.github/workflows/overcloud-host-image-build.yml
```diff
@@ -396,42 +615,530 @@ jobs:
...
+
+      - name: Generate terraform.tfvars
+        run: |
+          cat << EOF > terraform.tfvars
+          ssh_public_key = "id_rsa.pub"
+          ssh_username = "ubuntu"
+          aio_vm_name = "skc-host-image-builder-rocky-arm64"
+          # Must be an Ubuntu Noble host to successfully build all images
+          # This MUST NOT be an LVM image. It can cause confusing conficts with the built image.
+          aio_vm_image = "${{ vars.HOST_IMAGE_BUILD_IMAGE_ARM64 }}"
```
Alex-Welsh: I just want to check my understanding

This is how we know the runner spawns on an ARM64 host, and because the runner arch is ARM, the images will be as well?

Alex-Welsh (changes requested):
It would be nice to include some validation to make sure we're actually building images for the right arch.

Unfortunately it doesn't come out of qemu-img info, the cleanest I have of finding it is:

```
sudo apt install -y libguestfs-tools
sudo guestfish --ro -a path/to/image -i file /bin/bash | grep x86-64
```
or
```
sudo apt install -y libguestfs-tools
sudo guestfish --ro -a path/to/image -i file /bin/bash | grep aarch64
```

.github/workflows/overcloud-host-image-build.yml
cityofships: Not the runner, but the builder VM, yes. It's default for DIB to build for the platform it's running on.

 It is also 1) the quickest way of building 2) cross-arch building with `os_images` is currently unstable - `architecture` does not always match the underlying OS (`aarch64` vs. `arm64`) so it's trying to emulate ARM on ARM when running DIB.

.github/workflows/overcloud-host-image-build.yml
cityofships: Good catch

Alex-Welsh:
@cityofships looks like you have a conflict. Needs a rebase, and probably worth squashing while you're at it

priteau:
Can you prevent upload of ARM64 images to Leafcloud?

PR #1969, opened 2025-11-05
Add secret store unseal playbook for action runners

seunghun1ee:
When CI action runners are rebooted, their secret store needs to be unsealed.
Added unsealing playbook for them.

### DIFF
--- a/releasenotes/notes/add-secret-store-unseal-playbook-for-ci-runners-ec758e93a8984760.yaml
+++ b/releasenotes/notes/add-secret-store-unseal-playbook-for-ci-runners-ec758e93a8984760.yaml
@@ -1,0 +1,7 @@
+---
+features:
+  - |
+    Added a playbook for unsealing secret store deployed at CI runners (Hosts
+    in github-runners or gitlab-runners inventory group).
+    To run CIs after rebooting CI runners, secret store needs to be unsealed
+    using this playbook.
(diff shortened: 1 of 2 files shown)


Alex-Welsh (commented):
Needs a reno

Given how similar this and the other secret-store-unseal-*.yml playbooks are, we should probably just extend the vault_unseal role to reduce duplication a bit.

That's scope creep though, and I'm happy to do it this way for now.

releasenotes/notes/add-secret-store-unseal-playbook-for-ci-runners-ec758e93a8984760.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Added a playbook for unsealing secret store deployed at CI runners (Hosts
+    in github-runners or gitlab-runners inventory group).
+    To run CIs after rebooting CI runners, secret store needs to be unsealed
```
Alex-Welsh: nit: singular
```suggestion
    To run CI after rebooting CI runners, secret store needs to be unsealed
```

Alex-Welsh (commented):
One more general thought, should we just run this every time we trigger CI?

seunghun1ee:
I don't know... Maybe?

seunghun1ee:
Well even if we want to run this every time we run CI, I think that needs to be done via separate PR.

Alex-Welsh (approved):

PR #1973, opened 2025-11-10
Document Ceph Squid bug and workaround

priteau:


doc/source/operations/upgrading-ceph.rst
```diff
@@ -159,6 +159,16 @@ versions``. Once confirmed, run the following command:
 
    ceph osd require-osd-release squid
 
+.. warning::
+
+   There is a `known bug causing OSDs created in Squid to crash
+   <https://docs.clyso.com/blog/#squid-deployed-osds-are-crashing>`__. Disable
```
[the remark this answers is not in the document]
priteau: I updated slightly.

doc/source/operations/upgrading-ceph.rst
```diff
@@ -159,6 +159,18 @@ versions``. Once confirmed, run the following command:
 
    ceph osd require-osd-release squid
 
+.. warning::
```
Alex-Welsh, line 162: Should/Can we do this before the upgrade? if so, could we move the warning to the top of the section?

priteau: We cannot do it before the upgrade, because this option was only introduced in Squid. On Reef:

```
# ceph config set osd bluestore_elastic_shared_blobs 0
Error EINVAL: unrecognized config option 'bluestore_elastic_shared_blobs'
```

It should be safe to apply only after the upgrade is complete, as long as it is done before any new OSD is created.

Alex-Welsh (approved):

PR #1981, opened 2025-11-13
Docs theme 2023.1

Alex-Welsh:


Alex-Welsh:
Not worth the time to fix issues right now

PR #1982, opened 2025-11-13
GT-1561: Add basic Dell SONiC driver

dougszumski:
Basic support for Dell SONiC switches for Arcus.

Build: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/19330333096/job/55291529233
Patch: https://github.com/stackhpc/networking-generic-switch/pull/130


Alex-Welsh (approved):
LGTM but is there an equivalent patch for 2025.1? I'm a bit concerned we'll forget about it, upgrade, and break things

dougszumski:
> LGTM but is there an equivalent patch for 2025.1? I'm a bit concerned we'll forget about it, upgrade, and break things

Thanks Alex - I am trying to attack it from the other side for later releases since there is a Dell Sonic driver upstream. I'll leave the ticket open until I've resolved it one way or another.

PR #1989, opened 2025-11-19
(automated) Bump kolla-ansible to stackhpc/18.8.0.4

Alex-Welsh:
Replacement for #1985 which is having weird CI issues


Alex-Welsh:
recheck

PR #1992, opened 2025-11-21
Add Github CI runner proxy support

darmach:
Conditionally sets custom_env fact to leverage
monolithprojects.github_actions_runner role support for seting proxy for deployed Github runners.

Confirmed working on environment with mandatory proxy ONLY - it would be great if someone could give this a go in non-proxy setup. With custom_env defined monolithprojects role creates .env file in /opt/actions*... directory (per runner) which is ingested by the running runner (ran on the runner VM ,)


darmach:
Putting into draft, need to rework proxy support in docker role

PR #1997, opened 2025-11-24
docs: add warning for `Noble` upgrade on `Ceph` hosts

jackhodgkiss:
See bug https://tracker.ceph.com/issues/66389

### DIFF
--- a/doc/source/operations/ubuntu-noble.rst
+++ b/doc/source/operations/ubuntu-noble.rst
@@ -47,8 +47,11 @@ The following types of hosts will be covered in the future:
 
 .. warning::
 
-   Ceph node upgrades have not yet been performed outside of a virtualised test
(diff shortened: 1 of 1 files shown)


doc/source/operations/ubuntu-noble.rst
```diff
@@ -353,6 +355,7 @@ Storage
 Potential issues
 ----------------
 
+-  Ensure the Ceph cluster is running at least Reef v18.2.5. Upgrading hosts with an older Ceph version will cause daemons to fail.
```
Alex-Welsh: Could you wrap this line?

Otherwise content lgtm

Alex-Welsh (approved):
If I was being picky, I'd say we should include a link to the Ceph upgrade docs, but I don't think it should hold this up

PR #2001, opened 2025-11-25
add wazuh purge playbook

g0rgamesh:


doc/source/configuration/wazuh.rst
```diff
@@ -424,6 +424,12 @@ Verification
 The Wazuh agents should register with the Wazuh manager. This can be verified via the agents page in Wazuh Portal.
 Check CIS benchmark output in agent section.
 
+Removal
+-------
+The following playbook can be used to purge all Wazuh components from a host. This is particularly useful for Wazuh servers that are not hosted on an infra-vm.
```
Alex-Welsh: > The following playbook can be used to purge all Wazuh components from a host.

Specifically, the playbook runs on the wazuh-manager host at the moment.
Do you want to make it generic? i.e. add another play to purge the Wazuh agents?

I'm fine with either that or just being more specific that it only wipes the manager

Alex-Welsh (changes requested):

doc/source/configuration/wazuh.rst
g0rgamesh: It's specifically for wazuh-manager host. Removing agents is done via API so that would just be packaging curls into ansible, not sure if it's worth the hassle. If I will add it I will push it in a separate PR.

Alex-Welsh: To be clear, I mean uninstalling the wazuh-agent process from all the other hosts, like the overcloud/seed/etc. Is that not basically the same steps as shutting down the wazuh-manager processes?

([docs](https://documentation.wazuh.com/current/installation-guide/uninstalling-wazuh/agent.html))

g0rgamesh: When I tried it some time ago it just rendered the host marked as disconnected in Wazuh dashboard and it was visible there until it got removed via API.

Alex-Welsh: Well, the same playbook deletes the dashboard.

It's scope creep anyway, so I'm happy to focus on the manager deletion. Lets just make that clear in the doc and maybe change the name of the playbook to ``wazuh-manager-purge.yml``

g0rgamesh: Done, enrolling agents on a fresh manager just requires running ``$KAYOBE_CONFIG_PATH/ansible/deployment/wazuh-agent.yml`` playbook, even if the agent packages were not removed previously. I've tested that.

Alex-Welsh (approved):

PR #2006, opened 2025-11-27
disable zuul checks

g0rgamesh:
- diable zuul checks until they are fixed

### DIFF
--- a/zuul.d/project.yaml
+++ b/zuul.d/project.yaml
@@ -4,11 +4,11 @@
     check:
       jobs:
         - openstack-tox-pep8
-
-    check-review:
-      jobs:
-        - tenks-ubuntu-noble
-        - tenks-rocky-9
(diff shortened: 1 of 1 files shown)


Alex-Welsh:
I'm not sure we should disable them completely for now.
The check-review pipeline still runs, it just fails a lot.
I'd rather leave it running for now, and work on fixing it.

The check pipeline on the other hand, I don't see much benefit from. It just fails immediately, so we could remove that for now

g0rgamesh:
> I'm not sure we should disable them completely for now. The check-review pipeline still runs, it just fails a lot. I'd rather leave it running for now, and work on fixing it.
> 
> The check pipeline on the other hand, I don't see much benefit from. It just fails immediately, so we could remove that for now

Ok, I'll close it then.

PR #2019, opened 2025-12-01
Fixes in GPU docs

MoteHue:


Alex-Welsh (changes requested):
Also one here:
```
operations/bifrost-hardware-inventory-management.rst - To build ipa image with extra-hardware  you need to edit ``ipa.yml`` and add this:
```
Could you fix that as well?

I don't think there are any others

Alex-Welsh (approved):

PR #2026, opened 2025-12-05
Fix blackbox endpoints when blackbox-exporter placement is customised

jovial:
This variable needs to be defined on the prometheus group since that is where prometheus.yml is generated.


Alex-Welsh:
@jovial could you re-target this at 2025.1 and add a release note?

jovial:
> @jovial could you re-target this at 2025.1 and add a release note?

2025.1 is OK because the code was moved the automatic blackbox generation was moved into kolla-ansible :partying_face: . I can add a reno.

Alex-Welsh:
> > @jovial could you re-target this at 2025.1 and add a release note?
> 
> 2025.1 is OK because the code was moved the automatic blackbox generation was moved into kolla-ansible 🥳 . I can add a reno.

Unfortunately, not everything was moved to KA. We still have the backend endpoints defined in SKC in 2025.1. The [file still exists](https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2025.1/etc/kayobe/kolla/inventory/group_vars/prometheus-blackbox-exporter)

There was a plan to move backend endpoints as well to KA, but we decided to try and use [file-based service discovery](https://prometheus.io/docs/guides/file-sd/) in prometheus for it. That didn't work, so we implemented it downstream as a quick fix and no one's had time to upstream it yet.

Alex-Welsh (approved):

PR #2027, opened 2025-12-08
Bump stackhpc-openstack-gh-workflows to 1.5.0

seunghun1ee:
To make upgrading tests to work with new versioned terraform-kayobe-multinode branches.

### DIFF
--- a/.github/workflows/multinode-inputs.py
+++ b/.github/workflows/multinode-inputs.py
@@ -17,6 +17,7 @@ class OSRelease:
 @dataclass
 class OpenStackRelease:
     version: str
+    previous_version: str
     os_releases: t.List[OSRelease]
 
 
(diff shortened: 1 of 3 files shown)


Alex-Welsh (changes requested):
https://github.com/stackhpc/stackhpc-openstack-gh-workflows/pull/16/files#diff-ac002649dfe09359bfd52742a41205f21d61442615bad584c5b9e85e3bd158aeR101

Will fail on major upgrade if previous version is undefined. You'll need to update multinode-inputs.py

seunghun1ee:
Oops

Alex-Welsh:
Simplified the logic a bit because it was getting less readable. I've essentially reverted 3ba886b152ee780ea843f427f81a2b92ac317a73. IDK why I thought that was a good idea in the first place

Alex-Welsh (approved):

PR #2034, opened 2025-12-11
imag-build workflows: remove unused step

elelaysh:
the step was broken because git clone is in src/kayobe-config.

### DIFF
--- a/.github/workflows/ipa-image-build.yml
+++ b/.github/workflows/ipa-image-build.yml
@@ -61,11 +61,6 @@ jobs:
         with:
           path: src/kayobe-config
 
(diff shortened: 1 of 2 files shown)


Alex-Welsh (commented):
If this step is not used anywhere, should we just remove it?

elelaysh:
See fixed [output in workflow](https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/20129671668/job/57767770546#step:6:10)

elelaysh:
> If this step is not used anywhere, should we just remove it?

@Alex-Welsh done, thanks

Alex-Welsh (approved):

PR #2065, opened 2026-01-09
Rotate AIO vault password

Alex-Welsh:


Alex-Welsh:
Upgrade jobs will fail because it checks out the previous branch, and unless we do it all in reverse order, the previous branch won't have the right changes

priteau (approved):

PR #2061, opened 2026-01-07
Rocky Linux 9.7 containers

elelaysh:
This PR builds upon #2025.

- it provides an updated overcloud host image and containers based on RockyLinux 9.7;
- cadvisor is updated to 0.54.1 to fix vulnerabitilities;
- opensearch-dashboards is updated to 2.19.4 to fix vulnerabilities.

PR #2109, opened 2026-01-28
CI: Run container image pull in its own step

priteau:
This might slow down a full AIO job due to the additional iteration through kayobe and kolla-ansible roles, but it will allow us to better understand the impact of container image pull on CI performance.


priteau:
Closing in favour of https://github.com/stackhpc/stackhpc-kayobe-config/pull/2061

PR #2129, opened 2026-02-06
Use new Ark user credentials for multinode env

Alex-Welsh:
It's time we stop using the AIO user for everything. I've made a new skc-ci-multinode user in Ark and rotated out the old vault PW. I'll update the GitHub repo secrets once this merges and backport the change to other branches


priteau (approved):
I verified that I could decrypt both vaults 👍

PR #2135, opened 2026-02-06
Rotate credentials for ci-builder env

Alex-Welsh:
Created a new user (read only) for pulling package repos from Ark, instead of using the AIO user again
Also changed the vault PW be unique (added to password manager)
Same rules apply as normal, will need to update the secret in GitHub and backport once merged


Alex-Welsh:
Converting to draft so no one merges this until next week. I don't want to deal with any fallout over the weekend and I'll be off on Monday

priteau (approved):

PR #2143, opened 2026-02-09
Fix allowed vulnerability list not reset between images

elelaysh:
`.trivyignore` was not removed between image scanned

### DIFF
--- a/tools/scan-images.sh
+++ b/tools/scan-images.sh
@@ -46,12 +46,12 @@ file_prep() {
 # Gather image lists
 get_images() {
   local output_file="$1-scanned-container-images.txt"
-  
+
   docker image ls \
     --filter "reference=ark.stackhpc.com/stackhpc-dev/*:$2*" \
     --format "{{.Repository}}:{{.Tag}}" \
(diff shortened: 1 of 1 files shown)


tools/scan-images.sh
```diff
@@ -63,7 +63,7 @@ generate_trivy_ignore() {
   local image_vulnerabilities
   image_vulnerabilities=$(yq ."$imagename"'_allowed_vulnerabilities[]' src/kayobe-config/etc/kayobe/trivy/allowed-vulnerabilities.yml 2> /dev/null)
 
-  touch .trivyignore
+  cat > .trivyignore  # ensure we start from a clean slate
```
priteau: I think this will hang waiting on input? Why not use instead: `truncate -s 0 .trivyignore`

It will also create the file if missing.

priteau (dismissed):

tools/scan-images.sh
elelaysh: yes, you're right. `truncate` is great.

Alex-Welsh (approved):

PR #2162, opened 2026-02-16
Fix Amphora register playbook failing when using TLS

seunghun1ee:
CA certificate was not injected correctly for this playbook.
Plus, using openstack.cloud.image Ansible module for renaming the existing old Amphora image.


seunghun1ee:
It would be nice to backport this to Caracal too.

seunghun1ee:
Need more tests. Switch to draft.

PR #2169, opened 2026-02-23
Fix StackHPC cloud tests not run except on upgrades

elelaysh:
the conjunction produced the string 'false', which broke the playbook run


elelaysh:
See https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/22302783834/job/64515215280?pr=2126#step:41:9

> Kayobe playbook false is invalid: Path does not exist

elelaysh:
Issue is not present in stackhpc/2025.1 the line disappeared in the playbook during a refactor?

Alex-Welsh:
> Issue is not present in stackhpc/2025.1 the line disappeared in the playbook during a refactor?

Looks like the issue was introduced by a fix for a different issue which only applies to 2024.1

Alex-Welsh (approved):

priteau (approved):

PR #2170, opened 2026-02-23
Fix Docker version check

priteau:
We decided to keep 2024.1 at Docker Engine v28.


elelaysh (approved):
this will catch unwanted upgrade indeed.
Thanks

PR #2175, opened 2026-02-24
Use Docker 29 on 2024.1 Caracal

priteau:
Docker can be updated to version 29 using a package update. We risk downgrading to 28 when running `kayobe overcloud host configure`, which might be worse than the upgrade from 28 to 29.

Revert "limit docker-ce to 28 on Caracal"

This reverts commit 18027cac09dc03eae5cb623ad722832f04266b62.

Revert "Fix Docker version check"

This reverts commit 01de41aad7e239b6409de67ed1374448b38756b0.

### DIFF
--- a/etc/kayobe/ansible/stackhpc-cloud-tests.yml
+++ b/etc/kayobe/ansible/stackhpc-cloud-tests.yml
@@ -142,7 +142,7 @@
             # Inclusive min
             sct_docker_version_min: "24.0.0"
             # Exclusive max
-            sct_docker_version_max: "29.0.0"
+            sct_docker_version_max: "30.0.0"
             sct_selinux_state: "{{ selinux_state }}"
           failed_when: host_results.rc not in [0, 1]
           register: host_results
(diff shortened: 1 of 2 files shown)


elelaysh (approved):
OK, it's simpler than using versionlock to keep docker 28, that we would need to lift at a later upgrade

Alex-Welsh:
Why did we limit it in the first place? I was probably told before but have forgotten

elelaysh:
> Why did we limit it in the first place? I was probably told before but have forgotten

We have been seeing issues where OVS containers failed to fully delete: the container is gone from the docker ps -a output but still running in the background. This leads to PID conflicts with the new containers launched by kolla-ansible.

Alex-Welsh:
> > Why did we limit it in the first place? I was probably told before but have forgotten
> 
> We have been seeing issues where OVS containers failed to fully delete: the container is gone from the docker ps -a output but still running in the background. This leads to PID conflicts with the new containers launched by kolla-ansible.

And we're confident the error is gone?

priteau:
> > > Why did we limit it in the first place? I was probably told before but have forgotten
> > 
> > 
> > We have been seeing issues where OVS containers failed to fully delete: the container is gone from the docker ps -a output but still running in the background. This leads to PID conflicts with the new containers launched by kolla-ansible.
> 
> And we're confident the error is gone?

Not 100% sure but I haven't seen issues when updating to Docker 29 in another deployment.

Alex-Welsh (approved):

PR #2177, opened 2026-02-26
Add Rocky 10 repositories

owenjones:
- some of the RL9 repositories might also be required in RL10 (is there a list of what we want in each image somewhere?), so will possibly need to come back and fix the required flag on those to include them
- it feels like there's scope for templating this a bit better as there's a lot of duplication between RL9 and RL10 entries, but that will need a bit of planning


etc/kayobe/pulp-repo-versions.yml
Alex-Welsh, line 1: Lets keep this change to RL10 only. Could you remove all the other changes from this file?

etc/kayobe/pulp-repo-versions.yml
```diff
@@ -1,9 +1,18 @@
 ---
 # This file is autogenerated by Ansible using the following workflow:
 # https://github.com/stackhpc/stackhpc-release-train/actions/workflows/package-update-kayobe.yml
+stackhpc_pulp_repo_almalinux_10_proxysql_3_0_version: 20260220T202834
```
Alex-Welsh: ProxySQL 3.0 is only required for the Gazpacho (2026.1) release onwards.
Epoxy (2025.1) re-uses the same repo for both RL9 and RL10
See: https://opendev.org/openstack/kolla/src/branch/stable/2025.1/docker/base/proxysql.repo

So we don't need any of the ProxySQL 3.0 changes in this PR

etc/kayobe/pulp-repo-versions.yml
```diff
@@ -18,26 +27,58 @@ stackhpc_pulp_repo_doca_2_9_1_rhel9_5_modules_version: 20250115T150314
...
-stackhpc_pulp_repo_epel_9_aarch64_version: 20260204T223146
-stackhpc_pulp_repo_epel_9_version: 20260204T220346
-stackhpc_pulp_repo_grafana_version: 20260204T212232
+stackhpc_pulp_repo_epel_10_aarch64_version: 20260222T221647
+stackhpc_pulp_repo_epel_10_version: 20260221T212353
+stackhpc_pulp_repo_epel_9_aarch64_version: 20260222T214442
+stackhpc_pulp_repo_epel_9_version: 20260222T210438
+stackhpc_pulp_repo_grafana_version: 20260220T202109
 stackhpc_pulp_repo_opensearch_2_x_version: 20251106T202313
+stackhpc_pulp_repo_opensearch_3_x_version: 20260220T152827
```
Alex-Welsh: Similar story to proxysql
opensearch and opensearch dashboards are only updated to 3.x in Gazpacho. We'll re-use 2.x until then

etc/kayobe/pulp-repo-versions.yml
```diff
@@ -18,26 +27,58 @@ stackhpc_pulp_repo_doca_2_9_1_rhel9_5_modules_version: 20250115T150314
...
 stackhpc_pulp_repo_opensearch_dashboards_2_x_version: 20251106T202313
+stackhpc_pulp_repo_opensearch_dashboards_3_x_version: 20260220T152827
+stackhpc_pulp_repo_rhel10_rabbitmq_erlang_27_aarch64_version: 20260220T152827
 stackhpc_pulp_repo_rhel9_rabbitmq_erlang_26_aarch64_version: 20260112T224827
 stackhpc_pulp_repo_rhel9_rabbitmq_erlang_27_aarch64_version: 20260112T224827
 stackhpc_pulp_repo_rhel9_rabbitmq_erlang_version: 20251130T213625
-stackhpc_pulp_repo_rhel9_rabbitmq_server_version: 20260122T202504
+stackhpc_pulp_repo_rhel9_rabbitmq_server_version: 20260220T202834
+stackhpc_pulp_repo_rhel_10_fluent_6_aarch64_version: 20260220T152827
```
Alex-Welsh: Fluent 6 not required,
Epoxy uses fluent 5 (confusingly named treasuredata 5): https://opendev.org/openstack/kolla/src/branch/stable/2025.1/docker/base/td.repo

Same as proxysql, we'll use the rhel9 repo

etc/kayobe/pulp.yml
```diff
@@ -198,16 +198,16 @@ stackhpc_pulp_distribution_deb_production: >-
 
 # Content comes in from release repositories, via pinned repo versions.
 
+# Define the architecture suffix
+arch_suffix: "{{ '_aarch64' if kolla_base_arch == 'aarch64' else '' }}"
```
Alex-Welsh: Can this be reverted? I'm not sure why it's moved

etc/kayobe/pulp-repo-versions.yml
```diff
@@ -1,9 +1,18 @@
...
 stackhpc_pulp_repo_almalinux_9_proxysql_2_7_version: 20250627T134211
-stackhpc_pulp_repo_centos_stream_9_docker_aarch64_version: 20260202T213000
-stackhpc_pulp_repo_centos_stream_9_docker_version: 20260202T201529
+stackhpc_pulp_repo_centos_stream_10_docker_aarch64_version: 20260220T152827
+stackhpc_pulp_repo_centos_stream_10_docker_version: 20260220T152827
+stackhpc_pulp_repo_centos_stream_10_nfv_openvswitch_aarch64_version: 20260220T152827
+stackhpc_pulp_repo_centos_stream_10_nfv_openvswitch_version: 20260220T152827
+stackhpc_pulp_repo_centos_stream_10_storage_ceph_squid_aarch64_version: 20260220T152827
+stackhpc_pulp_repo_centos_stream_10_storage_ceph_squid_version: 20260220T152827
+stackhpc_pulp_repo_centos_stream_10_storage_ceph_tentacle_aarch64_version: 20260220T152827
```
Alex-Welsh: I don't think tentacle is required for this release. We definitely need Squid, and I'm now wondering if we need Reef...

I'll check that and get back to you

Alex-Welsh (changes requested):
This is a good start. 
When I added the repos to release train, I included repos for both Epoxy and Gazpacho, which seems to have caused some confusion. I apologise, I should've split those changes out into two PRs.

Alex-Welsh:
> * some of the RL9 repositories might also be required in RL10 (is there a list of what we want in each image somewhere?), so will possibly need to come back and fix the required flag on those to include them

I think you need a combination of [this file](https://opendev.org/openstack/kolla/src/branch/stable/2025.1/kolla/template/repos-el10.yaml) and the `*.repo` files in [this directory](https://opendev.org/openstack/kolla/src/branch/stable/2025.1/docker/base). Looking at it now, I think I missed kolla_el10 in release train. I'll recheck that soon.

> * it feels like there's scope for templating this a bit better as there's a lot of duplication between RL9 and RL10 entries, but that will need a bit of planning

There certainly is. My concern would be there's enough that _doesn't_ overlap that de-duplication of some parts and not others would just add complexity overall. In this case, let's keep going as we have before, but we should think about this for the next release. It would require a bit of planning as you say, and we're a bit late for that this time.

etc/kayobe/pulp-repo-versions.yml
seunghun1ee: Was this decided to be included behind the scene?

owenjones: noop, I just missed taking that out in here

seunghun1ee (dismissed):

etc/kayobe/pulp-repo-versions.yml
```diff
@@ -18,11 +26,17 @@ stackhpc_pulp_repo_doca_2_9_1_rhel9_5_modules_version: 20250115T150314
 stackhpc_pulp_repo_doca_2_9_1_rhel9_5_version: 20241211T171301
 stackhpc_pulp_repo_doca_2_9_3_rhel9_6_modules_version: 20250714T141841
 stackhpc_pulp_repo_doca_2_9_3_rhel9_6_version: 20250703T135021
+stackhpc_pulp_repo_doca_3_2_1_rhel10_aarch64_version: 20260220T152827
+stackhpc_pulp_repo_doca_3_2_1_rhel10_x86_64_version: 20260220T152827
```
Alex-Welsh, line 30: Does this work? given that it doesn't match the other DOCA var name formats it looks... suspicious

owenjones: Ah I used it from the [short name in release train](https://github.com/stackhpc/stackhpc-release-train/blob/b8f8513e6a929e13c66c8ac75a4599fc87fd4f75/ansible/inventory/group_vars/all/rpm-package-repos#L986) - will it need fixing there to be able to update correct package version here?

Alex-Welsh: Let's stick with this naming format. Unfortunately it's going to make the templating in ofed.yml even more complex, but soonish we'll be switching to Gazpacho, where we can finally stick to a single naming convention (this one)

Alex-Welsh (approved):

PR #2183, opened 2026-02-27
fix: sort `redfish_exporter_targets` to avoid config changes

jackhodgkiss:
When templating `redfish_exporter_targets` in `60-redfish.yml` the groups are written randomly and as a result can cause unnecessary updates to the templated file which can lead to reboots of prometheus as well as polluting configuration diff.


dougszumski (approved):
Great, thanks Jack!

PR #2192, opened 2026-03-02
Fix Ironic bug LP#2129260

seunghun1ee:
Fixed Ironic bug LP#2129260 "Large glance images causing conductor CPU starvation"


seunghun1ee:
Tested on a customer's test cloud

PR #2193, opened 2026-03-03
Fix manifest push failure handling

bbezak:


seunghun1ee (dismissed):

tools/multiarch-manifests.sh
mnasiadka: Can we rework this to meaningful values and move the set -o pipefail here?

mnasiadka (changes requested):

mnasiadka (approved):

seunghun1ee:
We need to backport this to Caracal right?

PR #2202, opened 2026-03-09
Fully qualify seed container images

m-bull:
Seed hosts configured to use the Podman container engine do not automatically resolve image short-names to docker.io like seed hosts using the Docker container engine.

Fully qualify seed container images so that Podman does not fail to resolve them out of the box.

Also update pre and post playbooks, which run before and after deploying seed containers, to call kayobe_container_* rather than docker_* modules, so that they respect the `container_engine` configuration variable.


priteau (dismissed):

priteau (commented):
Maybe use `container_engine` instead of `kolla_container_engine`, since this is purely Kayobe code?

releasenotes/notes/seed-container-image-locations-3b83bb42f6c50bab.yaml
```diff
@@ -0,0 +1,10 @@
+---
+fixes:
+  - |
+    Ensure that seed hosts using Podman pull seed container
+    images using their fully-qualified name, rather than
+    their short-name.
+  - |
+    Ensure that the container_engine variable is respected
+    when running pre.yaml and post.yaml for seed
+    containers.
```
Alex-Welsh, line 10: nit: formatting
```suggestion
  - |
    Ensure that the ``container_engine`` variable is respected
    when running ``pre.yaml`` and ``post.yaml`` for seed
    containers.
```

Alex-Welsh (approved):

PR #2205, opened 2026-03-10
Fix task for when there are no NVMe drives

technowhizz:


Alex-Welsh (changes requested):
Needs a release note

Alex-Welsh (approved):
The release note should technically be wrapped at 80 chars, but there's a bunch that don't follow the rule so I'm happy to merge

PR #2223, opened 2026-03-24
CI: Split tag generation into a separate job

cityofships:
This is to prepare for introduction of multiple parallel jobs that would be resulting in set of images sharing the same tag. A new create-tag job is therefore created and can be referenced as a dependency in the 'needs' lists.


cityofships:
Run here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/23492396505

seunghun1ee (approved):

bbezak (approved):

PR #2227, opened 2026-03-25
Update openbao deploy playbook for GitHub

maxstack:
This fixes openbao deployment issues on a GitHub runner.
The playbook looked to be Gitlab tested.

Initially this playbook was: 
etc/kayobe/ansible/deployment/deploy-openbao-kayobe-automation.yml
Changed name to tie in with other deploy/unseal playbooks:
etc/kayobe/ansible/secret-store/secret-store-deploy-openbao-runners.yml

Renamed counterpart unseal as "action" is GitHub specific:
etc/kayobe/ansible/secret-store/secret-store-unseal-action-runners.yml
renamed to:
etc/kayobe/ansible/secret-store/secret-store-unseal-openbao-runners.yml

Chages made:
 * generally define all variables similarly to the seed and overcloud deploy playbooks
 * use `docker0` network interface for GitHub as the kayobe automation containers can't contact the host's localhost
 * keep the network interface as `lo` for Gitlab deployments, which seems to be where this playbook was working
 * fix api_url adding missing protocol

### DIFF
--- a/doc/source/configuration/ci-cd.rst
+++ b/doc/source/configuration/ci-cd.rst
@@ -140,41 +140,116 @@ Runner Deployment
 10. Repeat the above steps for each environment you intend to deploy runners within.
     You can share the fine-grained access token between environments.
 
+OpenBao Deployment
+------------------
+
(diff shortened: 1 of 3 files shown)


Alex-Welsh:
@maxstack could you find someone specific to review this? It looks useful and I don't want it to get stuck forever

PR #2231, opened 2026-03-26
CI: Fix container image build workflow (Zed)

seunghun1ee:
Ubuntu Noble runners now need the `--break-system-packages` option to install system-wide Python packages. However this option is absent from pip on our Ubuntu Jammy arm64 runners.

Detect the distribution version and run the appropriate command.

(cherry picked from commit 82075675fe59fbeafd959f2dd80cc2b892e8aa02)

### DIFF
--- a/.github/workflows/stackhpc-container-image-build.yml
+++ b/.github/workflows/stackhpc-container-image-build.yml
@@ -148,7 +148,12 @@ jobs:
       # Normally installed during host configure.
       - name: Install Docker Python SDK
         run: |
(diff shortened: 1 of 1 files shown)


Alex-Welsh (approved):

Alex-Welsh:
@seunghun1ee could you look into the tox issues?

seunghun1ee:
This is not needed. It was found that the whole process doesn't work with Ubuntu Noble.

PR #2243, opened 2026-03-31
Pin GitHub Actions to commits instead of tags

Alex-Welsh:
Updates all actions that were previously pinned to a tag, to be pinned to the latest commit hash

This is to mitigate the risk of supply chain attacks that target upstream GitHub actions, and avoid [node 20 deprecation](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) warnings at the same time.

### DIFF
--- a/.github/workflows/amphora-image-build.yml
+++ b/.github/workflows/amphora-image-build.yml
@@ -80,7 +80,7 @@ jobs:
           pip install -r ../src/kayobe-config/requirements.txt
 
       - name: Install terraform
-        uses: hashicorp/setup-terraform@v3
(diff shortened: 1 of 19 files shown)


seunghun1ee (commented):
Can you comment which version or tag each SHA is pointing at? (If that's available)

Alex-Welsh:
> Can you comment which version or tag each SHA is pointing at? (If that's available)

Done, also got Gemini to script the process so it's faster and less susceptible to dumb Alex errors. New script pulls the head commit from the latest release, so it's a bit smarter as well. Gist is [here](https://gist.github.com/Alex-Welsh/f67bc2487ca99d77f0ea07922924cc30)

I've also tweaked the the slack action invocation because it was jumping a few versions, so needed a fix

priteau (approved):

PR #2258, opened 2026-04-08
Improve Amphora image registration logic

priteau:
Group the image rename and registration tasks in a block to ensure both are skipped if the correct image is already present.

Use a more robust check using selectattr and checksum to determine if the image needs to be updated: with the old check there was no guarantee that the latest image would be the first element in the list.


Alex-Welsh (changes requested):
Looks alright but needs linter fixes

PR #2261, opened 2026-04-08
CI: Fix path to diagnostics playbook for upgrades

priteau:


.github/workflows/stackhpc-all-in-one.yml
MoteHue, line 473: Do we not need the same here?

priteau:
This doesn't work correctly, needs to be reworked.

.github/workflows/stackhpc-all-in-one.yml
priteau: No, we only run the stackhpc-cloud-tests on the upgraded system.

Alex-Welsh (approved):

PR #2271, opened 2026-04-15
Ensure AIOs use correct flavor

Alex-Welsh:
This fixes ~two~ many issues
1. It ensures AIOs select their flavor/network combo using the right github env var
2. It changes the AIO flavor on SMS to a custom flavor that should stop it from landing on the slow hypervisors
3. It changes the AIO instance cleanup action to let jobs run for up to 6 hours
4. It increases the size of the AIO disk so upgrade jobs don't run out of space (mainly in the docker LV)


MoteHue (approved):

seunghun1ee:
The commit 9823037d5356d0e1098a2c15d484be031fa4d718 needs to be backported to stackhpc/2024.1 to be actually have an effect.
PR created at https://github.com/stackhpc/stackhpc-kayobe-config/pull/2309

PR #2273, opened 2026-04-15
fix: restore ability to dump all configuration

jackhodgkiss:
When running `kayobe configuration dump` would fail with

```
...FAILED! => \n    msg: to_nice_yaml - 'hostvars'. 'hostvars'
```

This is due to the existence of variables that use `net_ip` filter. Therefore, this can be avoided by using `lookup` filter.

Some of these changes are done within config and would be picked automatically. However some changes are done within the documentation and would require a user to update their configuration to benefit from the changes suggested.

Note: This only works with you use `--limit` as the firewall configuration is within `all` group and gets evaluated by hosts that do not have network_interfaces such as switches. Due to the command producing a lot of output running the command without a `limit` it not advisable.

### DIFF
--- a/doc/source/configuration/walled-garden.rst
+++ b/doc/source/configuration/walled-garden.rst
@@ -29,7 +29,7 @@ In ``etc/kayobe/inventory/group_vars/overcloud/time``:
    # Following deployment we include the OpenStack VIP
 
    chrony_ntp_servers:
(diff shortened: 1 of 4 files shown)


Alex-Welsh:
Reno would be nice, otherwise looks good.

jovial (approved):

dougszumski (approved):

PR #2283, opened 2026-04-23
Build Rocky 10 IPA Images

owenjones:
Rocky 10 IPA images (amd64 and aarch64) can now be built. This change also sets overrides to allow Rocky 9 IPA images to be built (rather than CentOS Stream 9 images which is what upstream Kayobe built previously).

### DIFF
--- a/.github/workflows/ipa-image-build.yml
+++ b/.github/workflows/ipa-image-build.yml
@@ -11,6 +11,14 @@ on:
         description: Build Rocky Linux 9 aarch64
         type: boolean
         default: true
+      rocky10:
+        description: Build Rocky Linux 10
+        type: boolean
+        default: true
+      rocky10-aarch64:
+        description: Build Rocky Linux 10 aarch64
+        type: boolean
+        default: true
       ubuntu-noble:
         description: Build Ubuntu 24.04 Noble
         type: boolean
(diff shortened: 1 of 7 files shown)


.github/workflows/ipa-image-build.yml
bbezak:
```suggestion
        if: ${{ steps.build_rocky_9_ipa_aarch64.outcome == 'failure' || steps.build_rocky_10_ipa_aarch64.outcome == 'failure' }}
```

bbezak (changes requested):
nice work! small NIT

.github/workflows/ipa-image-build.yml
bbezak: let's fail also when r10 build fails

bbezak (dismissed):
approval to start check-review job, please do not merge yet

.github/workflows/ipa-image-build.yml
```diff
@@ -130,15 +159,19 @@ jobs:
...
           aio_vm_network = "${{ vars.HOST_IMAGE_BUILD_NETWORK }}"
           aio_vm_subnet = "${{ vars.HOST_IMAGE_BUILD_SUBNET }}"
           aio_vm_interface = "ens3"
           aio_vm_volume_size = "${{ vars.HOST_IMAGE_BUILD_VOLUME }}"
+          aio_vm_tags = ${{ env.VM_TAGS }}
           EOF
         working-directory: ${{ github.workspace }}/src/kayobe-config/terraform/aio
+        env:
+          VM_NAME: "skc-ipa-image-builder-${{ github.run_id }}"
+          VM_TAGS: '["skc-host-image-build", "PR=${{ github.event.number }}"]'
```
Alex-Welsh: nit: PR=xxxxx is a bit strange if there's no PR right?

owenjones: yes true, I partially nabbed that from the DOCA build workflow (so failed IPA builds get picked up in the CI resource cleaning) and it also sets that tag for some reason - will tidy up once check_review tests have had a chance to complete

bbezak (dismissed):
approval just to run check-review

bbezak (dismissed):
approval just to run check-review

PR #2284, opened 2026-04-23
Implement CIS benchmark hardening for Rocky Linux 10

owenjones:
CIS benchmark hardening has been implemented for Rocky 10 using the RHEL10-CIS[0] role. Rules we override have been mapped to the overrides previously used for Rocky 9.

0: https://github.com/ansible-lockdown/RHEL10-CIS


Alex-Welsh (approved):
LGTM we can tweak further if needed when we have AIOs running

PR #2287, opened 2026-04-27
Reduce max Pulp workers to 12

priteau:
On seed hosts with more than 32 CPU threads, PULP_CONTENT_WORKERS and PULP_API_WORKERS were set to 32, which was causing the number of connections to the bundled Postgres instance to exceed the default max_connections.

Reducing the maximum number of PULP_CONTENT_WORKERS and PULP_API_WORKERS will prevent Postgres from running out of connection slots.

(cherry picked from commit 1bb897f5fedfcff63b7deae4ea06461e7a5405e6)

### DIFF
--- a/etc/kayobe/seed.yml
+++ b/etc/kayobe/seed.yml
@@ -111,8 +111,8 @@ seed_pulp_container:
     # s6-overlay-suexec starts as pid 1
     init: false
     env:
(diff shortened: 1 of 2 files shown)


Alex-Welsh (approved):

Alex-Welsh:
recheck

PR #2294, opened 2026-04-30
Use pull-through cache on SMS for AIOs

Alex-Welsh:
After recent changes at leafcloud, we're regularly getting rate-limited by the RGW backend on Ark.

This change makes AIOs pull through a new authenticating caching proxy on SMS, which should reduce load on Ark, and maybe improve performance at the same time.


Alex-Welsh:
I think this will break the "Check container image tags" action, so don't merge it yet

priteau (approved):

PR #2299, opened 2026-05-07
Bump repositories for CVE-2026-31431 (Copy Fail)

priteau:
This includes Rocky Linux kernels fixing the Copy Fail vulnerability. Mitigation on Ubuntu is applied through an updated kmod package [1].

[1] https://ubuntu.com/blog/copy-fail-vulnerability-fixes-available


bbezak:
is the next step to build container images out of those repos ?

bbezak (approved):

cityofships (approved):

priteau:
> is the next step to build container images out of those repos ?

It can be done in another pull request, since the priority here is to ship the new kernel to customers.

bbezak:
> > is the next step to build container images out of those repos ?
> 
> It can be done in another pull request, since the priority here is to ship the new kernel to customers.

and we need new doca build for new kernel for instance

PR #2300, opened 2026-05-07
Fix RHEL 9 aarch64 DOCA repo version

bbezak:
RHEL 9 DOCA x86 and aarch64 use the same timestamp, but promotion needs a version key for each repo. Add the missing aarch64 key.


cityofships (approved):

bbezak:
image

PR #2316, opened 2026-05-20
 fix: configure `requests` to use `CA` for `Pulp`

jackhodgkiss:
The `Pulp` playbooks used at client sites can fail when `Pulp` is configured to use TLS with certificates generated by our internal `TLS` process in combination with the newer `Pulp` collection and tools.


dougszumski (dismissed):

Alex-Welsh (approved):

Alex-Welsh:
I'd like to see check-review pass before merging, since it runs a short pulp sync

PR #2324, opened 2026-05-22
Add GitHub workflows to auto-label PRs

Alex-Welsh:
Adds two new jobs:
One labels PRs with "waiting-review" when they are opened
The other changes the label to "waiting-author-response" when it is reviewed as "Changes requested"

### DIFF
--- a/.github/workflows/stackhpc-pr-review.yml
+++ b/.github/workflows/stackhpc-pr-review.yml
@@ -1,0 +1,25 @@
+---
+name: Handle GitHub PR Review responses
+
+on:
+  pull_request_review:
+    types:
+      - submitted
+
+permissions:
+  pull-requests: write
+
+jobs:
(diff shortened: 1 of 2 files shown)


Alex-Welsh:
:sunglasses: 
image

priteau (changes requested):
Let's see if it works.

priteau (approved):
I am convinced it works.

priteau (commented):
Does it ever remove the `waiting-author-response` label if a revised commit is pushed?

Alex-Welsh:
> Does it ever remove the waiting-author-response label if a revised commit is pushed?

Not at the moment, I thought about it but decided against it because it's very common for people to just push a partial fix, or test something in CI.

The downside is that authors would have to set the review-required label manually. I'm not sure which is worse.

I think we try this for a bit and see how it goes

PR #2334, opened 2026-05-29
Bump Rocky Linux 9.7 repositories to their final version

priteau:


Alex-Welsh:
@priteau could we also rebuild containers/images etc based on these repos?

priteau:
> @priteau could we also rebuild containers/images etc based on these repos?

Sure, but we need to make sure all the Keystone CVE patches are present.

priteau:
> @priteau could we also rebuild containers/images etc based on these repos?

I have rebuilt all overcloud containers. Seed container build is broken so I pinned the old tag.

etc/kayobe/kolla.yml
```diff
@@ -148,6 +148,12 @@ kolla_sources:
     type: git
     location: https://github.com/stackhpc/stackhpc-inspector-plugins.git
     reference: 1.3.0
+  # TODO(priteau): Remove once
+  # https://review.opendev.org/c/openstack/keystone/+/990504 is merged
+  keystone-base:
+    type: git
+    location: https://github.com/stackhpc/keystone.git
```
Alex-Welsh, line 155: https://github.com/stackhpc/stackhpc-release-train/blob/main/ansible/inventory/group_vars/all/source-repositories#L250
We don't sync keystone 2025.1 right now. Should we?

priteau: We should sync 2025.1 temporarily, while we are waiting for the security patches to merge upstream, just in case an important fix is merged upstream in the meantime.

Currently we are missing two patches but they are not relevant, so we can proceed with the current container images:

* https://review.opendev.org/c/openstack/keystone/+/990476
* https://review.opendev.org/c/openstack/keystone/+/990621

etc/kayobe/pulp-repo-versions.yml
Alex-Welsh, line 1: I think I would prefer in the future to update all RPM repos at the same time when bumping a minor version like this, but I don't think that's a major issue

Alex-Welsh: Perhaps it would be better to split out the RPM and DEB repos to separate files 🤔

Alex-Welsh (approved):
LGTM, couple of comments but just theorising 🙂

etc/kayobe/pulp-repo-versions.yml
priteau: I wanted to limit the changes for this one. I am planning to bump all the repositories in the RL9.8 update.

PR #2336, opened 2026-06-01
docs: support mounting `ca-bundle.crt` in `ceph`

jackhodgkiss:
Services such as RGW integration with `keystone` have in the past required ignoring `SSL` verification failures. This can be avoided by mounting the `ca-bundle.crt` in the container for a more secure deployment.


jackhodgkiss:
Note this requires the following to be merged and released https://github.com/stackhpc/ansible-collection-cephadm/pull/207

PR #2354, opened 2026-06-08
Initial sync of stackhpc/2026.1 with upstream stable/2026.1

Alex-Welsh:


priteau:
Have you pushed all the new tags too?

Alex-Welsh:
> Have you pushed all the new tags too?

Nope, not yet. Not sure if we want to if we're changing our tagging for our own stable releases

mnasiadka (approved):

PR #2370, opened 2026-06-15
Add compatibility for NIC names with dashes

seunghun1ee:
When Ansible gathers facts, it silently converts dashes in network interface names into underscores, resulting precheck failure on Vault to OpenBao migration for overcloud.

Fixed with adding ``replace()`` filter when getting ``secret_store_bind_interface``.

### DIFF
--- a/etc/kayobe/ansible/secret-store/vault-bao-migration-overcloud.yml
+++ b/etc/kayobe/ansible/secret-store/vault-bao-migration-overcloud.yml
@@ -3,10 +3,6 @@
   any_errors_fatal: true
   gather_facts: true
   hosts: controllers
-  vars:
-    secret_store_bind_interface: "{{ internal_net_name | net_interface | replace('-', '_') }}"
-    secret_store_bind_address: "{{ ansible_facts[secret_store_bind_interface].ipv4.address }}"
-    secret_store_api_address: "https://{{ secret_store_bind_address }}:8200"
   tasks:
     - name: Fail if secret store is already configured to be OpenBao
       ansible.builtin.fail:
--- a/etc/kayobe/ansible/secret-store/vault-bao-migration-overcloud.yml
+++ b/etc/kayobe/ansible/secret-store/vault-bao-migration-overcloud.yml
@@ -4,7 +4,7 @@
   gather_facts: true
   hosts: controllers
   vars:
-    secret_store_bind_interface: "{{ internal_net_name | net_interface }}"
+    secret_store_bind_interface: "{{ internal_net_name | net_interface | replace('-', '_') }}"
     secret_store_bind_address: "{{ ansible_facts[secret_store_bind_interface].ipv4.address }}"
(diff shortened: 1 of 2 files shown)


etc/kayobe/ansible/secret-store/vault-bao-migration-overcloud.yml
```diff
@@ -4,7 +4,7 @@
   gather_facts: true
   hosts: controllers
   vars:
-    secret_store_bind_interface: "{{ internal_net_name | net_interface }}"
+    secret_store_bind_interface: "{{ internal_net_name | net_interface | replace('-', '_') }}"
```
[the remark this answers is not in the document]
seunghun1ee: oopsie. Fixed

seunghun1ee:
Tested on multinode with custom interface names with dashes

Alex-Welsh (commented):
Sorry to scope-creep this a bit, but should we also do the same for other `secret-store/*` playbooks?

seunghun1ee:
Not really. The equivalent variable only appears in secret-store-deploy-overcloud.yml. But the way it gets the ip address from the chosen network interface is different. The deployment playbook uses net_ip filter whereas the migration playbook uses attributes from ansible_facts.

I can't remember the background of this difference. Maybe because net_ip filter is known that it can fail and we wanted to move away from it?

etc/kayobe/ansible/secret-store/vault-bao-migration-overcloud.yml
mnasiadka: Shouldn't the imported playbook be fixed to do this type of replacement? That should be enough?

seunghun1ee: Do you mean it would be better to keep the line 42 and make collection's playbook to do replace instead?

mnasiadka: Yes, SKC is just a user of the collection - there might be other users of the collection. IMHO it's better to do the replace in the collection.

seunghun1ee: The collection doesn't really care about the network interface, it only care about the ip address of each overcloud vault. So, in this case, I think it's user's role to provide correct ip.
But I do need to remove vars from line 7 because these are not actually used from the tasks before collection.

mnasiadka: Well, maybe it would be easier to replace that in net_interface?

seunghun1ee: I'm not sure if all of use of net_interface in SKC would work with the converted names

mnasiadka: Of course fixing it here is fine, it was just an idea.

Alex-Welsh (approved):

PR #2373, opened 2026-06-16
ci: use diff of previous patchset for path filters

Alex-Welsh:
When a user makes a PR, CI is run conditionally based on the files that change. If a second commit is then pushed, e.g. to update docs or add a relase note, that commit would retrigger all CI even though it doesn't affect the filtered files. This change fixes that interaction, so the new patchset is compared to the previous, instead of the branch it is merging into.


Alex-Welsh:
Cool idea I think, but I've just realised it can be used to bypass CI if you push a major change, then quickly push a minor one - to docs or something - before CI on the first one is finished. Would have to check the status of the previous patchset and that seems like a bit of a mess. Dropping this for now

PR #2374, opened 2026-06-16
Remove hostname fix reboot

Alex-Welsh:
Previously there was a hostname-related OVN bug that required a reboot in the AIO CI instances to mitigate. It's been fixed for years[1] so this change reverts the workaround. It'll speed up CI and it's better to know if we regress.

[1] https://bugs.launchpad.net/kolla-ansible/+bug/2080552


Alex-Welsh:
Back to draft while I test a fix for CI

Alex-Welsh:
See also https://github.com/stackhpc/stackhpc-cloud-tests/pull/18

priteau (approved):

PR #2383, opened 2026-06-22
Valkey containers for Rocky 9 and Ubuntu Noble

owenjones:
To enable redis -> valkey migrations, we need valkey images buildable on OSes other than Rocky 10.

### DIFF
--- a/etc/kayobe/kolla-image-tags.yml
+++ b/etc/kayobe/kolla-image-tags.yml
@@ -13,3 +13,9 @@ kolla_image_tags:
     ubuntu-noble: 2025.1-ubuntu-noble-20260205T152450
   ovn:
     rocky-9: 2025.1-rocky-9-20260608T100935
+  valkey:
+    rocky-9: 2025.1-rocky-9-20260618T122602
+    ubuntu-noble: 2025.1-ubuntu-noble-20260618T122602
+  valkey_sentinel:
+    rocky-9: 2025.1-rocky-9-20260618T122602
+    ubuntu-noble: 2025.1-ubuntu-noble-20260618T122602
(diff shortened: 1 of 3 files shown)


etc/kayobe/kolla-image-tags.yml
```diff
@@ -13,3 +13,9 @@ kolla_image_tags:
     ubuntu-noble: 2025.1-ubuntu-noble-20260205T152450
   ovn:
     rocky-9: 2025.1-rocky-9-20260608T100935
+  valkey:
+    rocky-9: 2025.1-rocky-9-20260618T122602
+    ubuntu-noble: 2025.1-ubuntu-noble-20260618T122602
+  valkey_sentinel:
```
mnasiadka: Shouldn't we rather update kolla-image-tags.py so valkey matches also valkey_sentinel?

owenjones: ~~That would just be setting it [here](https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2025.1/tools/kolla-images.py#L101)? Currently it just matches how redis was setup.~~

We'd either need a new exception map/logic in `kolla-images.py` to be able to override what's in kolla-ansible, or [changes in kolla-ansible](https://opendev.org/openstack/kolla-ansible/src/branch/master/ansible/roles/valkey/defaults/main.yml#L31) made and backported to change the tag inheritance—second option is probably better but would be breaking upstream?

priteau: We will make an upstream change (trying to remain backward-compatible) to fix this properly, merging in the meantime.

priteau (approved):

PR #2389, opened 2026-06-26
Set kolla base version with ansible facts

owenjones:
Switch to setting `kolla_base_distro_version` using ansible facts, rather than a static mapping, so the correct containers get deployed on hosts when we have a mixture of versions (during Rocky 9 to 10 migration).

### DIFF
--- a/etc/kayobe/kolla/globals.yml
+++ b/etc/kayobe/kolla/globals.yml
@@ -5,21 +5,22 @@
 # This is necessary for os migrations where mixed clouds might be deployed
 kolla_base_distro: "{% raw %}{{ ansible_facts.distribution | lower }}{% endraw %}"
 
+kolla_base_distro_version_from_facts: true
+
 # Use facts so this is determined correctly when the control host OS differs
-# from os_distribuition.
-kolla_base_distro_version: "{% raw %}{{ kolla_base_distro_version_default_map[kolla_base_distro] }}{% endraw %}"
+# from os_distribution.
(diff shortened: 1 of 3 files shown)


etc/kayobe/kolla/globals.yml
```diff
@@ -6,20 +6,19 @@
...
 
 # Convenience variable for base distro and version string.
 kolla_base_distro_and_version: "{% raw %}{{ kolla_base_distro }}-{{ kolla_base_distro_version }}{% endraw %}"
 
-# Override the version map used, so we can deploy Rocky 10 container images
-# on Rocky 10 hosts.
-kolla_base_distro_version_default_map: {
-  "centos": "stream9",
-  "debian": "bookworm",
-  "rocky": "{{ os_release }}",
```
Alex-Welsh, line 20: Would it be easier to set os_release based on ansible facts? Or use the same lookup we do for CI https://github.com/stackhpc/stackhpc-kayobe-config/blob/c3f3c4a445eb3ebefef95633d759e6ccc360e26d/etc/kayobe/environments/ci-multinode/globals.yml#L54-L56?

owenjones: That wouldn't work for the migration though—currently we need to set `os_release` to 10 at the start of the process to pull in the correct new host images when re-provisioning nodes, but as we're setting the kolla base version based off that, running a service deploy after migrating the first controller causes all containers to move to 10, rather than matching their host version.

There's a rough plan to change this mapping upstream so it can be toggled to use facts to determine container version, this was put together to bridge the gap until then if needed for SMS migration.

PR #2391, opened 2026-06-29
Add 2026.1 documentation

priteau:
The OpenStack upgrade page still needs a full review.


Alex-Welsh (dismissed):
Gemini is correct, the `ubuntu-noble.rst` page no longer exists, which breaks a link and fails CI

Alex-Welsh (approved):

PR #2394, opened 2026-06-29
cephadm: Rocky 10 support for deploying ceph

owenjones:
* cephadm deployment now supported in Rocky 10; The cephadm collection has been bumped to 1.22.1 which updates the Ceph repository URL for Squid on Rocky 10. The Rocky crypto policy in use during Ceph deployment is required to be downgraded to LEGACY in order to use Ceph's release signing key. The policy is reverted after deployment or on deployment failure

### DIFF
--- a/etc/kayobe/ansible/ceph/cephadm-deploy.yml
+++ b/etc/kayobe/ansible/ceph/cephadm-deploy.yml
@@ -7,6 +7,37 @@
     - cephadm
     - cephadm-deploy
   tasks:
+    # note(owenjones): the Ceph release signing key still uses SHA-1, which requires us to
+    # downgrade to the LEGACY policy on Rocky 10 until a new key is published.
+    # A later Ceph release might fix this - https://tracker.ceph.com/issues/74861
+    - name: Discover current active crypto policy (Rocky 10)
+      ansible.builtin.command:
+        cmd: update-crypto-policies --show
+      become: true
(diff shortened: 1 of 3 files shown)


etc/kayobe/ansible/ceph/cephadm-deploy.yml
```diff
@@ -7,6 +7,36 @@
...
-      ansible.builtin.import_role:
-        name: stackhpc.cephadm.cephadm
+      block:
+        - name: Apply Cephadm role
+          ansible.builtin.import_role:
+            name: stackhpc.cephadm.cephadm
+      always:
+        - name: Revert crypto policy (Rocky 10)
+          ansible.builtin.command:
+            cmd: update-crypto-policies --set {{ crypto_policy.stdout | default('DEFAULT') }}
```
Alex-Welsh, line 38: What happens when a system is already on LEGACY, and you try to set the LEGACY policy?

i.e. do we need another condition at the end to say
``` when: crypto_policy.stdout | default('DEFAULT') != 'LEGACY'```
like above

owenjones: Re-setting the policy to LEGACY works fine, but skipping this step if not required also wouldn't hurt so will add it.

Alex-Welsh (approved):

PR #2399, opened 2026-06-30
Refresh 2025.1 images

priteau:
Bump all repositories and rebuild all images.


jovial (dismissed):

Alex-Welsh (approved):
Really, I'd very much like to multinode test change sets like this, but we just don't have the capacity to do that right now. I think this is good enough

PR #2403, opened 2026-07-02
Refresh 2025.1 images

jovial:


priteau:
I would prefer to keep using my branch which has more fixes.

PR #2416, opened 2026-07-07
Add copy of tempest config for aio environment

seunghun1ee:
Running tempest with the ``automated-setup.sh`` for ``aio`` environment fails because it can't locate tempest override configuration for ``aio``.
Since ``ci-aio`` Kayobe environment inherits ``aio`` environment, Renamed ``tempest-ci-aio.overrides.conf`` to
``tempest-aio.overrides.conf`` then made symlink to ``tempest-ci-aio.oerrides.conf`` to make them sync.

### DIFF
--- a/.automation.conf/tempest/tempest-aio.overrides.conf
+++ b/.automation.conf/tempest/tempest-aio.overrides.conf
@@ -1,0 +1,2 @@
+[volume]
+catalog_type = volumev3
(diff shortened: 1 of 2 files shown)


Alex-Welsh (approved):
Github doesn't really know how to handle the diff, but it looks good when checked manually

PR #2436, opened 2026-07-17
Amphorae build fixes

cityofships:
Arguably, we should be using the latest DIB and not pinned to the OpenStack release. This ensures recent OS versions compatibility and generic fixes and improvements included.

Also, fix Ubuntu vs. Rocky dependencies installation.

### DIFF
--- a/etc/kayobe/ansible/maintenance/octavia-amphora-image-build.yml
+++ b/etc/kayobe/ansible/maintenance/octavia-amphora-image-build.yml
@@ -2,7 +2,6 @@
 - name: Build an Octavia Amphora image
   hosts: "{{ amphora_builder_group | default('seed') }}"
   vars:
-    amphora_dib_upper_constraints_file: "{{ pip_upper_constraints_file }}"
     amphora_image_dest: "{{ image_cache_path }}/amphora-x64-haproxy-{{ openstack_release }}.qcow2"
   tasks:
     - name: Install EPEL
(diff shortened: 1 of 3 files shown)


etc/kayobe/stackhpc.yml
```diff
@@ -270,6 +270,9 @@ download_amphora_from_ark: true
 # Octavia Amphora image version
 stackhpc_amphora_image_version: "2025.1-20260319T131237"
```
Alex-Welsh: @cityofships did you mean to also bump the image at the same time?

cityofships: Good point, let's see how it behaves in the [workflow](https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/30525389173)

etc/kayobe/ansible/maintenance/octavia-amphora-image-build.yml
```diff
@@ -2,7 +2,6 @@
 - name: Build an Octavia Amphora image
   hosts: "{{ amphora_builder_group | default('seed') }}"
   vars:
-    amphora_dib_upper_constraints_file: "{{ pip_upper_constraints_file }}"
```
dougszumski, line 5: nit: It would be better to set the default here given that the playbook relies on this variable being defined. Then -e on the command line still works.

eg. similar to this - just a comment used in the file included with `-e`:
https://github.com/openstack/kayobe/blob/master/etc/kayobe/overcloud-dib.yml#L76

etc/kayobe/stackhpc.yml
cityofships: The workflow's been green, so updating 👍

PR #2438, opened 2026-07-20
CI: upload docs artifacts

bbezak:
it can be useful for reviewers


Alex-Welsh (approved):
Nice idea, it works well

PR #2446, opened 2026-07-21
Add upgrade-prerequsites.sh for Epoxy to Gazpacho

seunghun1ee:
Added Valkey migration script.
This checks how `kolla_enable_redis`` is evaluated.
If it's true, the script calls K-A's Valkey migration.
Otherwise, it's a NOOP.

### DIFF
--- a/tools/upgrade-prerequisites.sh
+++ b/tools/upgrade-prerequisites.sh
@@ -1,0 +1,39 @@
+#! /usr/bin/bash
+
+# This script is intended to be run in CI to test upgrades.
+# It executes any preparation steps that must be perfomed before upgrading
+# OpenStack services.
+
+# NOTE(upgrade): This script is unique to each release. It may not be required
+# for some releases.
+
+set -ex
+
+function prechecks() {
+    if [[ ! $KAYOBE_CONFIG_PATH ]]; then
+        echo "Environment variable \$KAYOBE_CONFIG_PATH is not defined"
+        echo "Ensure your environment is set up to run kayobe commands"
+        exit 2
+    fi
+
+    echo "Installing dependencies..."
+    if type dnf > /dev/null 2>&1; then
+        sudo dnf -y install jq
+    else
+        sudo apt update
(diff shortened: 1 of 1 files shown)


seunghun1ee:
CI in this PR won't test the script. But it was tested on AIO manually.

tools/upgrade-prerequisites.sh
```diff
@@ -0,0 +1,39 @@
...
+        sudo apt -y install jq
+    fi
+
+}
+
+function valkey_migration() {
+
+    enable_redis=$(kayobe configuration dump --var-name kolla_enable_redis -l localhost | jq .localhost)
+    if [[ $enable_redis == "true" || $enable_redis == "yes" ]]; then
+        kayobe kolla ansible run migrate-valkey -ke enable_redis=false -ke enable_valkey=true
```
Alex-Welsh, line 33: Rather than setting inline args, I think we should do the same `sed` trick we did for the 2024.1 rabbit migrations so the config persists.

Alex-Welsh (changes requested):

tools/upgrade-prerequisites.sh
seunghun1ee: The config did not persist on that one though. The process was
1. Add related config options at kolla/globals.yml with sed
2. Run K-A rabbitmq-migration
3. Remove related config options with sed
See https://github.com/stackhpc/stackhpc-kayobe-config/blob/stackhpc/2024.1/tools/upgrade-prerequisites.sh

Also, Unlike the 2024.1 to 2025.1, this involves switching the values of possibly already existing variables which is messy.
Furthermore, keeping changed ``kolla_enable_redis`` and ``kolla_enable_valkey`` is highly likely to cause git problem when moved to 2026.1.

Alex-Welsh: Seems fair

Alex-Welsh (approved):

PR #2448, opened 2026-07-22
GT-1835: Fix IPA for stream 10

dougszumski:
This switches to an IPA builder fork which is required for centos-10 stream IPA images, which are required (currently) to provide a version of qemu-img that can handle 4k sectors from Dell Perc RAID controllers which don't support other block sizes.

Anyone not affected can carry on using stream-9 images until Rocky ones exist.

We can drop this patch if we can backport
https://review.opendev.org/c/openstack/ironic-python-agent-builder/+/995589

Thanks to @owenjones for finding the patch ^.

### DIFF
--- a/etc/kayobe/ipa.yml
+++ b/etc/kayobe/ipa.yml
@@ -14,10 +14,14 @@
 #ipa_build_source_version:
 
 # URL of IPA builder source repository.
-#ipa_builder_source_url:
+# NOTE(dougszu): This fork is required for centos-10 stream IPA images, which
+# are required (currently) to provide a version of qemu-img that can handle
+# 4k sectors from Dell Perc RAID controllers which don't support other block
+# sizes.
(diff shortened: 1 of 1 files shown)


jovial (commented):
Looks good, but we should add this to release train: https://github.com/stackhpc/stackhpc-release-train/blob/3c0a6cc839afe09888534a344149cab4097d1789/ansible/inventory/group_vars/all/source-repositories so that the branch is kept up to date

priteau (approved):

dougszumski:
> Looks good, but we should add this to release train: https://github.com/stackhpc/stackhpc-release-train/blob/3c0a6cc839afe09888534a344149cab4097d1789/ansible/inventory/group_vars/all/source-repositories so that the branch is kept up to date

thanks, agree: https://github.com/stackhpc/stackhpc-release-train/pull/550

dougszumski:
CI fail is unrelated - manually tested on Sunrise system

PR #2452, opened 2026-07-23
Update RHEL10 CIS configuration

priteau:
Various fixes to our RHEL10 CIS configuration, see release note for details.


etc/kayobe/inventory/group_vars/cis-hardening/cis
```diff
@@ -88,12 +88,16 @@ rhel10cis_auditd_max_log_file_size: 1024
 rhel10cis_auditd_max_log_file: 10
 rhel10cis_auditd_space_left_action: syslog
 
+# Prevent system from halting if disk is full
+rhel10cis_auditd_disk_full_action: suspend
```
seunghun1ee, line 92: Why not setting it to syslog like we did before?

PR #2466, opened 2026-07-28
Bump repositories and rebuild OFED

priteau:


Alex-Welsh (approved):

Alex-Welsh:
I have just thought that really, we should rebuild all our kolla containers if we're bumping all package repos as well so they stay in sync. We've skipped that step in the past though, it's not the end of the world

priteau:
> I have just thought that really, we should rebuild all our kolla containers if we're bumping all package repos as well so they stay in sync. We've skipped that step in the past though, it's not the end of the world

I think we can do the container image rebuild a bit less regularly, since it is more likely to introduce bugs. I will propose one for the next sync.

PR #2480, opened 2026-08-04
Update host packages and reboot during AIO upgrade

priteau:
This is now required if the previous release is running an older kernel, otherwise installation of kernel-modules-extra fails.

(cherry picked from commit b9418fd2139f06f9e985400fde6d8af0aac5b496)


priteau:
The AIO VM doesn't come back online after rebooting.

PR #2490, opened 2026-08-06
CI: Skip kolla-packages during AIO upgrade

priteau:
AIO upgrade jobs on Rocky Linux are currently broken because ansible-collection-kolla tries to install kernel-modules-extra matching the running kernel after DNF repositories have been updated. The package can be unavailable if there is a minor version change (9.7 to 9.8) or if the kernel snapshot of the newer release is behind the one of the previous release.

Skip kolla-packages for the second overcloud-host-configure execution to avoid this issue until we find a better solution.


Alex-Welsh (changes requested):
LGTM if you drop the DNM commit

PR #2492, opened 2026-08-06
Use pull-through cache on SMS for AIOs (2024.1)

priteau:


Alex-Welsh:
I'm not sure this is possible without a lot of work. Shall we close it, or do you have an idea for how to fix it?

priteau:
> I'm not sure this is possible without a lot of work. Shall we close it, or do you have an idea for how to fix it?

I thought it was going to be simple, let's close it.

PR #2496, opened 2026-08-06
Make readthedocs config use wildcard

Alex-Welsh:
There's no point maintaining this ever-growing list of branches in the config when we can just use a wildcard. It might pull a few random feature branches like `stackhpc/2025.1-foo` but that's not really an issue.

I'm somewhat hoping this also fixes some recent rtd build failures if we backport it, not entirely sure it will though https://app.readthedocs.org/projects/stackhpc-kayobe-config/builds/33943470/

## Summary by CodeRabbit

* **Chores**
  * Read the Docs builds now automatically include all `stackhpc/*` branches, improving branch availability for documentation builds.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,7 +13,7 @@ build:
     python: "3.12"
   jobs:
     post_checkout:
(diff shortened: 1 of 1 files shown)


elelaysh (approved):
LGTM

priteau (approved):

PR #2501, opened 2026-08-10
Rocky 10 migration documentation

owenjones:
Documentation to cover the migration from Rocky Linux 9 to Rocky Linux 10 has been added.


doc/source/operations/rocky-10.rst
```diff
@@ -0,0 +1,464 @@
...
+
+    kayobe overcloud container image pull --kolla-tag valkey
+
+Perform the migration:
+
+.. code-block:: console
+
+    kayobe kolla ansible run migrate-valkey
+
+*Command to validate migration*
```
darmach: We could you the above-mentioned command in here

darmach (dismissed):

doc/source/operations/rocky-10.rst
owenjones: I was waiting for someone to give the 'proper' way to do it - during testing I was just using `ansible-vault edit` and copying the existing password over..

darmach:
It would be good to add the step to bump installed Kayobe otherwise we'll end up with outdated Kayobe if not working in new Beokay checkout.

We should sync containers **after** valkey flag enabled.

Missing Seed ToDO to be added - but thats obvious ;)

jovial (approved):

darmach:
@jovial @owenjones I see that this was merged, but we do have gaps in these docs - like missing pieces [(Deploy public key/install cephadm - this step needs checking <--](https://github.com/stackhpc/stackhpc-kayobe-config/blob/341118ec31aae0f9e066d34cc9ff6384981064fe/doc/source/operations/rocky-10.rst?plain=1#L285)  <- readding the node back to the cluster, or reference to non-existing playbook [for fixing openbao ](https://github.com/stackhpc/stackhpc-kayobe-config/blob/341118ec31aae0f9e066d34cc9ff6384981064fe/doc/source/operations/rocky-10.rst?plain=1#L310) (step which I stumbled upon just now)

PR #2508, opened 2026-08-12
Ensure root.crt cert exists for proxysql

Alex-Welsh:
Without this, proxysql upgrade to Gazpacho will fail saying 
`ERROR MissingRequiredSource:
/var/lib/kolla/config_files/ca-certificates/root.crt file is not found etc/kayobe/ansible/secret-store/secret-store-generate-internal-tls.yml`


seunghun1ee (approved):
Thanks for catching this

MoteHue (approved):

PR #2512, opened 2026-08-13
Bump cinder image tag to fix cinder NetApp ONTAP and self-signed certs

chouroukblizzard:
This fixes compatibility between Cinder and NetApp ONTAP when the storage backend is deployed using self-signed certificates.
[stackhpc/cinder#153](https://github.com/stackhpc/cinder/pull/153) has been merged and is included for the 2025.1 image build.


MoteHue (approved):
LGTM, thanks!

PR #2535, opened 2026-08-24
Bump Ubuntu Ceph libs version to Umbrella

mnasiadka:
Use download.ceph.com for that, because Ubuntu does not even have an SRU ticket for 19.2.6 (they have one for 19.2.4 that is still in progress).

Using Umbrella because Ceph upstream does not build older releases for Ubuntu Noble.


mnasiadka:
Doesn't make any sense - Umbrella doesn't contain aes256k yet - https://github.com/ceph/ceph/pull/71172

PR #2537, opened 2026-08-24
Add support for consumer grade gpus

GregWhiteyBialas:
Add suppot for consumer grade nvidia gpus. Those cards expose
audio devices which id, needs to be added as one of kernel 
arguments in vfio-pci.id variable along with display_id to work properly.
Support for Nvidia p4000 is also added as working example of those cards.


priteau (approved):
Tested and working, thanks!

priteau:
CI failures are unrelated.

PR #2557, opened 2026-08-27
add overrides upstream sync to OS upgrade documentation

g0rgamesh:


priteau:
Most Epoxy upgrades are now done, I suggest targeting 2026.1 instead.

PR #2558, opened 2026-08-27
Switch to upstream blazar

Alex-Welsh:
Switches to upstream blazar instead of our stackhpc fork.

Also bumps our kolla/kolla-ansible pins and switches back to using tags. This should let them auto-update with our weekly syncs again


Alex-Welsh:
Needs a rebase and a rebuild

PR #2561, opened 2026-08-27
Persist cinderv3 endpoint after upgrade

Alex-Welsh:


etc/kayobe/kolla/globals.yml
```diff
@@ -67,3 +67,23 @@ prometheus_blackbox_exporter_endpoints_kayobe:
...
 
 prometheus_openstack_exporter_interval: "{{ stackhpc_prometheus_openstack_exporter_interval }}s"
+
+#############################################################################
+# Additional configuration
+
+# NOTE(Alex-Welsh): The Gazpacho release removes the cinderv3 endpoint by
+# default. A lot of integrations still rely on the old endpoint, so we enable it
+# explicitly.
+cinder_ks_services:
```
bbezak, line 77: this will not be evaluated in `kayobe overcloud service upgrade` run, as upgrade.yml in k-a is not importing register.yml - only deploy.yml does - so we will end up with cloud without volumev3 on upgrade to gazpacho - subsequent service deploy will re-add it though

bbezak (changes requested):

etc/kayobe/kolla/globals.yml
Alex-Welsh: That's unfortunate. Better to patch our KA fork then?

Alex-Welsh: Reverting in KA fork https://github.com/stackhpc/kolla-ansible/pull/937

Alex-Welsh:
Backporting in kolla-ansible fork instead https://github.com/stackhpc/kolla-ansible/pull/937

PR #2563, opened 2026-08-28
Fix reboot prompt for multiple hosts

priteau:
When rebooting multiple hosts concurrently by overriding the serial value, Ansible would fail on the next task with:

    'pause_prompt' is undefined

Fix by using delegate_to/run_once.


Alex-Welsh (commented):
When you run this, does `play_hosts` correctly resolve now that it it delegated to localhost and run once?

PR #2564, opened 2026-08-28
[2025.1] Use Ceph Squid from download.ceph.com on EL9

bbezak:
The CentOS Storage SIG only has Squid 19.2.5 for el9, which misses the 19.2.6 CVE fixes. 
Rocky 10 stays on the SIG, it has 19.2.6 there.

Bumps the image tags for the Ceph consumers, so Rocky 9 and Rocky 10 both ship 19.2.6. 
Ubuntu images are bumped as well, so all three carry the same OpenStack service versions.

Needs the kolla change [1].

[1] https://review.opendev.org/c/openstack/kolla/+/1002473


etc/kayobe/kolla-image-tags.yml
```diff
@@ -8,6 +8,22 @@ kolla_image_tags:
...
+    rocky-10: 2025.1-rocky-10-20260828T080743
+    ubuntu-noble: 2025.1-ubuntu-noble-20260828T080743
+  glance:
+    rocky-9: 2025.1-rocky-9-20260828T080743
+    rocky-10: 2025.1-rocky-10-20260828T080743
+    ubuntu-noble: 2025.1-ubuntu-noble-20260828T080743
+  gnocchi:
+    rocky-9: 2025.1-rocky-9-20260828T080743
+    rocky-10: 2025.1-rocky-10-20260828T080743
+    ubuntu-noble: 2025.1-ubuntu-noble-20260828T080743
```
priteau: Did you enable Gnocchi in a custom branch?

bbezak: actually it was built just by adding regex to the container image build CI job

priteau (approved):

PR #2565, opened 2026-08-28
[2024.1] Use Ceph Squid from download.ceph.com on EL9

bbezak:
The CentOS Storage SIG only has Squid 19.2.5 for el9, which misses the 19.2.6 CVE fixes.

Bumps the Rocky 9 image tags for the Ceph consumers, so they ship 19.2.6. Ubuntu images are bumped as well, so they carry the same OpenStack service versions as Rocky 9.

Needs the kolla change [1].

[1] https://review.opendev.org/c/openstack/kolla/+/1002538


etc/kayobe/kolla-image-tags.yml
```diff
@@ -31,8 +31,13 @@ kolla_image_tags:
...
   glance:
-    ubuntu-jammy: 2024.1-ubuntu-jammy-20250903T130503
-    ubuntu-noble: 2024.1-ubuntu-noble-20250903T130503
+    rocky-9: 2024.1-rocky-9-20260828T084152
+    ubuntu-jammy: 2024.1-ubuntu-jammy-20260828T084152
+    ubuntu-noble: 2024.1-ubuntu-noble-20260828T084152
+  gnocchi:
+    rocky-9: 2024.1-rocky-9-20260828T084152
+    ubuntu-jammy: 2024.1-ubuntu-jammy-20260828T084152
+    ubuntu-noble: 2024.1-ubuntu-noble-20260828T084152
```
priteau: I don't think we build Gnocchi images as part of release train?

bbezak: fair point, thx

Alex-Welsh:
zuul check failure is unrelated

```
ERROR: Cannot install -r /home/zuul/src/github.com/stackhpc/stackhpc-kayobe-config/test-requirements.txt (line 6) because these package versions have conflicting dependencies.

The conflict is caused by:
    doc8 2.0.0 depends on stevedore
    The user requested (constraint) stevedore===5.9.1

Additionally, some packages in these conflicts have no matching distributions available for your environment:
    stevedore

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict
```

PR #2572, opened 2026-09-01
tox -e pep8 fixes and improvment

GregWhiteyBialas:
Fixing linter warnings and make CI/CD fail on any yamllint error.

Warnings are ignored when it comes to PR gatekeeping, however they are making output confusing when check will fail on real error. 
This PR is fixing all (current)  errors that yamllint is reporting and makes check more strict, so now all yamllint errors will cause whole workflow to fail.


Alex-Welsh:
Thanks Whitey, can we make tox error on warnings now?

GregWhiteyBialas:
@Alex-Welsh Done.

priteau (approved):
Nice!

PR #2577, opened 2026-09-03
Disable l2vni trunk reconciliation

TBExtent:
L2VNI trunk reconciliation was added to ironic neutron agent in gazpacho, and enabled by default. It assumes that neutron trunk and segment are enabled, which is not guaranteed, so it is safest if it is disabled by default instead.
See this patch: https://review.opendev.org/c/openstack/networking-baremetal/+/974619

### DIFF
--- a/etc/kayobe/environments/baremetal/kolla/config/neutron/ironic_neutron_agent.ini
+++ b/etc/kayobe/environments/baremetal/kolla/config/neutron/ironic_neutron_agent.ini
@@ -1,0 +1,5 @@
+[l2vni]
+{% if kolla_enable_ironic_neutron_agent | default(False) | bool %}
+enable_l2vni_trunk_reconciliation = False
+enable_l2vni_trunk_reconciliation_events = False
+{% endif %}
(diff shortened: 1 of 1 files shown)


priteau (changes requested):
Can you include more details on why this is necessary? Is there an upstream bug open (if this is a bug) or a configuration guide that we can link to?

priteau (changes requested):
For example, you can link to this change in the commit: https://review.opendev.org/c/openstack/networking-baremetal/+/974619

TBExtent:
Done :)

priteau (approved):

etc/kayobe/environments/baremetal/kolla/config/neutron/ironic_neutron_agent.ini
```diff
@@ -0,0 +1,5 @@
+[l2vni]
```
priteau: Can we put the `[l2vni]` header inside the Jinja block, or will it break templating?

priteau (changes requested):

etc/kayobe/environments/baremetal/kolla/config/neutron/ironic_neutron_agent.ini
TBExtent: It seems happy with the header inside the block from testing.

priteau (approved):

PR #2594, opened 2026-09-10
add override section to upgrade doc

g0rgamesh:


jovial (approved):
Never used sdiff, but this is a nice reminder!

PR #2614, opened 2026-09-18
feat: Add oslo config validation to all-in-one

MurrayGroves:


.github/actions/oslo-config-validator/action.yml
```diff
@@ -0,0 +1,60 @@
+---
+# This workflow can be run by any other workflow which has a Kayobe environment.
+# It should be run as a composite action (https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-a-composite-action-within-the-same-repository)
+# It runs the oslo-config-validator and reports warnings/errors.
+
+name: oslo-config-validator
+runs:
+  using: "composite"
+  steps:
+    - uses: LouisBrunner/checks-action@v2.0.0
```
Alex-Welsh: Third-party actions like this have to be individually approved, anything else is blocked across the GitHub Org. You can see a list of the currently approved actions here
https://github.com/stackhpc/github-actions-list/blob/main/stackhpc.pinned.txt

LouisBrunner/checks-action isn't in the list right now, so will be blocked.

PR #2616, opened 2026-09-21
Fix inspection plugins

jovial:


jovial:
Trying to build new images here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/35611805964

jovial:
Building here: https://github.com/stackhpc/stackhpc-kayobe-config/actions/runs/35611805964

